PR #38, opened 2013-10-04
Newkiss + Safebuff

johandc:
Pull request for both safebuff and newkiss. (The newkiss branch was made on the safebuff brance).

Safebuff:
A more safe memory allocator for csp with support for:
double free detection
buffer overflow detection
invalid pointer detection

Newkiss:
A new kiss driver which uses CSP's builtin CRC function and fixes a double free bug in kiss driver.
CSP's builtin CRC function now uses PROGMEM on AVR8 targets
Fix severe conn leak when using security check.

Change to interface driver API:
The TX function for drivers have been updated to include the csp_iface_t \* interface pointer. And the interface structure have been updated to include a void \* for the driver instance. This enables drivers to have multiple interfaces associated.

All changes open for discussion.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,8 +1,38 @@
-libcsp 1.x, xxxx-xx-xx
+libcsp 1.2, 25-10-2013
 ----------------------
+- Feature release
 - New: CMP service for peek and poke of memory
+- New: CMP interface statistics struct is now packed
+- New: Faster O(1) buffer system with reference counting and automatic alignment
+- New: Thread safe KISS driver with support for multiple interfaces
+- New: CSP interface struct now holds an opaque pointer to driver handle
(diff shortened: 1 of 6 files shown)


jledet:
I'm generally ok with this being pulled. I have added some comments where I have questions.

Just a note: I would have preferred if the commits had been squashed to hide commits with errors (accidentally committed debug output and so on), and if the newkiss/safebuf branches had been rebased instead of merged since this makes it a lot easier to review the code. Ideally, this had also been split in two or more pull requests (safebuf, newkiss, conn pool leak, waf stuff), but what the heck :)

I'm happy that we now have reference counting, overflow detection and proper alignment guarantees for the buffer system.

Acked-by: Jeppe Ledet-Pedersen jlp@satlab.org

PR #40, opened 2014-10-26
Added csp buffer overhead to buffer size

lbromo:
After debugging with @jledet due to an error on AAUSAT, we believe that _CSP_BUFFER_PACKET_OVERHEAD_ should be added when determining the size of a new buffer in _csp_buffer_init_


johandc:
Yes it should. I have always inited the buffer system with 20-30 bytes more than the maximum MTU i was going to use anyways, but adding it here is more userfriendly :+1:

PR #50, opened 2015-06-30
I had a conflict between libcsp expected FREERTOS_VERSION and the FREERTOS_VERSION defined in FreeRTOS header.

nicob87:
trcKernelPortFreeRTOS.h:#define FREERTOS_VERSION_NOT_SET         0
trcKernelPortFreeRTOS.h:#define FREERTOS_VERSION_7_3_OR_7_4      1
trcKernelPortFreeRTOS.h:#define FREERTOS_VERSION_7_5_OR_7_6      2
trcKernelPortFreeRTOS.h:#define FREERTOS_VERSION_8_0_OR_LATER    3


nicob87:
mmmm, you are right! renaming it looks like the right solution!

PR #54, opened 2015-07-16
Python binding using cffi

agustinhenze:
Hello, we are using libcsp in our project and let me tell you that this project is awesome. We needed to write the same code written here :), I really appreciate that you release it under an open source license. I know this is a pull request, but I did not want to miss the opportunity to give the thanks.

Our project has some computers running linux that interact with our embedded systems, so we want/need to use Python3, you can imagine my happiness when I found that you have a folder called python/binding. After a shots we discover that is not working really well, you need to use ctypes by hand sometimes and doesn't work on Python3 either.

So we made a binding for us using python-cffi module but it was too ad-hoc to ask for a pull request with this. BTW we are using csp+can+socketcan from Python successfully :). Thinking on that it should be uploaded here I started to think how would be the best way to do a binding as automatically as possible. After a few tries, I concluded it is not possible do it well without some metadata added in the sources.

CFFI for Python has two modes (the documentation mentions in-line and out-of-line but they are not really modes for me). I am using the API level mode, basically you need to use two functions, the first one is [cdef](https://cffi.readthedocs.org/en/latest/cdef.html#ffi-cdef-declaring-types-and-functions):

`ffi.cdef(source): parses the given C source. It registers all the functions,
types, constants and global variables in the C source. The types can be used
immediately in ffi.new() and other functions. Before you can access the
functions and global variables, you need to give ffi another piece of
information: where they actually come from (which you do with either
ffi.dlopen() or ffi.set_source()).`

The second function is [set_source](https://cffi.readthedocs.org/en/latest/cdef.html#ffi-set-source-preparing-out-of-line-modules), quoting the doc again

```In API mode, the c_header_source argument is a string that will be pasted
into the .c file generated. This piece of C code typically contains some
# include, but may also contain more, like definitions for custom “wrapper” C

functions.```

Trying to summarize the first one `cdef` is metadata that says to cffi which symbols you want to use from Python and how to "convert" them in a Python object. With set_source we are giving access to the actual definitions and symbols, so cffi mix this information generating a .c that can be compiled as a Python module.

My idea was to generate the binding without write any additional metadata but as I already said above that is not possible because you don't know what symbols you want to access from Python. I have tried to do some fuzzy^Wnasty logic to collect them but I failed in the attempt...

As many other projects I propose to use a macro something like EXPORT_SYMBOL(symbol_name) and then use this macro to collect the symbols using ctag, pycparser, whatever and build the `cdef` automatically. It will allow us to generate a binding integrated with waf automatically. So what do you think?

Uff I forgot to mention that on this pull request, there is a proof of concept with an example (using loopback, something really simple) where you can see cffi in action.

### DIFF
--- a/bindings/python/cffi_csp_exported_symbols.h
+++ b/bindings/python/cffi_csp_exported_symbols.h
@@ -1,0 +1,150 @@
+enum csp_reserved_ports_e {
+    CSP_CMP,
+    CSP_PING,
+    CSP_PS,
+    CSP_MEMFREE,
+    CSP_REBOOT,
+    CSP_BUF_FREE,
+    CSP_UPTIME,
+    CSP_ANY,
+    CSP_PROMISC,
+};
+
+typedef enum {
+    CSP_PRIO_CRITICAL,
+    CSP_PRIO_HIGH,
+    CSP_PRIO_NORM,
+    CSP_PRIO_LOW,
+} csp_prio_t;
+
+enum {
+    CSP_SO_NONE,
+    CSP_SO_RDPREQ,
+    CSP_SO_RDPPROHIB,
+    CSP_SO_HMACREQ,
+    CSP_SO_HMACPROHIB,
+    CSP_SO_XTEAREQ,
+    CSP_SO_XTEAPROHIB,
+    CSP_SO_CRC32REQ,
+    CSP_SO_CRC32PROHIB,
+    CSP_SO_CONN_LESS,
+    CSP_NODE_MAC,
+    CSP_ROUTE_FIFOS,
+    CSP_RX_QUEUES,
+    CSP_ID_PRIO_SIZE,
+    CSP_ID_HOST_SIZE,
+    CSP_ID_PORT_SIZE,
+    CSP_ID_FLAGS_SIZE,
+    CSP_DEFAULT_ROUTE,
+};
+
+typedef union {
+    uint32_t ext;
+    ...;
+} csp_id_t;
+
+typedef struct {
+    uint16_t length;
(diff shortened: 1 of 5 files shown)


johandc:
I have read about libffi a couple of months ago, but it is the first time i hear about cffi. A lot have happened since we first implemented the ctypes bindings, and i must say that cffi looks very promising.

The problem with ctypes was this interface file. We build that using a program called ctypesgen which now seems to be unmaintained and also lacking the support for packed data structures. So i'm very grateful that you are willing to add this to libcsp.

Just as you are, i'm very focused on making the bindings as automatically as possible. Previously it was a tedious task to update.

Regarding the EXPORT_SYMBOL define: libcsp is generally due for a header file cleanup, back in the days we just threw everything into one big csp.h file that was supposed to be _the_ list of exported symbols. However as time passed by, more and more functions have now been put into separate header files and smaller code modules. So instead of collecting all exported functions in one file, i think that adding the EXPORT SYMBOL define to every function where it's needed is just the right solution.

I think the normal practice is to have the symbols exported in the .c file, and not in the headers.

sorennohr:
Hi agustinhenze.

We are trying to reopen the task of getting some proper bindings for libcsp and in that context I was looking at pull-request.
I am trying to get the example to work, but I am missing the imports in the example (lib and ffi).

Did you forget some files or are these supposed to be created by cffi_gen?

Søren

johandc:
The issues about bindings seems to have finally come to a conclusion with @sorennohr's python module for libcsp. Libffi will therefore not become the official bindigs. Thank you for the pull request to @agustinhenze.

PR #55, opened 2015-07-17
Can multiple interfaces support

### LINKED ISSUE #52: CAN/ driver doesn't receive interface! multiple can interfaces on the same microcontroller not supported.
Hi!, I have implemented our drivers for your can interface, and it is working perfectly, but I have found an issue:
    We have 2 can peripherals in our micro-controller, so we need to create 2 can interfaces, and route them independently to different destinations.
    I think this is happening because all of your hardware nodes only have "one" interface of each type. And yo have never deal with this issue.
    To fix this problem some modifications needs to be done:

1) User needs to be able to create several instances of an interface. Right now interface is just a global variable declared in csp_if_can.c. Each interface would have at list a different name "can0", "can1".

2) "can_send" function needs to receive as parameter a pointer to the interface. And also the tx_callback. Right now I see you are receiving the interface in the csp_can_tx function, but you don't use it for nothing!

3) csp_can_init function, maybe should be split into 2 functions, one to be called only once for initializing queues and rxtask, and other to be called once for each interface....

If you are ok,  I can try to do some refactor to implement this modifications, and that way I will be more sure about all modifications involved in fixing this issue. And the I show you the results. I won't be able to test all the drivers! I will only modify the signature of them.

I would like to know what do you think about this. Maybe you are already working on this...

Best Regards! Nicolas.

nicob87:
Hi! 
This pull request suppose to fix #52 . That is, support multiple can interfaces. In addition to that, add support for multiple instances to the socketcan driver.

As @johandc told me, I have taken a look to the kiss interface, I have seen there you are using the "driver" pointer. I think this case is not exactly the same, because here, we just need to be able to tell the driver "which" interface we want to use, but not necessarily using different drivers. 

My first goal was to ensure that all my work would be "backward" compatible, so if you call the "csp_can_init" function as usual, everything works, and the old global interface is used.

   If you want to use the multiple interfaces support new feature, you call "csp_can_init_ifc" for each interface, rather than "csp_can_init". (see example!).

-Now driver functions: "can_init", "can_send" and callbacks should receive the "interface" as argument.
- I have added support for the multiple instance only for "socketcan" driver, and that is what I have tested in the example. To make this I have connected two of this: "http://www.8devices.com/usb2can" in loopback.
- For all the other drivers ("can_at90can128.c","can_at91sam7a1.c","can_at91sam7a3.c") I have only made the necessary modifications to make them "backward compatible", this is, just passing NULL as argument for the interface pointer in callbacks and functions. I have not been able to test this because I don't have the hardware for that, but, I have compiled them successfully, and I have done the same test with our driver, in "backward compatible" mode and worked just fine. So I think everything is ok.

Please ask me any question you want, and tell me any critique or modification you want me to do. :-)

Thanks for all the work you are sharing here! It is being really helpful to us, and I hope my work would be useful for you too.

Best Regards, Nicolás.

### DIFF
--- a/examples/csp_if_can.c
+++ b/examples/csp_if_can.c
@@ -1,3 +1,23 @@
+/*
+Cubesat Space Protocol - A small network-layer protocol designed for Cubesats
+Copyright (C) 2012 GomSpace ApS (http://www.gomspace.com)
+Copyright (C) 2012 AAUSAT3 Project (http://aausat3.space.aau.dk)
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2.1 of the License, or (at your option) any later version.
+
+This library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with this library; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
 /* This example was tested using two usb2can connected in loopback, in posix.*/
 
 #include <unistd.h>
--- a/examples/csp_if_can.c
+++ b/examples/csp_if_can.c
@@ -40,7 +40,6 @@ void client_server_pseudo_task(void)
 
 
     for (;;) {
-
         if (type == TYPE_CLIENT) {
                 packet = csp_buffer_get(strlen(message0));
                 if (packet) {
@@ -90,10 +89,6 @@ int main(int argc, char **argv)
     if (csp_init(me) != CSP_ERR_NONE || csp_buffer_init(10, 300) != CSP_ERR_NONE) {
         return;
     }
-
-
-    /* last two arguments only matters if first argument is true (single_interface) */
-    csp_can_init(false, 0, NULL);
 
     csp_can_init_ifc(&csp_if_can0, CSP_CAN_MASKED, &can_conf0);
     csp_can_init_ifc(&csp_if_can1, CSP_CAN_MASKED, &can_conf1);
--- a/examples/csp_if_can.c
+++ b/examples/csp_if_can.c
@@ -1,0 +1,110 @@
+/* This example was tested using two usb2can connected in loopback, in posix.*/
+
+#include <unistd.h>
+
+#include <csp/csp.h>
+#include <csp/csp_interface.h>
+#include <csp/arch/csp_clock.h>
+#include <csp/interfaces/csp_if_can.h>
+
+#define TYPE_SERVER 1
+#define TYPE_CLIENT 2
+#define PORT        10
+#define BUF_SIZE    250
+
+char *message0 = "Testing CSP0";
+char *message1 = "Testing CSP1";
+int me, other, type;
+static csp_iface_t csp_if_can0;
+static csp_iface_t csp_if_can1;
+
+void _send(csp_packet_t *packet, csp_iface_t *ifc, char * message)
+{
+    printf("Sending: %s, with %s\r\n", message, ifc->name);
+    strcpy((char *) packet->data, message);
+    packet->length = strlen(message);
+    csp_route_set(other, ifc, CSP_NODE_MAC);
+    csp_sendto(CSP_PRIO_NORM, other, PORT, PORT, CSP_SO_NONE, packet, 1000);
+}
+
+void client_server_pseudo_task(void)
+{
+    csp_socket_t *sock;
+    csp_packet_t *packet0;
+    csp_packet_t *packet1;
+    csp_packet_t *packet;
+    bool toggle_interface = true;
+
+
+    if (type == TYPE_SERVER) {
+        sock = csp_socket(CSP_SO_CONN_LESS);
+        csp_bind(sock, PORT);
+    }
+
+
+    for (;;) {
+
+        if (type == TYPE_CLIENT) {
+                toggle_interface = false;
+                packet0 = csp_buffer_get(strlen(message0));
+                if (packet0) {
+                    _send(packet0, &csp_if_can0, message0);
+                }
+                toggle_interface = true;
+                packet1 = csp_buffer_get(strlen(message1));
+                if (packet1) {
+                    _send(packet1, &csp_if_can1, message1);
+                }
+            sleep(1);
+        } else {
+            packet = csp_recvfrom(sock, 1000);
+            if (packet) {
+                printf("Received: %s\r\n", packet->data);
+                csp_buffer_free(packet);
+            }
+        }
+    }
+}
+
+int main(int argc, char **argv)
+{
+    static unsigned char ucParameterToPass;
+    struct csp_can_config can_conf0 = {.ifc = "can0"};
+    struct csp_can_config can_conf1 = {.ifc = "can1"};
+
+    if (argc != 2) {
+        printf("usage: %s <server/client>\r\n", argv[0]);
+        return -1;
+    }
+
+    /* Set type */
+    if (strcmp(argv[1], "server") == 0) {
+        me = 1;
+        other = 2;
+        type = TYPE_SERVER;
+    } else if (strcmp(argv[1], "client") == 0) {
+        me = 2;
+        other = 1;
+        type = TYPE_CLIENT;
+    } else {
+        printf("Invalid type. Must be either 'server' or 'client'\r\n");
+        return -1;
+    }
+
+    /* Init CSP and CSP buffer system */
+    if (csp_init(me) != CSP_ERR_NONE || csp_buffer_init(10, 300) != CSP_ERR_NONE) {
+        return;
+    }
+
+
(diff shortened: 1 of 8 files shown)


johandc:
Thank you for your addition to libcsp.

We have reviewed the code and we will gladly merge it upstream on one condition; There is a lot of whitespace changes that we think clutters the pull-request. So could we ask you to remove the <tab> to whitespace conversions?

We can see the trouble you have gone through in order to make this. And we would like to make some general changes to CSP in the future which takes multiple interfaces with multiple and different drivers into account. However that is a topic for another day, so we will merge your changes to begin with, and then use that as a starting point of a larger refactoring of the CAN driver and interface code.

nicob87:
Hi guys!
Good to hear your future plan! it sounds great.

Sorry about the whitespaces issue, my bad. I should have been more careful
reviewing my code.

I have dome some new commits to fix this issue. I have just let a fix on
socketcan driver (line 175-178) because it was wrongly indented.

Please let me know if you are ok with this.

Regards! Nicolás

2015-07-27 10:28 GMT-03:00 Johan De Claville Christiansen <
notifications@github.com>:

src/drivers/can/can_socketcan.c
```diff
@@ -191,32 +228,54 @@ int can_mbox_init(void) {
...
 	/* Init mailbox pool semaphore */
-	sem_init(&mbox_sem, 0, 1);
-
+	sem_init(&csi->mbox_sem, 0, 1);
 	return 0;
 
 }
 
-int can_send(can_id_t id, uint8_t data[], uint8_t dlc, CSP_BASE_TYPE * task_woken) {
+can_socket_info_t * find_can_socket_info(csp_iface_t *ifc)
```
johandc: Here you should use the csp_iface_t->driver pointer to refer from interface to private driver data.
This entire function should be unnecessary.

src/drivers/can/can_socketcan.c
```diff
@@ -191,32 +228,54 @@ int can_mbox_init(void) {
...
 
 	int i, found = 0;
 	mbox_t * m;
+	can_socket_info_t *csi;
+	sem_t *mbox_sem;
 
 	if (dlc > 8)
 		return -1;
 
+	csi = find_can_socket_info(csp_if_can);
```
johandc: Calling an O(n) function for each can_send call is expensive.
If you use the interface->driver pointer you will reduce complexity to O(1).

src/drivers/can/can_socketcan.c
```diff
@@ -236,12 +295,48 @@ int can_send(can_id_t id, uint8_t data[], uint8_t dlc, CSP_BASE_TYPE * task_woke
...
+	if(total_interfaces < MAX_SUPPORTED_CAN_INSTANCES) {
+		csi = &can_socket_info[total_interfaces];
+		total_interfaces++;
+		csi->in_use = true;
+		return csi;
+	}
+	return NULL;
+}
+
+static void init_socket_info_in_use(void)
```
johandc: Because the in_use flag is only checked one place (in the find csi function above) you can completely remove this init call.
Also you should remove the in_use field.

src/drivers/can/can_socketcan.c
```diff
@@ -274,20 +369,31 @@ int can_init(uint32_t id, uint32_t mask, can_tx_callback_t atxcb, can_rx_callbac
...
 			return -1;
 		}
 	}
 
+	/* Disable "local loopback of sent frames". See
+	 * "https://www.kernel.org/doc/Documentation/networking/can.txt"
+	 * in case of running multiple csp instances in a single host operating
+	 * system, when one instance "sends" the other instance receives the
+	 * sended frame because of this feature, we don't want this to happen.*/
+	int loopback = 0; /* 0 = disabled, 1 = enabled (default) */
```
johandc: We disagree with the loopback being disabled because we like to have the capability to have multiple csp programs open on the same linux pc.

nicob87:
Hi! good review!
I have done what you asked.
Regarding the "loopback" feature, I don't know how are you planning to use the interface, but, what I was disabling was a feature that makes other sockets connected to the same can interface "receive" what other process has "sent". this means that if you are really doing a hardware loopback, you will rececive every frame "twice". (one due other sockets receive everything that was sended, and the other because of the loopback). And this is the reason why I had to modify the example.
Anyway, we are ok with this modification. It is is the same for us if this feature is disabled or enabled.
Regards!

nicob87:
Hi guys,
    I just want to know what is the status of this pull request.
Best Regards! Nicolas.

src/drivers/can/can.h
```diff
@@ -51,19 +51,25 @@ typedef struct {
...
 	CAN_NO_ERROR = 1,
 } can_error_t;
 
 /** TX Callback function prototype */
-typedef int (*can_tx_callback_t)(can_id_t id, can_error_t error, CSP_BASE_TYPE * task_woken);
+typedef int (*can_tx_callback_t)(csp_iface_t *csp_iface, can_id_t id, can_error_t error, CSP_BASE_TYPE * task_woken);
 
 /** RX Callback function prototype */
-typedef int (*can_rx_callback_t)(can_frame_t * frame, CSP_BASE_TYPE * task_woken);
+typedef int (*can_rx_callback_t)(rx_queue_element_t *e, CSP_BASE_TYPE * task_woken);
```
johandc: Can you please add csp_can_ in front of rx_queue_element_t?

src/drivers/can/can.h
```diff
@@ -51,19 +51,25 @@ typedef struct {
 	};
 } can_frame_t;
 
+/** To support multiple interfaces, rx must know the interface where message arrived */
```
johandc: We would like to have the *interface pointer moved to within the can_frame_t structure. This will simplify a lot of your code to not having to create queue elements to pass around.

src/interfaces/csp_if_can.c
```diff
@@ -457,18 +463,23 @@ int csp_tx_callback(can_id_t canid, can_error_t error, CSP_BASE_TYPE *task_woken
...
 
 	pbuf_element_t *buf;
 	uint8_t offset;
-	
+
+	csp_iface_t *interface;
+
+	interface = csp_iface;
+	if (NULL == csp_iface) {
+		interface = &csp_if_can;
```
johandc: if (frame->interface == NULL) {
    frame->interface = &csp_if_can;
}

...

frame->interface->rx_error++;
etc...

src/drivers/can/can_socketcan.c
```diff
@@ -130,34 +158,40 @@ static void * mbox_tx_thread(void * parameters) {
...
+		nbytes = read(csi->can_socket, frame, sizeof(*frame));
 
-	if (nbytes < 0) {
-		csp_log_error("read: %s", strerror(errno));
-		break;
-	}
+		if (nbytes < 0) {
+			csp_log_error("read: %s", strerror(errno));
+			break;
+		}
```
johandc, line 163: Indentation.

src/drivers/can/can_socketcan.c
```diff
@@ -82,17 +83,44 @@ typedef struct {
...
-/* List of mailboxes */
-static mbox_t mbox[MBOX_NUM];
+typedef struct {
+	int can_socket;         /** SocketCAN socket handle */
+	mbox_t mbox[MBOX_NUM];  /* List of mailboxes */
+	sem_t mbox_sem;         /** Mailbox pool semaphore */
+	csp_iface_t *csp_if_can;
+} can_socket_info_t;
+
+can_socket_info_t can_socket_info[MAX_SUPPORTED_CAN_INSTANCES];
```
johandc: What about putting the iface ptr and the semaphore pointer into the mbox structure itself?
Same argumentation as before with the queue elements.

examples/csp_if_can.c
```diff
@@ -0,0 +1,92 @@
...
+				printf("Received: %s, with %s\r\n", packet->data, csp_interface.name);
+				csp_buffer_free(packet);
+			}
+		}
+	}
+}
+
+int main(int argc, char **argv)
+{
+	static unsigned char ucParameterToPass;
```
johandc: Two things:
1) Naming. We do not use camel case, and variables start with namespace csp_ and module can_
2) This variable is not used. :)

examples/csp_if_can.c
```diff
@@ -0,0 +1,92 @@
+/* This example was tested using two usb2can connected in loopback, in posix.*/
```
johandc, line 21: The file needs the csp copyright header. You can find it in another file :)

src/interfaces/csp_if_can.c
```diff
@@ -738,19 +738,59 @@ int csp_can_init(uint8_t mode, struct csp_can_config *conf) {
 		return CSP_ERR_NOMEM;
 	}
 
+	return CSP_ERR_NONE;
+}
+
+static bool common_resources_already_initialized = false;
```
johandc: Can you move this to the top of the file, and rename it to something a bit shorter? csp_can_common_initialized for example?

nicob87:
Hi! as you will see, I have created a new structure csp_can_socket in socketcan driver. I think that was the best way to implement your recommendation. Let me know if you thought something different.

Regards!

jledet:
Thank you for your contribution! I would like to generalize this approach so all interfaces get support for multiple instances (covered by issue https://github.com/GomSpace/libcsp/issues/61). I'm closing this pull request for now.

PR #59, opened 2015-11-17
Define build script OS env variables before they are used

bfh-reseiwe:
Define build script OS environment variables before they are used, such that LIBS environment variable is correctly set depending on with_os parameter.

Also fixes syntax error in "lib" build parameter for examples.


jledet:
Looks good to me.

Issue #42, opened 2014-11-04
Remove drivers


jledet:
The drivers should not be included in CSP.

Issue #53, opened 2015-07-08
buffer leak in can driver


johandc:
The can driver is leaking packets when the hardware is not responding.

jledet:
Fixed

PR #60, opened 2015-12-03
Refactor CAN

jledet:
This patch set restructures the CAN interface to do away with the strange callback architecture. Instead csp_can_tx simply calls the can_send function in a loop. This work also serves as a preparation to support multiple CAN interfaces.

Implementations of can_send _must_ guarantee that messages are transmitted in the order that can_send is called. However, they are free to use any buffering/FIFOs available in hardware. CAN drivers must also guarantee that csp_can_rx_frame is called in the order that frames are received (i.e. CFP does not support reordered frames).

The patch also removes all other CAN drivers than SocketCAN. Device specific drivers usually require hardware dependent header files for MMIO registers and detailed knowledge of the hardware configuration (e.g. clocks). Maintaining these drivers in CSP has never really been ideal, so lets just remove them. It should be a trivial fix to update them to the new CAN architecture and maintain them outside CSP, if they're still used by someone. The Linux SocketCAN driver is generic and useful as an example of the CAN driver interface, so it's allowed to stay.


jledet:
Fixes #42 and #53.

johandc:
A very nice simplification of the code.

PR #62, opened 2016-01-19
csp: interfaces: can: fixed tx_count overflow in csp_can_tx()

pacheco017:
`tx_count` should be 16 bits long, same as CSP packet length; otherwise, overflow
bugs will happen when sending packets with 256 or more bytes of data.


jledet:
Correct. The MTU is set to 256 bytes, but it will indeed fail for maximum sized packets. 

Thanks!

PR #63, opened 2016-02-11
Move defining ctx.env.OS earlier so it's defined when adding LIBS.

Dolmio:
Also rename undefined variable libs to ctx.env.LIBS in examples part.
Fixes running examples on posix machine


jledet:
I have merged the patch, but without the additional newline in 141. Thanks!

PR #64, opened 2016-02-11
Use bind instead of connect when creating the publisher socket

Dolmio:
I tried to setup two csp nodes communicating through zmq.
I wasn't able to get things working without changing this.
New to zmq so not entirely sure if that's the way how pub
and sub sockets should be started.

### DIFF
--- a/src/interfaces/csp_if_zmqhub.c
+++ b/src/interfaces/csp_if_zmqhub.c
@@ -127,7 +127,7 @@ int csp_zmqhub_init_w_endpoints(char _addr, char * publisher_endpoint,
 	/* Publisher (TX) */
     publisher = zmq_socket(context, ZMQ_PUB);
     assert(publisher);
-    assert(zmq_connect(publisher, publisher_endpoint) == 0);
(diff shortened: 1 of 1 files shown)


pacheco017:
Hi, Dolmio.

The key to use ZMQ **without any modification to current libcsp code**, is to have a "ZMQ Forwarder" process available. Be sure to check http://learning-0mq-with-pyzmq.readthedocs.org/en/latest/pyzmq/devices/forwarder.html

Basically, you need to run `forwarder_device.py` with the same endpoints you pass to `csp_zmqhub_init_w_endpoints`, but taking into account that a PUB socket connects to a SUB socket and so on (check the graph on the link above).

With this architecture, the forwarder must be always available, the endpoints it uses are "well-known", and you can connect as many CSP processes as you wish.

Dolmio:
Hah, thx. No surprise that I was missing something :)

jledet:
Right, you need a proxy service that forwards messages between the connected nodes. Closing this pull request. I've added an issue to move our zmq service to the public tree.

PR #67, opened 2016-03-14
Bugfix: removed memory leak when receiving CSP_PS

magalno:
When running the csp_ps() received packet buffers are not freed, resulting in the system running out of buffers after repeated calls to csp_ps(). 

There is a clean up section at the button of the function. However this only frees packets if they are non-NULL. As the last call to csp_read() always returns NULL when there is no more to receive, the packet buffer fails the if(packet != NULL) test and is never freed. 

Proposed fix:
Free all received buffers after the content has been read.

### DIFF
--- a/include/csp/csp.h
+++ b/include/csp/csp.h
@@ -28,7 +28,7 @@ extern "C" {
 /* Includes */
 #include <stdint.h>
 
-#include <csp/csp_autoconfig.h>
+#include "config/conf_csp.h"
 
 /* CSP includes */
 #include "csp_types.h"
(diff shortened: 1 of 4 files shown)


johandc:
Thank you for reporting this issue.
- The original version of csp_ps() only supported a length up to one packet, so a bug was introduced when we started supporting a longer process list.

Can you make a clean pull-request containing the fix only? We don't wish to merge your csp_nuts.h header into the mainline.

jledet:
Thanks for the bug report. We have fixed this and another bug in the service handler here: https://github.com/GomSpace/libcsp/commit/3863d583f77d9bd07ddbba0e1250c2aed8183a85

PR #73, opened 2016-06-24
csp: drop support for Windows and OS X

jledet:
The Windows and OS X archs have never been primary citizens of CSP. We do not use them at GomSpace and do not have the infrastructure to do proper development and release testing of the archs. I prefer to remove them entirely, rather than having operating systems with unknown status in the
tree.

If you use this code and know it to be working, please let me know.


teijo:
In [Reaktor Space](https://twitter.com/hashtag/ReaktorSpace?src=hash) project, we've built some portable development utilities during the spring that run on OS X and Linux. Those haven't really been used in the past few months with our shift in focus but we'll probably get back to them after summer.

Depending on how it turns out, we might still have use for the OS X capability, enough to keep a fork of the project or maintain it. Can't say for sure yet though...

jledet:
Thanks for the feedback. I'll drop this pull request and try to set up a test of the OSX arch instead. I'm not sure what to do about the Windows port, though. Perhaps just keep it, but mark it as unmaintained.

PR #75, opened 2016-08-18
can: socketcan: enable blocking writes for socketcan

jledet:
By default, SocketCAN interfaces have a txqueue length of 10 frames. If
no buffers are available, writes to the interface returns ENOBUFS
immediately, without blocking. See this thread and others on the
socketcan lists for some background:

http://socket-can.996257.n3.nabble.com/Solving-ENOBUFS-returned-by-write-td2886.html

Our current SocketCAN interface detects the ENOBUFS error and retries
the transmission with a 10 ms delay. This is quite inefficient because
the queue often runs empty while we wait, and as a result our CAN speeds
are much lower than wire speed.

This patch lowers the CAN socket send buffer to the minimum allowed
value (2048), to ensure that the transmissions are blocked at socket
level and not dropped by the interface.

My testing shows a transfer speed increase of 2.5x - 5x in FTP transfers.

Signed-off-by: Jeppe Ledet-Pedersen jlp@gomspace.com

### DIFF
--- a/src/drivers/can/can_socketcan.c
+++ b/src/drivers/can/can_socketcan.c
@@ -97,7 +97,6 @@ static void * socketcan_rx_thread(void * parameters)
 int can_send(can_id_t id, uint8_t data[], uint8_t dlc)
 {
 	struct can_frame frame;
-	int i, tries = 0;
 
 	if (dlc > 8)
 		return -1;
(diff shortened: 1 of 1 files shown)


pacheco017:
Correct me if I'm wrong, but I think that changing only the socket option -without changing the retry logic- would achieve the transfer speed gains you mention.

That would be the "best-case" scenario, ie. no `ENOBUFS`. My worry is that, in a less than ideal scenario, a temporal `ENOBUFS` would abort the transmission of the whole packet. Without changing the retry logic this was graciously handled, instead.

To tell you the truth, I think I cannot fully grasp the linked discussion, but this raises my eyebrows:

> The above described situation is not the only way how can an application get ENOBUFS, (...)

In fact, I think that the 10ms backoff and 1000 retries limit could be improved (maybe the _maximum-retry-timeout_ should be the one the application code passes as `csp_send`'s `timeout`, instead of `1000*10ms`), but I think this is a different issue.

Just my two cents.

pacheco017:
I've tried this PR with [this simple example program](https://gist.github.com/pacheco017/8003647a6fa563b214c627fab23fac66). By running:

```
$ ./test_can_speed server &
$ time ./test_can_speed client
```

on a dev PC with two [usb2can](http://www.8devices.com/products/usb2can)s connected as `can0`&`can1` I actually see a performance regression when applying this patch ("real" time as reported by `time`):

```
76d3f18d6f7c6f3e78397372f8e23379f4a86528 (current master):   ~10 seconds
e1478eb23b83f891fa06c200a038825b2e90179c (socketcan-block):  ~18 seconds
```

johandc:
I have found that the best performance is achieved by increasing the number of TX buffers. Here's my little can-init script:

#!/bin/sh
sudo ip link set dev can0 down
sudo ip link set dev can0 up type can bitrate 1000000 restart-ms 100
sudo ip link set dev can0 txqueuelen 100

restart-ms is also very important in order to handle bus errors.
On the spaceinventor branch i have enabled this by default in the socketcan driver (using libsocketcan).

perwulff:
Since there hasnø't been identified a single "best" solution, this pull request is closed.

PR #77, opened 2016-08-20
CSP_MAX_DELAY & Interrupts handling on POSIX Queues

pacheco017:
Changes:
- Retries on interrupts (`EINTR`) when waiting on queue operations.
- Special handling of `CSP_MAX_DELAY` as timeout of queue operations.

### DIFF
--- a/src/arch/posix/pthread_queue.c
+++ b/src/arch/posix/pthread_queue.c
@@ -104,7 +104,7 @@ static inline int wait_slot_available(pthread_queue_t * queue, struct timespec *
 
 		ret = pthread_cond_timedwait(&(queue->cond_full), &(queue->mutex), ts);
 
-		if (0 != ret) { //TODO: can be EINTR
+		if (0 != ret && errno != EINTR) {
 			return PTHREAD_QUEUE_FULL; //Timeout
 		}
 	}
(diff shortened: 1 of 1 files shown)


johandc (approved):
Looks good.

Seems simple enough to merge into mainline for stability testing before the next release.

jledet:
Looks OK to me as well. I have one minor nitpick. I don't like these types of comparisons:

`if (PTHREAD_QUEUE_OK == ret)`

I know it's an attempt to avoid accidental assignments, but the style differs from the rest of the codebase. This type of error is also caught by gcc with -Wparentheses (included in -Wall).

But I like the fixes, so I agree with Johan that this should be merged. I can add a style cleanup patch later then.

perwulff:
Ok by me. I would prefer to use CLOCK_MONOTONIC, instead of CLOCK_REALTIME, but that may a relic from before it was possible to set the clock on the pthread_condattr_t.

pacheco017:
@perwulff: I also prefered `CLOCK_MONOTONIC` so I had the changes ready on another branch, now merged in this PR (I did a `--ff-only` merge, so github shows https://github.com/libcsp/libcsp/pull/77/commits/94140c30169b4f4e33ca333073326b63da497da0 before your feedback, sorry if that's confusing).

@jledet: I believe that style consistency is very important, so there is no need to postpone the changes. I think https://github.com/libcsp/libcsp/pull/77/commits/b6581d81071063f187fd58fdade72343795d5b9b addresses your comment.

Thanks for the review!

johandc:
Perfect, looks ready for a merge.

PR #78, opened 2016-08-22
Better CAN pbuf cleanup

pacheco017:
For CAN interface, if incoming traffic rate is high (not one second without a packet) and fragment loss is common, current code will leave all _pbufs_ marked as "used" -and thus, unable to receive new CAN packets-.

These changes allow avoiding this situation by using a new _pbufs_ cleanup policy:
- always call `csp_can_pbuf_cleanup()`, whether fragments are received or not; this function now receives a parameter to avoid checking pbufs until some time has passed since last check.
- keep a counter of available pbufs, and force a cleanup (`csp_can_pbuf_cleanup(0)`) when one is requested through `csp_can_pbuf_new()` if no one is free.
- as a last resort, `csp_can_pbuf_new()` frees a random pbuf if none is free even after the forced cleanup (not enabled by default, compile with `-DENABLE_CAN_OOM_RANDOM_FREE`).
- every `csp_if_can.c` `#define` is now overridable with `-D`.

### DIFF
--- a/src/interfaces/csp_if_can.c
+++ b/src/interfaces/csp_if_can.c
@@ -99,6 +99,9 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 /* Buffer element timeout in ms */
 #define PBUF_TIMEOUT_MS		10000
 
+/* Enable freeing a random packet buffer when no one is free */
+//#define ENABLE_CAN_OOM_RANDOM_FREE
+
 /* CFP Frame Types */
 enum cfp_frame_t {
 	CFP_BEGIN = 0,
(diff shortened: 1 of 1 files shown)


johandc:
Hi,

I have been working on a can interface that does not require a task in order to run. In our (space inventor) master branch a greatly simplified can interface can be found. Our implementation trades off a very small CPU overhead in the ISR with an entire task that requires its own stack and everything in order to run. This is a better tradeoff i feel when dealing with small MCU's where memory is limited. The additional CPU cycles in the ISR (in order run pbuf find and memcpy of 8 bytes can frame) will be negligible. Most likely it's not more heavy than copying a frame to the old can rx task queue.

The newer and simpler pbuf implementation is here:
https://github.com/spaceinventor/libcsp/blob/0f51f5a171487b6f4b2884f95def4762a710c591/src/interfaces/csp_if_can_pbuf.c

My suggestion is to close this request and merge the new task-less implementation.

PR #79, opened 2016-09-09
Fix all warnings from -Wmissing-declarations

johandc:
Warn if a global function is defined without a previous declaration.
Do so even if the definition itself provides a prototype.
Use this option to detect global functions that are not declared in
header files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ build
 .waf3*
 waf-*/
 *.o
+*.d
 *.pyc
 .project
 .cproject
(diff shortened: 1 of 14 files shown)


johandc:
I'm cleaning up in old branches and pull requests, and this is so simple that i'm taking the change of merging now.

jledet:
Why is .d files required in .gitignore?

johandc:
When building with Make (for example atmel software framework) it produces .d files next to the object files inside the tree.

Note: I quickly stopped using Make again.

PR #81, opened 2016-10-28
Apply the fixes to the following CVEs:

eyalitki:
CVE 2016-8596, 2016-8597, 2016-8598

### DIFF
--- a/src/csp_route.c
+++ b/src/csp_route.c
@@ -90,6 +90,9 @@ static int csp_route_security_check(uint32_t security_opts, csp_iface_t * interf
 #ifdef CSP_USE_XTEA
 	/* XTEA encrypted packet */
 	if (packet->id.flags & CSP_FXTEA) {
+		/* Make sure that has a nonce */
+		if (packet->length < sizeof(nonce))
+			csp_log_error("Too short packet for XTEA, %u", packet->length);
 		/* Read nonce */
 		uint32_t nonce;
 		memcpy(&nonce, &packet->data[packet->length - sizeof(nonce)], sizeof(nonce));
@@ -115,9 +118,9 @@ static int csp_route_security_check(uint32_t security_opts, csp_iface_t * interf
 
 	/* CRC32 verified packet */
 	if (packet->id.flags & CSP_FCRC32) {
-#ifdef CSP_USE_CRC32
 		if (packet->length < 4)
 			csp_log_error("Too short packet for CRC32, %u", packet->length);
+#ifdef CSP_USE_CRC32
 		/* Verify CRC32 (does not include header for backwards compatability with csp1.x) */
 		if (csp_crc32_verify(packet, false) != 0) {
 			/* Checksum failed */
--- a/src/csp_services.c
+++ b/src/csp_services.c
@@ -158,8 +158,8 @@ void csp_ps(uint8_t node, uint32_t timeout) {
 		if (packet == NULL)
 			break;
 
-		/* We have a reply, add our own NULL char */
-		packet->data[packet->length] = 0;
+		/* We have a reply, enforce that there is a NULL char in it */
+		packet->data[(packet->length - 1) ? packet->length >= 1 : 0] = 0;
 		printf("%s", packet->data);
 
 		/* Each packet from csp_read must to be freed by user */
--- a/src/csp_sfp.c
+++ b/src/csp_sfp.c
@@ -119,12 +119,29 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 			return -1;
 		}
 
+		/* validate the read enough bytes for the header */
+		if (packet->length < sizeof(sfp_header_t)) {
+			csp_debug(CSP_ERROR, "Not enough bytes for the SFP header");
+			return -1;
+		}
+
 		/* Read SFP header */
 		sfp_header_t * sfp_header = csp_sfp_header_remove(packet);
 		sfp_header->offset = csp_ntoh32(sfp_header->offset);
 		sfp_header->totalsize = csp_ntoh32(sfp_header->totalsize);
 
 		csp_debug(CSP_PROTOCOL, "SFP fragment %u/%u", sfp_header->offset + packet->length, sfp_header->totalsize);
+
+		/* make some sanity checks first:
+		 * 1. offset is sane (no overflow)
+		 * 2. Integer-Overflow safe check: 'length' isn't overflowing
+		 */
+		if (sfp_header->offset >= sfp_header->totalsize ||
+			sfp_header->totalsize - sfp_header->offset < packet->length) {
+			csp_debug(CSP_ERROR, "SFP RX overflow: %u, %u, %u", sfp_header->offset, sfp_header->totalsize, packet->length);
+			csp_buffer_free(packet);
+			return -1;		
+		}
 
 		if (sfp_header->offset > last_byte + 1) {
 			csp_debug(CSP_ERROR, "SFP missing %u bytes", sfp_header->offset - last_byte);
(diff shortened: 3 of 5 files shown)


johandc:
Hi, thank you for your contribution.

Security has never been a real concern of libcsp, so it's nice to see somebody pick up on and find these errors. Jeppe and I will review these changes and merge them upstream as soon as possible.

src/csp_services.c
```diff
@@ -158,8 +158,8 @@ void csp_ps(uint8_t node, uint32_t timeout) {
 		if (packet == NULL)
 			break;
 
-		/* We have a reply, add our own NULL char */
-		packet->data[packet->length] = 0;
+		/* We have a reply, enforce that there is a NULL char in it */
+		packet->data[(packet->length - 1) ? packet->length >= 1 : 0] = 0;
```
niklas88, line 162: This line is pretty confusing. So according to [StackOverflow](http://stackoverflow.com/questions/7687403/is-the-true-result-of-or-defined) in C99 packet->length >= 1 is either 0 or 1, so this should never set the last byte to 0. Also why does the original code always write 1 past the end of data[] which this code still does for the packet->length == 0 case?

Shouldn't this be something like

```
if(!packet->length){
   break;
}
packet->data[packet->length-1] = 0;
```

eyalitki: Indeed it seems that I had a typo in this line.
As some projects prefer one-line-checks I tried to write the fix in a single line, and it does seems to be too complicated.
The fix you suggest is much cleaner, and it will do the trick for this specific problem.

Should mention though that this code fix is the minor part of the overall patch, as the rest of the fixes address major security issues (as listed in the above CVE list).

src/csp_route.c
```diff
@@ -90,6 +90,9 @@ static int csp_route_security_check(uint32_t security_opts, csp_iface_t * interf
 #ifdef CSP_USE_XTEA
 	/* XTEA encrypted packet */
 	if (packet->id.flags & CSP_FXTEA) {
+		/* Make sure that has a nonce */
+		if (packet->length < sizeof(nonce))
+			csp_log_error("Too short packet for XTEA, %u", packet->length);
```
jledet, line 95: Agree with this, but we should return CSP_ERR_XTEA and increment autherr here.

src/csp_route.c
```diff
@@ -115,9 +118,9 @@ static int csp_route_security_check(uint32_t security_opts, csp_iface_t * interf
 
 	/* CRC32 verified packet */
 	if (packet->id.flags & CSP_FCRC32) {
-#ifdef CSP_USE_CRC32
 		if (packet->length < 4)
 			csp_log_error("Too short packet for CRC32, %u", packet->length);
+#ifdef CSP_USE_CRC32
```
jledet, line 123: Agree, we need a check for length >= 4 if CSP_FCRC32 is set. If not we should drop the packet (and not just print an error message).

src/csp_services.c
jledet: Agree with @niklas88 code.

src/interfaces/csp_if_can.c
```diff
@@ -320,6 +320,13 @@ static int csp_can_process_frame(can_frame_t *frame)
 		buf->packet->id.ext = csp_ntoh32(buf->packet->id.ext);
 		memcpy(&(buf->packet->length), frame->data + sizeof(csp_id_t), sizeof(uint16_t));
 		buf->packet->length = csp_ntoh16(buf->packet->length);
+		/* Check 'length' field doesn't overflow the RX buffer */
+		if (buf->packet->length > CSP_CAN_MTU) {
+			csp_log_error("CAN packet is too big for MTU");
+			csp_can_pbuf_free(buf);
+			csp_if_can.frame++;
+			break;
+		}
```
jledet, line 329: Agree.

src/interfaces/csp_if_zmqhub.c
```diff
@@ -88,6 +88,14 @@ CSP_DEFINE_TASK(csp_zmqhub_task) {
...
 		}
 
+		/* makre sure there is no overflow */
+		if (datalen > 256) {
+			csp_log_warn("ZMQ: Too long datalen: %u", datalen);
+			while(zmq_msg_recv(&msg, subscriber, ZMQ_NOBLOCK) > 0)
+			zmq_msg_close(&msg);			
+			continue;
+		}
+
```
jledet, line 98: Also agree with this check, but I think it should be merged with the check for short frames (datalen < 5). Since datalen includes the 4 byte CSP header and a 1 byte "satid", the maximum length is also 261 bytes I think the ZMQ interface could use a cleanup and a comment to explain the protocol.

@johandc, is the `while(zmq_msg_recv(&msg, subscriber, ZMQ_NOBLOCK) > 0)` needed? It is calling `zmq_msg_close()` in a loop, but it is difficult to tell whether this is intentional or not as `zmq_msg_close()` is not indented. If the CSP packet allocation fails, `zmq_msg_close()` is just called directly.

src/csp_sfp.c
```diff
@@ -119,13 +119,30 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 			return -1;
 		}
 
+		/* validate the read enough bytes for the header */
+		if (packet->length < sizeof(sfp_header_t)) {
+			csp_debug(CSP_ERROR, "Not enough bytes for the SFP header");
+			return -1;
+		}
+
```
jledet, line 127: Agree.

src/csp_sfp.c
```diff
@@ -134,20 +151,26 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
...
+		if (*dataout == NULL) {
 			*dataout = csp_malloc(sfp_header->totalsize);
+			*datasize = sfp_header->totalsize;
+		}	
 		if (*dataout == NULL) {
 			csp_debug(CSP_ERROR, "No dyn-memory for SFP fragment");
 			csp_buffer_free(packet);
 			return -1;
 		}
+		/* Comapre the data output to be consistent */
```
jledet, line 164: Compare

src/csp_sfp.c
```diff
@@ -134,20 +151,26 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
...
+		else if ( *datasize != sfp_header->totalsize) {
```
jledet, line 165: Leading space in `if()`

src/csp_sfp.c
```diff
@@ -134,20 +151,26 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
...
+			csp_buffer_free(packet);
+			return -1;			
+		}
 
-		/* Copy data to output */
-		*datasize = sfp_header->totalsize;
 		memcpy(*dataout + sfp_header->offset, packet->data, packet->length);
 
-		if (sfp_header->offset + packet->length >= sfp_header->totalsize) {
+		if (sfp_header->offset + packet->length == sfp_header->totalsize) {
```
jledet, line 173: Why not keep the `>=`?

jledet (changes requested):
I have added a couple of review comments.

src/csp_sfp.c
perwulff: Agree, keep it.

src/interfaces/csp_if_zmqhub.c
johandc: The reason why it is there in the first place is that if (datalen < 5) there must be some kind of syncronization error. The only way to recover is to empty the entire buffer and hope that is itself will not create a new syncronization error, which is actually likely.

As Jeppe may remember this became a notorious problem on GOMX-3's ground station. Somehow data got stuck in the stream and never emptied out. This meant that sometimes packets were delayed by one. Which gave some pretty weird operating behaviour. For example. A ping message may go through only after the second ping is transmitted.

I don't understand why libzmq did this in the first place. I belive they had some bug that was fixed at a point, perhaps just not on our old groundstation.

Anyways. As you say we should group both min and max length checks into one check. Usually if this happens it's because of some kind of protocol sync error.

perwulff:
Again, thank you for your contribution.
 All the issues and a number of additional issues have been resolved in #129

PR #82, opened 2016-12-09
Minor fixes for usart_linux driver

sorennohr:
added support for two new baudrates (3500000 and 4000000) + usart_init() now defaults baudrate to 500000 instead of 0

### DIFF
--- a/src/drivers/usart/usart_linux.c
+++ b/src/drivers/usart/usart_linux.c
@@ -44,6 +44,7 @@ int getbaud(int fd) {
 	int inputSpeed = -1;
 	speed_t baudRate;
 	tcgetattr(fd, &termAttr);
+
 	/* Get the input speed. */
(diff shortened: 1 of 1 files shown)


kristianbay (approved):
Looks OK by me.
Only comment is, that indentation is a mix of tabs and spaces, not easy to read/compare 
Consider changing the indentation
NB: Please note that you are not on the head of master...

jledet:
Looks good apart from the indentation mix Kristian mentioned. While we're adding baudrates, can we these as well just for the sake of completeness? I believe they're the last ones missing from termbits.h:

B230400
B576000
B1152000

sorennohr:
@jledet, @kristianbay: I agree on the indention issues. Actually the indention was screwed in us art_init() and I was trying to fix that. Might not have been to succesfull.

I will add the rates @jledet requested and try to fix the indention issues.

I will open a new pull request for this.

PR #84, opened 2017-01-29
added missing baudrates + usart_init() now defaults to a baudrate of …

sorennohr:
…500000 instead of 0


sorennohr:
Duplicate :-)

PR #87, opened 2017-02-07
Csperf

johandc:
Hi i'm making a pull request in order to discuss the contents of this branch and maybe get it ready for a merge.


johandc:
Here's what i like:
1) The common terminal color function
2) The removal of cpu_set_reset_cause (this was an old compace specific thing and should be handled by a hook instead)
3) Of course i like the whole csperf utility idea. I haven't tested it, but the idea seems cool.
4) It may be that it's not completely finished, but i would rather include a 90% finished tool and see some usage of it, and maybe some contributions in the future, than i would leave it dangling on an old branch that nobody would see. This is too good to be forgotten about :)

What i don't like:
1) Somehow you have the "csp: drop support for Windows and OS X" commit merged into this branch. I don't know if it's because csperf would only build for linux, but we need to remove that commit befor merging.
2) Oh and yes, there's a minor conflict in the usart driver that will be easy to fix.

jledet:
I have squashed and split the csperf commits out into its own branch here:

https://github.com/satlab/libcsp/tree/csperf

I'll try to clean it up a bit, test that it still works, and make it ready for merge. I think the biggest blocker right now is the lack of a common usleep function for other archs than Linux, but I will try to come up with a solution for that.

The other commits (sans `csp: drop support for Windows and OS X` which has been dropped) have been moved to my `for-upstream` branch. I'll send a PR for those in the coming week.

https://github.com/satlab/libcsp/tree/for-upstream

I think we should close this PR and remove the upstream `csperf` branch.

perwulff:
@jledet  how come support for Windows and MacOS are dropped? Not used, difficult to maintain?

johandc:
@perwulff 
We do not have any maintainer on Windows. Which means that it has been completely unused and untested for 5 years now. So it was time to drop it.

However, we recently got some help with testing our POSIX arch on OS X where it seems to work fine. There was a few bits about the included serial driver, but i think that's it :)

perwulff:
We need to keep Windows and Mac support, so the pull request needs to be updated.
The pull request will be closed, until resources are available to do the update.

PR #89, opened 2017-02-07
Space Inventor: Cleanups

johandc:
Hi,

I have gathered a lot minor commits into one big pull-request.

Please note the following things:
  added csp_hex_dump()
  moved csp/crypto headers into include/csp/crypto - I have moved these so drivers written outside the libcsp source tree can benefit from these functions.
  added a new clock_get_time64 interface to get nanosecond clock (this is used for better timesync feature comming in future csp versions)
  Note: the wscript now uses the freertos instead of the freertos_h include (this may break build for gomspace?)
  Note: I have disabled the cflags from the wscript also. I can't remember right now what they were needed for, but they conflicted with our libasf (which defines the cflags for the chip).


johandc:
@perwulff Can we get some comments on this, in particular:


> Note: the wscript now uses the freertos instead of the freertos_h include (this may break build for gomspace?)
> Note: I have disabled the cflags from the wscript also. I can't remember right now what they were needed for, but they conflicted with our libasf (which defines the cflags for the chip).

I know you mentioned a way for `waf` to identify the top level, but maybe that could be added in a future commit?

If i don't hear any objections, i will merge this soon. We need to stabilize and think of our due release.

jledet (approved):

jledet:
Woops, didn't mean to approve everything...

include/csp/csp.h
```diff
@@ -499,6 +499,11 @@ int csp_conn_print_table_str(char * str_buf, int str_size);
  */
 void csp_buffer_print_table(void);
 
+/**
+ * Hex dump to stdout
+ */
+void csp_hex_dump(const char *desc, void *addr, int len);
+
```
jledet, line 506: Is this used anywhere in CSP?

src/csp_hex_dump.c
```diff
@@ -30,7 +30,7 @@ void csp_hex_dump (const char *desc, void *addr, int len)
 				printf ("  %s\r\n", buff);
 
 			// Output the offset.
-			printf ("  %04x ", (int) addr + i);
+			printf ("  %p ", addr + i);
```
jledet: This could be merged with the previous commit.

wscript
```diff
@@ -107,8 +107,8 @@ def configure(ctx):
         ctx.env.FEATURES += ['cstlib']
 
     # Setup CFLAGS
-    if (len(ctx.env.CFLAGS) == 0):
-        ctx.env.prepend_value('CFLAGS', ['-Os','-Wall', '-g', '-std=gnu99'])
+    #if (len(ctx.env.CFLAGS) == 0):
+    #    ctx.env.prepend_value('CFLAGS', ['-Os','-Wall', '-g', '-std=gnu99'])
```
jledet: Please add an explanation in the commit message of why this is necessary. And if it is necessary, why don't we just remove those lines?

src/rtable/csp_rtable_cidr.c
```diff
@@ -114,7 +114,7 @@ static int csp_rtable_parse(char * buffer, int dry_run) {
 
 	while ((str) && (strlen(str) > 1)) {
 		int address = 0, netmask = 0, mac = 255;
-		char name[100] = {};
+		char name[10] = {};
```
jledet, line 117: Ack. Perhaps we should just change the if name field to be a char[10] and add a define for the maximum length.

wscript
```diff
@@ -260,7 +260,7 @@ def build(ctx):
         target = 'csp',
         includes= ctx.env.INCLUDES_CSP,
         export_includes = ctx.env.INCLUDES_CSP,
-        use = 'include freertos_h',
+        use = 'include freertos',
```
jledet: Again, a comment in the commit message of why this is necessary (and how it affects users) would be nice.

src/rtable/csp_rtable_cidr.c
```diff
@@ -229,4 +230,4 @@ void csp_rtable_print(void) {
 	}
 
 }
-
+#endif
```
jledet, line 233: Ack.

src/crypto/csp_hmac.c
```diff
@@ -27,8 +27,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 /* CSP includes */
 #include <csp/csp.h>
 
-#include "csp_hmac.h"
-#include "csp_sha1.h"
+#include <csp/crypto/csp_hmac.h>
+#include <csp/crypto/csp_sha1.h>
```
jledet, line 31: What is the usecase for this? I guess this is okay, but if we export it we should be aware that it becomes part of our API.

include/csp/crypto/csp_hmac.h
```diff
@@ -46,6 +46,25 @@ int csp_hmac_append(csp_packet_t * packet, bool include_header);
...
+ */
+int csp_hmac_memory(const uint8_t * key, uint32_t keylen, const uint8_t * data, uint32_t datalen, uint8_t * hmac);
+
+/**
+ * Save a copy of the key string for use by the append/verify functions
+ * @param key HMAC key
+ * @param keylen HMAC key length
+ */
+int csp_hmac_set_key(char * key, uint32_t keylen);
+
```
jledet: Return values? Otherwise OK.

src/csp_route.c
```diff
@@ -198,11 +198,16 @@ int csp_route_work(uint32_t timeout) {
...
+		input.interface->drop++;
 		csp_buffer_free(packet);
 		return 0;
 	}
 #endif
 
+	/* Now we count the message (since its deduplicated) */
+	input.interface->rx++;
+	input.interface->rxbytes += packet->length;
+
```
jledet, line 210: ACK.

include/csp/arch/csp_clock.h
```diff
@@ -33,10 +33,12 @@ typedef struct {
...
-__attribute__((weak))
-extern void clock_get_time(csp_timestamp_t * time);
-__attribute__((weak))
-extern void clock_set_time(csp_timestamp_t * time);
+__attribute__((weak)) extern void clock_get_time(csp_timestamp_t * time);
+__attribute__((weak)) extern void clock_set_time(csp_timestamp_t * time);
+
+/* Additional functions for 64-bit nanosecond type */
+__attribute__((weak)) extern uint64_t clock_get_time64(void);
+__attribute__((weak)) extern void clock_set_time64(uint64_t time);
```
jledet: If this is not used in CSP yet, I don't think we should add it. AFAIK we also have a wish to get rid of the weak symbols, so I don't think we should add new ones until we have a solution for the "system interface".

jledet (dismissed):

src/csp_services.c
```diff
@@ -107,7 +107,7 @@ void csp_ping_noreply(uint8_t node) {
 	packet->data[0] = 0x55;
 	packet->length = 1;
 
-	printf("Ping ignore reply node %"PRIu8".\r\n", node);
+	printf("Ping ignore reply node %u.\r\n", (unsigned int) node);
```
perwulff, line 110: Personally I'm not a big fan of casts, as they tend to hide problems. I thought PRIxx macros was part of the Cxx standard, so it seems strange that they have been dropped.

wscript
perwulff: It would be ok with me, to remove them completely - as they are normally set on a higher level. However, we should agree on a minimum warning level.
GOMspace currently uses "-Wall", "-Wextra", "-Wshadow", "-Wcast-align", "-Wwrite-strings", "-Wno-unused-parameter" as minimum, so I would suggest using at least these values.

Instead of removing them (easy way of defining minimum warning level), the if statement could be changed to
    if len(ctx.stack_path) <= 1:
which should only be true, if the build is done directly in the csp folder. If libcsp as part of a bigger project, the length will be bigger.

wscript
perwulff: Yes, it will break GOMspace builds , but that is ok if there is a rational reason behind it :) - a comment would be good.

include/csp/arch/csp_clock.h
perwulff: I agree on removing use of weak symbols.
Generally I agree on not adding stuff, until it is required - it is easier to get the interface correct, when there is a use case. I would prefer the naming of the function to indicate, what it returns, e.g. clock_get_nsec().
If it wasn't for compatibility, I would prefer that the functions were prefixed with csp, because clock_get_time() is very close to clock_gettime (C standard).

wscript
johandc: The reason why i dind't remove it was that i knew it would break builds for libcsp.so (in bindings and such). As we agreed we do want to move more towards supporting this, i therefor think that ctx.stack_path is a perfect solution.
:+1:

src/rtable/csp_rtable_cidr.c
johandc: Hmm, that might be a good idea:

Here's what we have today:

interface name: CAN0 (5 bytes)
csp_iface_t->name pointer (4 bytes)
Total size = 9 bytes.

What you are suggesting
Interface name = 10 bytes
end of story.

For static allocation this won't be a big overhead and for dynamic allocation, the user saves a call to strdup or malloc in order to preserve the string in memory. So i like it. (Not that we have any dynamically allocated interfaces yet ).

include/csp/csp.h
johandc: The hex dump is usefull for dumping csp frames in debug messages. It's a simple utility function, and if you don't use it, it won't be linked with your code. So it does not add any overhead as such.
I know that gomspace have their hexdump in a libutil, and that's okay, but i just belive that a few helper functions for this could be usefull to ship with csp.

Maybe we should make it even better, and call it csp_buffer_dump() and print the entire contents of the padding bytes, the csp header and the data field?

wscript
johandc: I'll change this back since we have moved to meson now.

src/csp_services.c
johandc: PRIu8 is not suppported by the iprintf function in the newest version of newlib nano.
I belive this is a problem, but i have no idea how we would fix this in a pretty fashion?

src/crypto/csp_hmac.c
johandc: The reason is that we use the HMAC functions in one of our CSP interfaces to add Layer-2 HMAC over the radio-link.

johandc:
I have reverted the wscript, added some return value documentation requested by jeppe, and removed the 64-bit clock type.

Issue #61, opened 2015-12-07
Support multiple instances of interfaces


jledet:
We generally only support a single instance of each interface type, e.g. only one CAN interface because the csp_iface_t structs are statically defined. The KISS interface has some code to support multiple UARTs, but it would be nice to generalize this to cover all interface types. It should be possible to have both multiple instances of the same type (e.g. two SocketCAN interfaces) and of the same type but with different drivers (e.g. a platform I2C controller and an external controller).

johandc:
We now have a design guideline for this see pull-request #90

johandc:
Fixed in #90

PR #90, opened 2017-02-07
Space Inventor: Multican

johandc:
This is another attempt at solving #61

I have cleared the following design choices with jeppe:
  1) There should no longer be an interface init function (csp_can_init for example)
  2) Drivers should have their own init functions (csp_socketcan_init for example)
  3) Drivers should allocate their own csp_iface_t structure (either statically like i have chosen to for the first version of socketcan, or using malloc like a future version of socketcan with multiple instance support could have)
  4) The former csp_can_rx_task have been removed and the former csp_can_process_frame function have been rename to csp_can_rx and given the option to run directory from an ISR. This will be the best solution on some MCU's  where a reduced memory usage is traded off for a little longer ISR execution. For MCU's like the AVR8 where added ISR time cannot be accepted, it is still possible to use the csp_can_rx_task but it must be implemented in the driver.
  5) The pbuf code have been simplified and put into another c file to help readability of the code.


jledet:
Can we perhaps split this in multiple pull requests and squash some of the commits (at least the fixes)? It's quite difficult to track what has changed across 18 commits. A couple of comments below:

1/2/3: Yes, I agree that we should "reverse the call chain", so we have driver specific init functions that allocate the interface struct and use the interface "driver" pointer to hold driver specific data.

4: I understand the motivation for doing the frame assembly directly in the ISR to save memory, but I think it is difficult to select a "best" solution across all use cases. Some users might prefer having as short ISRs as possible to avoid blocking other IRQs (especially if you don't have nested interrupts).

I think I would prefer rewriting csp_can_rx to be able to run in interrupt and task context (like your commit), so people have the choice, but keeping the csp_can_rx_task as the default option. This would also allow existing users to keep their code.

Another option would be to try moving the frame processing to the router task (perhaps a generic driver callback scheduling?), but this is probably a larger change.

johandc:
Regarding point 1/2/3: Okay, so we are agreed on the future interface design guidelines now. This will be added to our wiki here on github.

Regarding point 4: We agree to leave deferred frame processing as an optional add-on. Possibly using some kind of ressource sharing with the router task to make the best use of the memory in the system. For now, it's going to be up to each CAN driver to choose, and since can drivers are out of CSP scope there will be no new configure option for this. @jledet i could bring back the helper functions for the old can_rx_task in a separate file like csp_if_can_task.c or csp_if_can_deferred.c?

I will squash the commits as well. I'm thinking of merging this week if there are no further comments.

perwulff:
I have no specific comments to the implementation at this stage, at least nothing we didn't already cover on the last meeting. For future changes of this size (for any size actually), I would really like to see some automatic testing to back up the implementation.

johandc:
I have done a lot of testing on this patch already, but I can't say that it was automatic.

If CSP should have a test suite we should make some kind of hardware simulator. Putting real hardware into the test would be a much bigger endavour.

perwulff:
@johandc, I could really use the improvements done for reducing memory footprintf in the CAN driver. Any chance we could merge this into the master branch very soon?

PR #91, opened 2017-02-13
Fix SocketCAN driver includes

mndza:
bits/socket.h should never be included directly, and only sys/socket.h must be used (which is already included).

This change fixes build with musl libc.


johandc:
Thank you for the contribution. I have tested without bits/socket.h and it works fine here. 
On what platform are you using the musl libc?

mndza:
You're welcome.
I am using it on a Xilinx Zynq running Linux built with a custom toolchain.

Issue #70, opened 2016-04-14
Proper python bindings


jledet:
As correctly pointed out by issue #54, we need a better way of generating the bindings. Something with support for multiple languages would be nice, but python bindings should be our main focus.

johandc:
@perwulff @sorennohr I have marked this as comming in v1.5. From what i've seen in your pr #92 it lookeds pretty solid already. So i was hoping you could make it?

perwulff:
Yes, we are working on it internally and once we are satisfied, it will be pushed.

johandc:
Hi Per,

What's the status on that python module?

perwulff:
It is not ready to be merged into 1.5. Best guess at the moment, is within a month - maybe sooner.

smorad:
Has there been any update with this? Ctypesgen does not work with python3.

sorennohr:
See https://github.com/libcsp/libcsp/pull/112 @smorad

PR #92, opened 2017-03-03
added bindings source + a simple example

sorennohr:
Hi Guys,

Inspired by MTA's work in pyparam I have created this, still WIP, set of python bindings for CSP. For our ground station software we would really much like to have an approach to bridging the gap between c-libraries and python code.

I know we have had a number of other approaches but I think this one has a few advantages;

* it hides csp implementation details by using Py_Capsules to make structs opaque
* it is an integrated part of the libcsp.so shared object file

A drawback is that it is not auto-generated.

The example can be run like this;

./waf configure build --enable-bindings --enable-if-zmqhub --enable-crc32
PYTHONPATH=build python examples/python_bindings_example.py

This pull-request is as such not ready for integration, but should more be seen as an suggestion to how we could approach the tasks of creating bindings for our c-libraries in general.

What say you?

Søren

### DIFF
--- a/examples/python_bindings_example.py
+++ b/examples/python_bindings_example.py
@@ -1,0 +1,40 @@
+#!/usr/bin/python
+
+import sys
+import os
+import time
+
+import libcsp
+
+if __name__ == "__main__":
+
+    # init csp
+    libcsp.csp_buffer_init(10, 300);
+    libcsp.csp_init(27);
+    libcsp.csp_zmqhub_init(27, "localhost")
+    libcsp.csp_rtable_set(7, 5, libcsp.csp_zmqhub_if(), libcsp.CSP_NODE_MAC)
+    libcsp.csp_route_start_task(1000, 0)
+    time.sleep(1) # allow router startup
+    libcsp.csp_rtable_print()
+
+    print "pinging addr 7, rc=" + str(libcsp.csp_ping(7, 5000, 10))
+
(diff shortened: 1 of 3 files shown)


johandc:
Overall it looks good. I like the idea of using a proper Python extension. Maybe it could be built using setup.py and installed using pip like a 'real' python module?

I think that the Proper bindings issue #70 have been open for too long without a real solution yet. It seems that getting good bindings requires some manual effort. The ctypes method was manual and an attempt to wrap the C calling convention. A python extension is sill manual, but it's written in C and it's standard. The auto-generated bindings worked well, but the tool is no longer maintained.

I see this as a good direction to go and it doesn't look too difficult to maintain, especially since it's written in C and can be built together with the rest of the source. This way we will remember to fix stuff when we change the API because the build will fail immediately, instead of having to wait for some python runtime error.

A slight disadvantage is that it will have to be rewritten for python 3 (i think there's some bigger changes to the init code, but not to the methods).

sorennohr:
Hi Johan,

Thanks for you feedback.

Actually my intent was to raise this pull-request on the GomSpace internal fork of libcsp, but I must have mistakenly selected the libcsp/libcsp for as a base.

I will close the pull-request here for now and hopefully raise a more complete pull-request with the bindings in the near future when we have agreed upon this approach internally.

Thanks,

Søren

PR #98, opened 2017-08-04
Remove compile-time configuration options

### LINKED ISSUE #93: Remove compile-time configurations
**We seriously have too many options by now.**

A lot of them are coming from the fact that i like static allocations and CSP was made that way. However, a problem with them is that they need to be pre-defined, which leads to all of these weird `CSP_SOMETHING_MAX` defines. So eventually we have shifted to have dynamic allocation for queues but we still have weird compile time defines such as `CSP_CONN_QUEUE_LENGTH` to set that up. 

Our `csp_init()` call today is a sorry excuse for itself. It takes one argument only (the address). CSP init is doing multiple things which relies on compile-time settings. However we also have a set of functions, get/set hostname, model, revision and even a get/setter for the address. Furthermore the buffer system is inited separately by csp_buffer_init(). This is confusing...

So yes, i think there is room for improvement here...

johandc:
Fixes #93 

Configuration struct:
I have created a configuration struct. It's not opaque, we need to discuss this. I don't believe it has to be opaque necessarily. 

The configuration struct is copied on csp_init, and shared to all csp modules through a global csp_conf symbol declared in csp_init.h. We could perhaps eliminate the use of a global if we wish. It would require internal functions in csp_conn_innit() and friends to take more arguments.

Allocations:
Generally i have replaced all static allocations with one-time calls to calloc/malloc.

Note: This was just my quick fix for this. I believe it could be made prettier with some love. Also i think we should update documentation and such.


src/csp_init.c
```diff
@@ -0,0 +1,70 @@
...
+csp_conf_t csp_conf;
+
+uint8_t csp_get_address(void)
+{
+	return csp_conf.address;
+}
+
+int csp_init(csp_conf_t * conf) {
+
+	/* Make a copy of the configuration
```
perwulff: Please make the argument const, so an actual const/pre-filled struct can be parse in.
What about making the variable csp_conf static and add a method for getting it, i.e. const * csp_conf_t csp_get_conf(). This way, the application can also retrieve the information, e.g. 'revision'. And it will prevent externals from changing the content, by simply adding a local "extern csp_conf_t csp_conf" in their code.

wscript
```diff
@@ -199,13 +199,6 @@ def configure(ctx):
...
     ctx.define_cond('CSP_USE_INIT_SHUTDOWN', ctx.options.enable_init_shutdown)
-    ctx.define('CSP_CONN_MAX', ctx.options.with_max_connections)
-    ctx.define('CSP_CONN_QUEUE_LENGTH', ctx.options.with_conn_queue_length)
-    ctx.define('CSP_FIFO_INPUT', ctx.options.with_router_queue_length)
-    ctx.define('CSP_MAX_BIND_PORT', ctx.options.with_max_bind_port)
-    ctx.define('CSP_RDP_MAX_WINDOW', ctx.options.with_rdp_max_window)
-    ctx.define('CSP_PADDING_BYTES', ctx.options.with_padding)
-    ctx.define('CSP_CONNECTION_SO', ctx.options.with_connection_so)
     
     if ctx.options.with_bufalign != None:
```
perwulff: Consider removing this option and always align buffer(s). It might take a little bit more memory, but makes things simpler/safer.

perwulff (approved):
Looks good to me, only a few comments.
These changes, along with the previous 1.5 (route/CAN) changes, breaks the API - so should the next release be 2.0? Or is a major version reserved for something not compatible on the wire.

johandc:
I think the major version should correspond to the protocol header, not the API. So all csp 1.x is compatible and so forth.
In almost every release we have had minor API changes, this however affects everybody. But considering we only release once every year or two, i think it would be okay to go with just version 1.5.

wscript
johandc: Yes. I will have to look into that.

src/csp_init.c
johandc: Ok.

perwulff:
The data structure should be static, not const.

static csp_conf_t csp_conf;

// get default shouldn't touch the actual csp_conf data
csp_conf_get_default(csp_conf_t * conf)
{
   conf->xx = something;
}

int csp_init(const csp_conf_t * conf)
{
   csp_conf = *conf; // make copy
}

const csp_conf_t * csp_conf_get(void)  // const provides read access, but prevents/warn about write access
{
   return &csp_conf;
}

This provides support for better module programming, where one part of the software can initialize csp with address, model, revision etc. and another (generic part) can read for some purpose, printf("revision: %s", csp_conf_get()->revision).


________________________________

johandc:
Yes, disregard that reply. I found out what you meant after sending it, so i changed it back on github, but it still arrived at your inbox....

I have factored in your comments in two small commits.

The internal variable in csp is still not static, so it can be shared with other internal csp modules through csp_init.h - I think that's just easier as long as we ensure csp_init.h is not exported.

wscript
johandc: I have removed the configure option. But the code in csp_buffer will still look for the compile variable, or use sizeof(int *) as default.

PR #103, opened 2017-08-17
Fix a waste of memory

Delbringer:
On POSIX systems, a memory space (288 bytes for Linux) is reserved by each threads to return informations while they terminate. This memory space is not freed until pthread_join() is called on them and return. As CSP does not include any function to join a thread, this is a memory waste at the best, or a memory leak if threads are started and stopped as a regular operation. Furthermore, Valgrind detect this issue as a memory leak and prevent users to reach the "valgrind-clean" status.
Setting the thread as detached does not allocate this memory space and avoid any trouble.
At the same time, we can reuse the attribute structure to take in consideration the stack size provided by users.

### DIFF
--- a/src/arch/posix/csp_thread.c
+++ b/src/arch/posix/csp_thread.c
@@ -27,5 +27,28 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/arch/csp_thread.h>
 
 int csp_thread_create(csp_thread_return_t (* routine)(void *), const char * const thread_name, unsigned short stack_depth, void * parameters, unsigned int priority, csp_thread_handle_t * handle) {
-	return pthread_create(handle, NULL, routine, parameters);
+	pthread_attr_t attributes, *attr_ref;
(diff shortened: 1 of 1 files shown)


johandc:
Thanks for the patch.

This will come in very handy especially on embedded linux systems where memory might be of a concern.

I assume you have performed some testing of this already, so we will merge this into release-1.5 branch for further testing before the next release.

Delbringer:
In the last patch I used the PTHREAD_STACK_MIN macro which should be included by pthread.h but got moved on newer versions into limits.h, sorry for that. Here is the quick-fix.
[quick-fix.txt](https://github.com/libcsp/libcsp/files/1234332/quick-fix.txt)

johandc:
Yes, i found it and have included limits.h already. I also corrected the attr_ref thing. So we should be good 👍

Delbringer:
Look like a sync issue between my IDE and the file on an overloaded hard drive, which must have rolled back the & character. I missed this line when giving you the quickfix. Look like you got my back, thanks! 
I hope it will be usefull !

Issue #96, opened 2017-06-02
Possible socket allocator memory leak


bandren:
In the function [```csp_socket()```](https://github.com/libcsp/libcsp/blob/17bbef37142d67ebb8bf87044296bf6a79d3a799/src/csp_io.c#L173-L185) of the current commit in master at the time of writing this.

A socket is allocated using the ```csp_conn_allocate( )```. Then, if connectionless, a queue is allocated for the socket using ```csp_queue_create( )```. If the queue failed to allocate then the function returns without freeing the socket allocation.

I'm not sure if this is the intended functionality. If it is, just delete this issue.

perwulff:
It seems to be a bug, as the reference to 'sock' is lost (not returned). Will look into that.

Thanks for the input.

perwulff:
Fixed in branch release-1.5.

PR #106, opened 2017-08-23
Fixed possible socket allocator memory leak (#96), if csp_socket() fails.

perwulff:
Fixed general socket allocator memory leak by csp_socket(CSP_SO_CONN_LESS) -> csp_close().


perwulff:
In order to test these fixes, the unit test needed access to the internal csp_conn_t array in csp_conn.c, in order to verify things have been closed/opened.
I added csp_conn_get_array() in the internal csp_conn.[ch], and included this header in the unit test. There are other ways of making it testable, this is just the simplest I could think of.

johandc (approved):

PR #110, opened 2017-09-27
Improvements to SFP.

mtl:
This commit addresses the following issues:

1. sfp_send_\*() leaves the connection with CSP_FFRAG set, which results in malformed packets (FRAG flag set but no sfp_header in the packet) if the connection is later reused for non-SFP traffic.
2. Resolves ambiguity over whether the caller to sfp_recv_\*() must invoke csp_free() on dataout when the return value is -1 and dataout is non-NULL.
3. Resolves ambiguity over whether csp_buffer_free() has been called on the packet when the return value is -1. Note that this may cause a problem if the caller expects to perform subsequent diagnostics on the invalid packet. However, other failures do free the packet, so without this change the caller would not know the reason for the error (all failures return -1) and thus whether the packet has been freed. If diagnostics are required in this case (i.e., sfp_recv_\*() called on a packet with CSP_FFRAG clear), the caller can check for CSP_FFRAG prior to invoking sfp_recv_\*() and avoid the error.

### DIFF
--- a/src/csp_sfp.c
+++ b/src/csp_sfp.c
@@ -81,8 +81,10 @@ int csp_sfp_send_own_memcpy(csp_conn_t * conn, void * data, int totalsize, int m
 		/* Send data */
 		if (!csp_send(conn, packet, timeout)) {
 			csp_buffer_free(packet);
+			conn->idout.flags &= ~CSP_FFRAG; /* Clear fragment flag */
 			return -1;
 		}
+		conn->idout.flags &= ~CSP_FFRAG; /* Clear fragment flag */
 
 		/* Increment count */
 		count += size;
@@ -101,6 +103,8 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 
 	unsigned int last_byte = 0;
 
+	*dataout = NULL; /* Allow caller to assume csp_free() can always be called when dataout is non-NULL */
+
 	/* Get first packet from user, or from connection */
 	csp_packet_t * packet = NULL;
 	if (first_packet == NULL) {
@@ -116,6 +120,7 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 		/* Check that SFP header is present */
 		if ((packet->id.flags & CSP_FFRAG) == 0) {
 			csp_debug(CSP_ERROR, "Missing SFP header");
+			csp_buffer_free(packet); /* Always free the packet */
 			return -1;
 		}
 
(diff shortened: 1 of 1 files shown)


src/csp_sfp.c
```diff
@@ -81,8 +81,10 @@ int csp_sfp_send_own_memcpy(csp_conn_t * conn, void * data, int totalsize, int m
 		/* Send data */
 		if (!csp_send(conn, packet, timeout)) {
 			csp_buffer_free(packet);
+			conn->idout.flags &= ~CSP_FFRAG; /* Clear fragment flag */
 			return -1;
 		}
+		conn->idout.flags &= ~CSP_FFRAG; /* Clear fragment flag */
```
johandc: I feel like this is something best solved in csp_conn_new()

src/csp_sfp.c
```diff
@@ -101,6 +103,8 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 
 	unsigned int last_byte = 0;
 
+	*dataout = NULL; /* Allow caller to assume csp_free() can always be called when dataout is non-NULL */
```
johandc, line 104: agreed

src/csp_sfp.c
```diff
@@ -116,6 +120,7 @@ int csp_sfp_recv_fp(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t
 		/* Check that SFP header is present */
 		if ((packet->id.flags & CSP_FFRAG) == 0) {
 			csp_debug(CSP_ERROR, "Missing SFP header");
+			csp_buffer_free(packet); /* Always free the packet */
```
johandc: This looks like a regular memory leak, thanks for finding this. However we could do without the comment.

johandc (approved):

src/csp_sfp.c
johandc: In fact, after checking, i can see that this is handled in csp_conn_new, or rather in the csp_connect call. 

So the answer would be, that do not reuse a connection. It's better to call csp_close() on the SFP connection and csp_connect() to open up a new one.

So my revised answer is, that this line should be omitted, and the connection reuse should be changed/fixed instead.

johandc (changes requested):
I was too quick to approve, after looking at it again, i would like the change in csp_sfp.c to be revised.

src/csp_sfp.c
mtl: Ok, this change is backed out in 2a7b3114d03906d673082ea28eba98677a295d50.

src/csp_sfp.c
mtl: Comment removed in 2a7b3114d03906d673082ea28eba98677a295d50. Thanks.

johandc (approved):

PR #115, opened 2018-02-01
Fixes regression from 6a3174 ' Fix issue where ack is not sent when d…

Dolmio:
…uplicate packet

arrives and delayed acks is disabled.'
where duplicate acks are being sent when delayed_acks is false,
because csp_rdp_check_ack is called both from csp_read and csp_rdp_new_packet
and csp_rdp_check_ack will always return true.


Dolmio:
I'm also wondering whether it's desirable to have the csp_rdp_check_ack at all in the csp_io? It was added there in d6aa309c

johandc:
The original argumentation for adding the check in csp_read is still valid. We want to send and ack as soon as possible when the RX buffer is cleared.

PR #117, opened 2018-08-21
Upstream 1.5

johandc:
Dear Everybody,

Here is a collection of single commits that can all be individually reviewed. They have accumulated over the past 1.5 years on our master branch. Most of them relate to simple things like compiler warnings and api calls in upgrading FreeRTOS to the v8/v9 API.

More noteably, the verbosity of some of the debug have been reduced. And a new callback have been added to for the purpose of connectionless and task-free sockets.

### DIFF
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -113,6 +113,11 @@ csp_socket_t * csp_socket(uint32_t opts) {
 
 }
 
+void csp_socket_set_callback(csp_socket_t * socket, void (*callback)(csp_packet_t * packet)) {
+	socket->opts |= (CSP_SO_CONN_LESS | CSP_SO_CONN_LESS_CALLBACK);
+	socket->socket = callback;
+}
+
 csp_conn_t * csp_accept(csp_socket_t * sock, uint32_t timeout) {
 
 	if (sock == NULL)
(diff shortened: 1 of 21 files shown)


src/csp_io.c
```diff
@@ -113,6 +113,11 @@ csp_socket_t * csp_socket(uint32_t opts) {
 
 }
 
+void csp_socket_set_callback(csp_socket_t * socket, void (*callback)(csp_packet_t * packet)) {
+	socket->opts |= (CSP_SO_CONN_LESS | CSP_SO_CONN_LESS_CALLBACK);
+	socket->socket = callback;
```
perwulff, line 118: Ugly hack :) - socket->socket should have been changed to an union. If thats too much work, at least make a prototype (typedef) for the callback, which the socket can be casted to. Doesn't make much prettier, but more readable.

perwulff (approved):
Please remember to update public API documentation.

johandc:
What is the reason for the close? @perwulff

src/csp_io.c
johandc: As it's internal I don't see a problem in sharing the socket pointer, which normally would point to a queue, as long as the callback flag is set. The two are mutually exclusive. 

Adding a union would make it a little more readable. In that case I suggest calling them queue and callback, instead. Also adding the typedef could help on the API readability.

perwulff:
I didn't close it per say, it was closed when I deleted the release-1.5 branch. We get support questions about the release-1.5 branch, whether it is an official release or not. And it is not, e.g. CHANGELOG not updated, known bugs, etc. - so it seemed simpler to delete it and avoid the confusion.
We had the discussion about version 1.5 versus 1.6 a long time ago, where we agreed the next official version would be 1.6 - that's when the develop-.1.6 came about.
develop-1.6 is based on release-1.5, so the pull request can be done on develop-1.6 instead.
GomSpace will be pushing a number of pull requests very soon, with the hope of eventually merging develop-1.6 into master - hopefully before christmas :)

PR #119, opened 2018-08-29
Bug-fixes and improvements.

### LINKED ISSUE #94: Unused functions and variable - remove?
These functions and variable are not used in libcsp. Is there a greater scheme - or simply remove them?

file: libcsp/include/csp/drivers/usart.h: 
int usart_messages_waiting(int handle);
static inline int usart_stdio_msgwaiting(void) {
	return usart_messages_waiting(0);
}

file: libcsp/src/drivers/usart/usart_linux.c
int usart_stdio_id = 0;

perwulff:
First of, sorry for this being so "big" - but its actually small changes/fixes. There is no real code change, except for some additional 'const' and the fixes for the broken bindings/examples.

Most of the changes are documentation, and there will be more of that - as we trying to document all public APIs, using Doxygen.

Fixes #94.

This will be merged in on Friday the 31st, unless I hear otherwise :) 

Changelog:
Fixed compile issues in Python bindings and examples, after introducing new init/configuration concept.
Consistently use csp_qfifo_write(), instead of a mix between csp_new_packet() and csp_qfifo_write().
Updated documentation.
Removed unused options from wscript, after introducing new init/configuration concept.
csp_conf_get_defaults() ensure all members are initialized (prevent valgrind errors).
Made API arguments 'const', where pointers are read-only.
Added csp_transaction2(), supporting socket/connection options.
Removed unused functions/defines: csp_new_packet, usart_stdio_msgwaiting(), csp_route_get_mac()
Added new Python bindings: csp_send, csp_sendto, csp_connec( rtable_check, rtable_load, peek, pooke, clock
Added internal csp_qfifo_wake_up() for testing purposes.
Added examples/buildall.py for verifying build with different options.

### DIFF
--- a/doc/interfaces.rst
+++ b/doc/interfaces.rst
@@ -44,20 +44,20 @@ The interface also needs to receive incoming packets and pass it to the CSP prot
        csp_packet_t *buf = csp_buffer_get(BUF_SIZE);
        /* Wait for packet on fifo */
        while (read(rx_channel, &buf->length, BUF_SIZE) > 0) {
-           csp_new_packet(buf, &csp_if_fifo, NULL);
+           csp_qfifo_write(buf, &csp_if_fifo, NULL);
            buf = csp_buffer_get(BUF_SIZE);
        }
    }
 
-A new CSP buffer is preallocated with csp_buffer_get(). When data is received, the packet is passed to CSP using `csp_new_packet()` and a new buffer is allocated for the next packet. In addition to the received packet, `csp_new_packet()` takes two additional arguments:
+A new CSP buffer is preallocated with csp_buffer_get(). When data is received, the packet is passed to CSP using `csp_qfifo_write()` and a new buffer is allocated for the next packet. In addition to the received packet, `csp_qfifo_write()` takes two additional arguments:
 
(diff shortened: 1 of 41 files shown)


sorennohr (approved):
Looks great!

johandc:
This looks very good indeed.

One thing though. Is it on purpose that you removed csp_transaction? I think that is a quite frequently used part of the external API.

perwulff:
csp_transaction() has been changed to a static inline function in the header file csp.h - so its all compatible, no need for code change - but do require a re-compilation.

johandc:
Okay i missed that. I would have done the same, so I agree on that. 

What is the reasoning behind the naming? 

The only times I have seen the usage of numbers in function names were if you wanted to break a public function but ensure everybody updates their code accordingly. And then to switch back after a few revisions.

Otherwise, why don't we name transaction2 a little bit better, now that it becomes a part of the public API? 

Something like csp_transaction_w_opts()

perwulff:
Renamed csp_transaction2() to csp_transaction_w_opts().
Branch merged into release-1.6.

PR #123, opened 2019-03-04
Update Waf to 2.0.14

daniellandau:
Old Waf doesn't work with Python 3.7 because of PEP 479.

https://stackoverflow.com/a/51701040/1517818


perwulff:
hi Daniel,
Thanks for the contribution.
Only the self-extracting Waf file should be added to GiT - not all the extracted files.

daniellandau:
Found it, now fixed.

johandc:
We should target this on the release branch instead.

daniellandau:
@johandc I clicked the "Allow edits from maintainers" so feel free to change the base branch.

PR #124, opened 2019-03-06
fixed bug that checked for negative length of unsigned int

glennlive:
This may not ever cause an issue on real serial ports, but when using UARTS over TCP connections, it can certainly caused errors. When checking for negative numbers, we should use signed integers.


perwulff:
Thanks for the contribution.

PR #127, opened 2019-05-07
Fixed compilation on Windows and Mingw64

mandrek44:
"interface" is defined in mingw headers as struct, because of that compilation fails: https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/rpc.h#l12

The RTL_CONDITION_VARIABLE_INIT may be already defined and thus gcc will fail.

Additionaly, when linking shared lib or executable with the libcsp under Wondows, the symbol `csp_sys_tasklist_size` is not defined.


perwulff:
Thank you for the contribution.

PR #129, opened 2019-12-13
Fixed buffer overrun issues in HMAC, XTEA, CRC32 and SFP.

perwulff:
Cleaned up interfacing between CSP and the HMAC, XTEAm CRC32 and SFP.
Removed checks on packet pointers, as they can never be zero.
Improved checking/validation in csp_sfp.c


johandc (approved):
Looks good.

PR #131, opened 2019-12-16
Refactored rtable CIDR/static implementation.

perwulff:
Changes are backward compatible, except for load/save format for "static" routes (no longer binary -> and thereby decoupled from the actual code).

 - common code moved to src/rtable/csp_rtable.c
 - use same format for storing table (text) - no binary format for static (no pointers saved).
 - added csp_rtable_route_t holds route entry.
 - support for simple format without mask, e.g. "20 CAN" (instead of 20/5 CAN)

csp_rtable_route_t will be used in optimizing route look up, to avoid double lookup when using MAC (via) address.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 libcsp 1.6, DD-MM-YYYY
 ----------------------
 - refactored rtable CIDR/static impl., e.g. use same format for storing table (text). csp_rtable_route_t (new) holds route entry.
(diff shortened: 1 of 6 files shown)


perwulff:
I can see the diff seems big, perhaps I can move some code around to make it smaller.
Let me know, if that is prefered.

johandc:
What this does, is really adopting the CIDR notation for both static and CIDR cases. So now the difference is just wether it's stores in a linked list or written out into a full static routing table. I like it.

johandc (approved):

PR #132, opened 2019-12-16
Align internal timeouts.

perwulff:
conn_lock and sport_lock uses different timeout values. These are simple internal locking, and should never fail, because they protect a very short piece of code
I would prefer no timeout at all, as they will never timeout anyway, and if they did, things will be more confusing. So we need to decide on a number :) 
I picked CSP_MAX_TIMEOUT (closet to no timeout).

CSP_SO_INTERNAL_LISTEN (internal flag), allowing things to be freed again. csp_rtable_free() forgotten in previous commit. Both are mainly for testing.

### DIFF
--- a/src/csp_conn.c
+++ b/src/csp_conn.c
@@ -187,7 +187,7 @@ csp_conn_t * csp_conn_allocate(csp_conn_type_t type) {
 
 	static uint8_t csp_conn_last_given = 0;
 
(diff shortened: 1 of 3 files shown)


johandc (approved):
I generally don't like infinite spinlocks, because it tends to silently hang threads. But since we are semi sure this would never happen in this case, we might as well use max timeout as we could use 100ms.

PR #133, opened 2019-12-16
Refactored all interfaces:

perwulff:
- accept csp_rtable_route_t, instead of csp_iface_t.
- no static members -> multiple interfaces of all types.
  - added csp_iface_t.interface_data for holding interface data.
  - added csp_iface_t.driver_data for holding driver data (unknown to interface level).
- checks for buffer overrun
- Generic template "int csp_<interface>_add_interface(csp_iface_t * iface)" for adding interface, with pre-allocated interface/driver data.
- set MTU if not already set, and it make sense on the respective interface.
- Driver Tx function is now a callback, must be set by the application in the interface data.

csp_packet_t no longer packed, instead padding is increased from 8 to 10 bytes to make it correctly aligned. Removing packed improces performance and reduces code on certain architectures. CSP_STATIC_ASSERT(...) is used to verify correct alignmemt, where csp_packet_t padding is used.

ZMQhub:
- support for multiple rx-filter(s), what addresses to receive (instead of just the local CSP address).
- support for transfering MAC (via address), used by csp_bridge.

I2C:
- driver layer sort of removed, no callback back into application for setting up I2C.

CAN/socketcan:
- split open of CAN device and adding CSP interface - both for reuse and improved error handling.
- added convenience function csp_can_socketcan_open_and_add_interface() -> open and add interface.

KISS/usart:
- split open of uart device and adding CSP interface - both for reuse and improved error handling.
- added convenience function usart_open_and_add_kiss_interface() -> open and add interface.

### DIFF
--- a/src/csp_bridge.c
+++ b/src/csp_bridge.c
@@ -27,22 +27,34 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include "csp_io.h"
 #include "csp_promisc.h"
 
-static csp_iface_t* if_a = NULL;
-static csp_iface_t* if_b = NULL;
+typedef struct {
+    csp_iface_t* iface;
+    bool is_zmq;
+} bridge_interface_t;
+
+static bridge_interface_t if_a;
+static bridge_interface_t if_b;
+
+static uint8_t get_mac(bridge_interface_t * iface, const csp_packet_t * packet) {
+
+	if (iface->is_zmq) {
+		return ((const csp_zmqhub_csp_packet_t *)packet)->mac;
+	}
+	return CSP_NODE_MAC;
+}
 
 static CSP_DEFINE_TASK(csp_bridge) {
-
-	csp_qfifo_t input;
-	csp_packet_t * packet;
 
 	/* Here there be bridging */
 	while (1) {
 
 		/* Get next packet to route */
-		if (csp_qfifo_read(&input) != CSP_ERR_NONE)
+		csp_qfifo_t input;
+		if (csp_qfifo_read(&input) != CSP_ERR_NONE) {
 			continue;
+		}
 
-		packet = input.packet;
+		csp_packet_t * packet = input.packet;
 
 		csp_log_packet("Input: Src %u, Dst %u, Dport %u, Sport %u, Pri %u, Flags 0x%02X, Size %"PRIu16,
 				packet->id.src, packet->id.dst, packet->id.dport,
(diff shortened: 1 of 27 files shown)


src/csp_bridge.c
```diff
@@ -54,33 +66,46 @@ static CSP_DEFINE_TASK(csp_bridge) {
...
 	return CSP_TASK_RETURN;
 
 }
 
+static bool is_zmq_interface(const char * ifname)
+{
+	if (ifname) {
+		// if the interface contains zmq, we assume it is a ZMQ interface
+		for (unsigned int i = 0; i < strlen(ifname); ++i) {
+                    if (strncasecmp(&ifname[i], "zmq", 3) == 0 ) {
```
johandc: This is not looking good. We can do better.

johandc (changes requested):
I think we might be making a mistake here. I'll explain a bit, and then you can tell me if you follow me or not.

I fist realised that this have become a problem now that we are passing routes, instead of interface pointers around, and then suddenly there is a function called get_mac, that uses string comparison, for each packet.

So point number 1. Is that CSP routing table should have never had any idea about the MAC address. It is a routing table, it should have absolutely no knowledge about lower layers, and is it a hack that I misused that in the first place.
2. What we really need is a real next hop address field. Like you have on hour ip route table on your Linux PC. So for example you probably have a default route 0.0.0.0/0 via 192.168.0.1 or something. This is what we need for CSP. The notion of MAC address should simply be replaced with next hop address. Which is a valid CSP address.

3. When sending a packet, the route lookup should return the route entry, and the tx function should then generate a layer two packet to either the next hop address, or the destination address (if on local segment). 

4. The way the interface translates from CSP node to Layer 2 address should be left only to the interface to describe. And here we have easy rules for all interfaces. So no ARP tables are required.

Can you follow this reasoning?

The changes would mostly be around the naming, bringing that more into line with ip routing table terminology. And then this ugly get_mac thing found in csp_bridge can go away, because that will be done on the zmqhub tx function.

src/csp_bridge.c
perwulff: First of, there is no string comparison for each packet sent - it is only done once, when the bridge is started.
Secondly, because we use csp_bridge for bridging between two zmqhubs, the mac/via information is needed to be transfered, because 1) there is no routing table on the bridge, and 2) we do not want a routing table on the bridge (too complicated).
I will remove the csp_bridge stuff and handling of via information from the public libcsp.

I'm not sure I follow completely, what are you proposed solution? Are talking about simply changing the naming "mac" to something else, or are we talking about a more complicated setup, going back to multiple lookup's for sending a single packet?

perwulff:
Removed transfer of MAC/via in csp_if_zmqhub and csp_bridge.
However, did keep the bridge_interface_t in csp_bridge, to make it easier (reduce differences), IF "someone" needs to store more data :)

perwulff:
@johandc  please review again.

johandc (approved):

PR #135, opened 2019-12-31
Fixed memory leak corner case in csp_sfp_recv_fp()

jsacane:
If a transient error caused the CSP_FFRAG bit to be unset in an SFP packet,
the caller would have no way to free the buffer.


perwulff:
hi, thank you for your input. This issue (and more severe issues) has been fixed in #129, mered into the develop branch.
The develop branch is on its way to become the next officiel release 1.6 (merged into master), hopefully during january.

PR #137, opened 2020-01-15
Use Travis CI for continuous build

perwulff:
Biggest change: Renamed 'interface' to 'iface', to avoid compile problems on Windows - without having to '#undef interface'. The word 'iface' was already used in several other places.

The CI.rst holds links to Travis-CO - perhaps this should be moved to README.rst ? But it may make it  more difficult to "include" README.rst.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 # Travis CI
 
-branches: [develop-travis]
 language: c
 
 jobs:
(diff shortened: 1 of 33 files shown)


johandc (approved):
Is interface a reserved word with Travis?

perwulff:
The problem with "interface" is only on Windows. It probably always been there since Windows was introduced. The previous work-around was to undef interface:
#include <Windows.h>
#undef interface
So Windows has something named interface, probably a macro, which causes problems.
But doing #undef is not so nice, especially because it was also done in public header, thus having an impact on whatever included our headers.

PR #139, opened 2020-01-18
Fix compiler warnings

tniessen:
These functions never return, and should be marked as such. This fixes corresponding compiler warnings.

### DIFF
--- a/src/csp_bridge.c
+++ b/src/csp_bridge.c
@@ -29,7 +29,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 static csp_iface_t* if_a = NULL;
 static csp_iface_t* if_b = NULL;
 
-static CSP_DEFINE_TASK(csp_bridge) {
+static __attribute__((noreturn)) CSP_DEFINE_TASK(csp_bridge) {
 
 	csp_qfifo_t input;
 	csp_packet_t * packet;
--- a/src/csp_route.c
+++ b/src/csp_route.c
@@ -313,7 +313,7 @@ int csp_route_work(uint32_t timeout) {
 	return 0;
 }
 
-static CSP_DEFINE_TASK(csp_task_router) {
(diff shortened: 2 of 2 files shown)


perwulff:
Thanks Tobias. Can I ask what compiler you are using and what warnings you have set/enabled?

tniessen:
I am using gcc 7.4.0 on x86_64-linux-gnu with no custom `CFLAGS`, so it should use `-Wall`. This is the output without the patch:

```
../src/csp_bridge.c: In function ‘csp_bridge’:
../src/csp_bridge.c:74:1: warning: no return statement in function returning non-void [-Wreturn-type]
 }
 ^

../src/csp_route.c: In function ‘csp_task_router’:
../src/csp_route.c:323:1: warning: no return statement in function returning non-void [-Wreturn-type]
 }
 ^
```

perwulff:
Ok, just realized that the master branch is missing:
return CSP_TASK_RETURN;
compared to the develop branch, where I don't see this warning.
The develop will soon be merged into the master branch, as version 1.6.

tniessen:
Ohh I didn't know there was a develop branch. Should PRs target that branch instead of master? If so, it might be worth mentioning that somewhere :)

perwulff:
You are absolutely right, that pull request should be targeted at the master branch - but unfortunately is pretty old at the moment :)
The develop branch is only used for ongoing merge work, before going to the master branch as an official release.
I think we will go with what we have in the develop branch for now (the added return statement).
If you build the develop branch, the warning should be gone.
The develop branch do have a couple of other warnings on Mac and Windows, we fixing at the moment.

tniessen:
Okay, thank you for the explanation.

tniessen:
@perwulff @johandc Do you still accept PRs for this project? I'd like to work on some issues, but I'm not sure whether to target the master branch or the develop branch, and whether in this project or in https://github.com/spaceinventor/libcsp. Also, is there a timeline for a v2.0 release? Thank you!

PR #145, opened 2020-01-25
Develop

jlangvand:
Waf would not throw an error, but simply ignore missing python-dev-packages, even "enable bindings" attributes were provided. Marking them as mandatory fixes this.

It was unclear that --enable-bindings was required for --enable-python3-bindings to work. I renamed --enable-bindings to --enable-python2-bindings and made --enable-python3-bindings work independently from --enable-python2-bindings. If this has possible side effects, please let me know.

### DIFF
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -25,7 +25,7 @@ options += [
 waf = ['./waf']
 if os in ['posix']:
     options += [
-        '--enable-bindings',
+        '--enable-python2-bindings',
         '--enable-python3-bindings',
         '--enable-can-socketcan',
         '--with-driver-usart=linux',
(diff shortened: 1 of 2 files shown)


perwulff:
Thanks for the contribution.
I'm thinking of dropping Python2 support entirely - what do you think?

jlangvand:
Yeah, I agree. As I see it, the problem with the code as it is, is that it is unclear that you need to specify --enable-bindings for --enable-python3-bindings to work. Also, when the python-dev packages are not specified as mandatory, the script will just silently ignore the python flags if the packages are missing (this happened to me, and it took a while for me to understand what was going on).

Removing python2 might be a good alternative, but I would recommend setting the required python3-dev package as mandatory :)

perwulff:
Python2 has been removed from libcsp, and all required packages are "mandatory" now.
Documentation needs to be updated in the develop branch.
Thanks for your input.

PR #147, opened 2020-02-06
Updated init, interfaces and examples

perwulff:
First of, sorry for the number of commits in a single request - but all changes ties into the example code.
Please disregard Python bindings and documentation, will come in a later update.

- Moved csp_buffer_init() into csp_init().
- ZMQ interface:
  - Added parameter for returning the added csp_iface_t
  - Added 'flags' parameter for future support of setting options on a ZMQ connection, e.g. alive.
- Removed unused rtable functions: csp_rtable_find_iface()/csp_rtable_finc_mac()
- Scoped usart with csp and consolidated code for Windows and Linux implementation.
- Updated example code
  - Updated example code to a single csp_server_client.c implementation, which can be run standalone or as separate server and client node, communicating through zmqproxy.
  - Added examples/csp_arch.c, which runs a number of simple test on most cross-platform APIs. Travis runs the same code on LInux, Mac and Windows.


src/bindings/python/pycsp.c
```diff
@@ -35,7 +35,7 @@ static PyObject* pycsp_service_handler(PyObject *self, PyObject *args) {
     PyObject* conn_capsule;
     PyObject* packet_capsule;
     if (!PyArg_ParseTuple(args, "OO", &conn_capsule, &packet_capsule)) {
-        return NULL; // TypeError is thrown
+        return NULL; // TypeError is thrownfcsp_init
```
johandc: Typo

src/csp_buffer.c
```diff
@@ -52,23 +48,21 @@ CSP_STATIC_ASSERT(offsetof(csp_packet_t, length) == 10, length_field_misaligned)
...
 
-	data_size = _data_size;
-	buffer_size = data_size + CSP_BUFFER_PACKET_OVERHEAD;
 
 	// calculate total size and ensure correct alignent (int *) for buffers
-	const unsigned int skbfsize = CSP_BUFFER_ALIGN * ((sizeof(csp_skbf_t) + buffer_size + (CSP_BUFFER_ALIGN - 1)) / CSP_BUFFER_ALIGN);
+	const unsigned int skbfsize = CSP_BUFFER_ALIGN * ((sizeof(csp_skbf_t) + csp_buffer_size() + (CSP_BUFFER_ALIGN - 1)) / CSP_BUFFER_ALIGN);
 
-	csp_buffer_pool = csp_calloc(buf_count, skbfsize);
+	csp_buffer_pool = csp_calloc(csp_conf.buffers, skbfsize);
```
johandc: I think we want to use malloc. Not calloc.
All buffers are considered dirty when gotten.
The extra cost of clearing out all buffers can increase boot time on small systems

johandc (changes requested):

src/bindings/python/pycsp.c
perwulff: Fixed for now - a larger binding update will follow soon.

perwulff: Changed to malloc - and set ref_count to1 instead of ++.

src/csp_buffer.c
perwulff: Fixed for now - a larger binding update will follow soon.

johandc (approved):

PR #148, opened 2020-02-11
Removed timeout from internal send functions and CSP interface

perwulff:
The send timeout was only used in the I2C CSP interface, where it was passed to the driver layer - unless it is a RDP connection, in which case the connection timeout is used.
The timeout is kept in the public APIs (compatibility), but in the future ignored and documented as "unused".
Because the the next_hop/tx function already had been broken (csp_iface_t -> csp_route_t), the timeout parameter was removed.

Hardcoded timeouts in the CSP interfaces aligned to 1000 mS (ranged from 1000 to infinite).

Major update of public header documentation, alignment on terms used, timeout description, etc.

Moved csp_promisc prototypes to own header file: csp/csp_promisc.h.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 libcsp 1.6, DD-MM-YYYY
 ----------------------
+- Removed timeout for send (internally only) and interface tx functions - the timeout was only used on I2C interface.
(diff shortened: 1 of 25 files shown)


johandc (approved):
Looks good

PR #167, opened 2020-08-19
Add support for RIOT OS

millotp:
Added [RIOT OS](https://github.com/RIOT-OS/RIOT) to libcsp, similar to [FreeRTOS](https://www.freertos.org/), only for the CAN bus.

Here is the list of modification:
- Add `riotos` to the waf build system, and in the examples
- Add `src/arch/riotos`
- Rename `src/drivers/can/can_socketcan.c` to `src/drivers/can/can_linux.c` to support other can drivers
- Add `src/drivers/can/can_riotos.c`
- Few changes to make libcsp compatible with arm-none-eabi-gcc

Now the driver can be selected at build time with the option `--with-driver-can=<linux | riotos>`

The implementation of the semaphore is not perfect, it does not fully pass csp_arch but csp_server_client works fine.


src/csp_iflist.c
```diff
@@ -21,7 +21,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/csp_iflist.h>
 
 #include <stdio.h>
-#include <string.h>
+#include <strings.h>
```
johandc, line 25: Why not use the more portable string.h?

src/rtable/csp_rtable.c
```diff
@@ -28,6 +28,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include "../csp_init.h"
 
+#include <strings.h>
+
```
johandc, line 32: Again, why is this needed? We build with that toolchain too, and have never needed anything from strings.h

wscript
```diff
@@ -24,7 +24,7 @@ import os
 APPNAME = 'libcsp'
 VERSION = '1.6'
 
-valid_os = ['posix', 'windows', 'freertos', 'macosx']
+valid_os = ['posix', 'windows', 'freertos', 'macosx', 'riotos']
```
johandc, line 27: CSP will also support the meson build system in v2.0 but it should be very easy to add to that too.

src/drivers/can/can_riotos.c
```diff
@@ -0,0 +1,189 @@
...
+#include <csp/drivers/can_socketcan.h>
+
+#include <stdlib.h>
+#include <csp/csp.h>
+#include <csp/arch/csp_thread.h>
+
+#include "can/can.h"
+#include "can/can_trx.h"
+#include "can/conn/raw.h"
+#include "tja1042.h"
```
johandc, line 30: Who provides this header?

johandc (changes requested):
Hi, this looks fine.
Im pussled over the usage of stings.h
I don't know how I feel about changing the build configuration and the socketcan rename. It would break a lot of Linux builds for no reason other than a name change I think.
Maybe it was cleaner if each driver would just have their own option to enable or disable it. 
Finally CSP 2.0 is coming out. We may consider merging the riotos stuff for that one.
You can try it out (see the 2.0 pull request)

johandc:
We did away with platform.h

src/csp_iflist.c
millotp: You are right we can use string.h, and when using arm-none-eabi-gcc, the flag `_POSIX_C_SOURCE=200809L` must be set before the inclusion of string.h, otherwise function like `strnlen`, `strtok_r` and `strcasecmp` cannot be found.  
Should the macro be define in `src/csp_iflist.c` and `src/rtable/csp_rtable.c` or passed as an argument to gcc ?

src/drivers/can/can_riotos.c
millotp: They are provided by RIOT directly, [here](https://github.com/RIOT-OS/RIOT/blob/master/sys/include/can/can.h).    
So --with-driver-can=riotos can only be compiled within riot os

wscript
```diff
@@ -57,7 +57,7 @@ def options(ctx):
 
     # Drivers and interfaces (requires external dependencies)
     gr.add_option('--enable-if-zmqhub', action='store_true', help='Enable ZMQ interface')
-    gr.add_option('--with-driver-can', default=None, metavar='DRIVER', help='Enable can driver [socket, riotos]')
+    gr.add_option('--with-driver-can', default=None, metavar='DRIVER', help='Enable can driver [linux, riotos]')
```
yashi, line 60: Please keep `meson.build` sync'ed if you are requesting merging into develop branch.

yashi:
@millotp, please add some commit messages to each commit.
https://github.com/erlang/otp/wiki/writing-good-commit-messages

johandc:
Are you still interested in getting this in?

PR #187, opened 2021-02-15
Allow unknown transaction response size in python bindings

Wint3rmute:
When using python bindings, there is no way to perform a `csp_transaction` without knowing the response size beforehand. This patch adds an optional argument, to `pycsp_transaction` that allows it to accept responses of any size.


johandc:
Thanks for your contribution.

If I understand correctly, this is an optional argument, and so it should not break the python API?

johandc (approved):

Wint3rmute:
@johandc that's correct, the argument is optional. If not supplied, the API works in the same way as before

Thanks for merging!

PR #205, opened 2021-06-11
ability to toggle promiscuous mode of CAN driver

arrooney:


johandc:
Any reason why you closed this again?

arrooney:
> Any reason why you closed this again?

Only because I haven't tested these changes, and probably won't have time to for a little while, but if this is actually something that seems useful, then I can run some manual tests to make sure this does what I wanted it to.

johandc:
I couldn't see anything wrong with it. So can merge if you want. We are targeting develop branch anyways.

src/drivers/can/can_socketcan.c
```diff
@@ -211,6 +204,39 @@ int csp_can_socketcan_open_and_add_interface(const char * device, const char * i
 	return CSP_ERR_NONE;
 }
 
+int csp_can_socketcan_set_promisc(const bool promisc, int *socket) {
+	static int *s_socket = NULL;
```
yashi, line 210: @arrooney, any reason you want to keep the given socket here?

src/drivers/can/can_socketcan.c
```diff
@@ -218,14 +218,22 @@ int csp_can_socketcan_set_promisc(const bool promisc, int *socket) {
...
-		if (setsockopt(*socket, SOL_CAN_RAW, CAN_RAW_FILTER, &filter, sizeof(filter)) < 0) {
-			csp_log_error("%s: setsockopt() failed, error: %s", __FUNCTION__, strerror(errno));
+		if (setsockopt(ctx->socket, SOL_CAN_RAW, CAN_RAW_FILTER, &filter, sizeof(filter)) < 0) {
+			csp_log_error("%s[%s]: setsockopt() failed, error: %s", __FUNCTION__, ctx->name, strerror(errno));
+			socketcan_free(ctx);
 			return CSP_ERR_INVAL;
 		}
-
 	} else {
 		struct can_filter filter = {.can_id = CFP_MAKE_DST(csp_get_address()),
```
yashi, line 238: Should we have a condition with `csp_conf.version` here, as we do in `!promisc`?

src/drivers/can/can_socketcan.c
johandc: I presume it is so it's possible to only include the socket in the init call and subsequently not need the reference.

I think the best solution here is to move the socket pointer to the interface driver structure and pass the CSP interface reference rather than the socket can pointer...

src/drivers/can/can_socketcan.c
johandc: I don't think so. There is already a check for the version and the mask generated

src/drivers/can/can_socketcan.c
yashi: currently, when you pass `socket` as `NULL`, you are not using `s_socket` in the following code.  But I get the idea.  Let me see what I can do.

yashi: hmm...  do you mean `csp_iface_t`, `csp_can_interface_data_t`, or `can_context_t`?  I'm assuming that we don't need to know the socket in abstracted `csp_iface_t` because a socket is socketcan specific. I'm also assuming that that's why you said _driver_.

PR #211, opened 2021-07-25
Build system

yashi:
Hi @johandc,

This is a PR to fix a few build system related failures.  Hope you like it.

For `csp_if_i2.c`, I just removed the file from build because meson wasn't building it. Do you have any opinion how to proceed with it?

For static rtable, I haven't remove it just yet.  If you are OK to remove it, I'll add the file to the commit.

Other changes should be trivial.  Let me know if you have any question.

### DIFF
--- a/src/interfaces/csp_if_udp.c
+++ b/src/interfaces/csp_if_udp.c
@@ -27,6 +27,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/arch/csp_thread.h>
 #include "../csp_id.h"
 
+#ifndef MSG_CONFIRM
+#define MSG_CONFIRM (0)
+#endif
+
 struct sockaddr_in peer_addr;
 static int lport = 9600;
 static int rport = 9600;
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ tags
 builddir
 .settings
 *~
+GPATH
+GRTAGS
+GTAGS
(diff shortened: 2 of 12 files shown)


src/interfaces/csp_if_udp.c
```diff
@@ -27,6 +27,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/arch/csp_thread.h>
 #include "../csp_id.h"
 
+#ifndef MSG_CONFIRM
+#define MSG_CONFIRM (0)
```
yashi, line 31: @johandc, what was the intention to use `MSG_CONFIRM` in the first place?  Linux man page states that

> **MSG_CONFIRM** (since Linux 2.3.15)
>     Tell the link layer that forward progress happened: you got a successful reply from the other side.  If the link layer doesn't get this it will regularly reprobe the neighbor (e.g., via a unicast ARP).  Valid only on SOCK_DGRAM and SOCK_RAW sockets and currently implemented only for IPv4 and IPv6.  See arp(7) for details.

Did you want to prevent reprobe? on FreeRTOS?

johandc (approved):

src/interfaces/csp_if_udp.c
johandc: If UDP is usually used with Linux. I am working on a new low level csp_if_eth which uses ethernet directly on bare bones such as Xilinx. But on Linux I guess the use of MSG_CONFIRM is nice to do.

johandc:
Thanks a lot for this.

Good suali stuff all around. Even caught the frame buf alignment issue.

Yes for now, if I2C needs to be deactivated. Gomspace has promised to port and test that during the fall.

PR #219, opened 2021-07-29
Add `IS_ENABLED()` macro

yashi:
Add Linux inspired `IS_ENABLE()` for feature check.  This PR has one source file, `csp_conn.c`, converted to using it for reference purpose.  If we like it, we'll convert other source files as well.
Because other projects integrating libcsp might have the same macro, we protect it with `#ifndef`.

### DIFF
--- a/include/csp/csp_util.h
+++ b/include/csp/csp_util.h
@@ -1,0 +1,33 @@
+/*
+ * Cubesat Space Protocol - A small network-layer protocol designed for Cubesats
+ * Copyright (C) 2021 Space Cubics, LLC.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _CSP_UTIL_H_
+#define _CSP_UTIL_H_
+
+#include <csp/csp_util_internal.h>
+
+/* IS_ENABLED(FOO) expands to 1 if defined to 1, or 0 otherwise even
+ * if not defined.  This allows us to use defines in C functions
+ * rather than #ifdef, which is much cleaner and compiler can verify
+ * the code even if an option is not defined. This macro is originated
+ * in Linux kernel and used in many open source projects. */
+#ifndef IS_ENABLED
+#define IS_ENABLED(config) CSP_IS_ENABLED1(config)
+#endif
+
+#endif /* _CSP_UTIL_H_ */
--- a/include/csp/csp_util_internal.h
+++ b/include/csp/csp_util_internal.h
@@ -1,0 +1,45 @@
+/*
+ * Cubesat Space Protocol - A small network-layer protocol designed for Cubesats
+ * Copyright (C) 2021 Space Cubics, LLC.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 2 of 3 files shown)


johandc:
I'm sorry but I don't like this.

In my opinion it's like crossing the river for water. You are adding two new header files, and 3 macro functions when you could simply write

If (CSP_USE_RDP)...

The IS_ENSBLED has zero benefits when checking on flags generated by the meson system or the WAF autoconfig. Because they will always be defined as 1, or #undef if they are 0

But even better, just keep
#if CSP_USE_RDP 

Because: I like the style better where the code removal is done more explicitly on a top level, line in the case of a #if. Putting code removal inside the tree just makes it look like it was a runtime check, and hides the fact that it's removed by the preprocessor.

The difference it makes might be subtle and more of a cosmetic thing. And it's all up to preferences i guess.

Thanks for your continued interest in the quality of the code. If you would like, we should have a short face to face, and I can give you some more permissions as a contributor.

yashi:
The whole point of having `IS_ENABLE()` is to make compiles check the whole source code without defines, instead of _removing_ from portion of source code from the view sight of compilers.  It seems that we are still in the state that all options / defines are orthogonal, but soon we will fail to check all combination of options.  In fact, we only run our CI with all options enabled.  No CI machine is checking without them.

In 2021, many, if not all, modern compilers have a good dead code removal optimization.  We can relay on that capability without developers explicitly stating which code to remove from executable.  We can even make build system to print executable size if we really care about the size.

If the introduction of the macro is only about a preference of personal favor, I'm happy to drop it.  But I see a real benefit of having the macro in projects.

For readability, having 15 `#if` in 415 lines of code (`csp_io.c`) isn't fun even with editor's support.  OK, this _is_ a personal preference. :stuck_out_tongue_winking_eye: 

For f2f, I'm happy to do that.  Drop me a line with times you are available.   Or are you on any IRC network or other realtime chat?

johandc:
> The whole point of having `IS_ENABLE()` is to make compiles check the whole source code without defines, instead of _removing_ from portion of source code from the view sight of compilers. It seems that we are still in the state that all options / defines are orthogonal, but soon we will fail to check all combination of options. In fact, we only run our CI with all options enabled. No CI machine is checking without them.

Okay, now I see the idea.

> 
> In 2021, many, if not all, modern compilers have a good dead code removal optimization. We can relay on that capability without developers explicitly stating which code to remove from executable. We can even make build system to print executable size if we really care about the size.

Agree I don't see any reason this shouldn't work. As I said it's more of a preference, because I think when having it inside the nested source makes it look as if the code is enabled.  Specifically the editor reads the preprocessor defines and greys out the code.

In space, dead code removal is mandatory, and sometimes it's easier if the code is greyed out when not in use.

I know the disadvantage is that you cannot make changes and test that part of the code unless the section is enabled. So stuff like syntax errors are first detected when that particular option have been enabled.

How would IS_ENABLED work for removing entire functions, and not just sections of code within functions?

Our meson always prints the binary size. But this is usually handled after linking, so for a library it makes it more tricky to evaluate the size.

> 
> If the introduction of the macro is only about a preference of personal favor, I'm happy to drop it. But I see a real benefit of having the macro in projects.


What could be done to simplify it?

Whats the reason to have two headers just for this? And whats the reason for the 1, 2 and 3 macro variations?  And why even have the macro we could just change the behaviour of the configuration system so it defines the CSP_USE_RDP to zero instead of not defining it.


> 
> For readability, having 15 `#if` in 415 lines of code (`csp_io.c`) isn't fun even with editor's support. OK, this _is_ a personal preference. 😜

Well I think it's exactly the opposite. Because the editor greys out which parts are enabled or not. This means you can much more easily skip over large portions of rarely used code.

> 
> For f2f, I'm happy to do that. Drop me a line with times you are available. Or are you on any IRC network or other realtime chat?

Drop me an email Johan at space-inventor.com we can arrange

yashi:
> How would `IS_ENABLED` work for removing entire functions, and not just sections of code within functions?

It simply doesn't work.  You have to either
- Use traditional `#ifdef`, or
- Use a build system feature to not build the entire file with the function.

For a library like libcsp, this happens with the functions with the same signature because C doesn't allow you to have multiple functions with the same name (unless with week attribute).

> Whats the reason to have two headers just for this?

I thought users don't care about implementation detail.  Thus, hiding under `csp_util_internal.h`.  This assumes that we will have more macros in `csp_util.h`.  The macro is just four lines, so we can put them in `csp_util.h`.  There is no technical reason at all.

> whats the reason for the 1, 2 and 3 macro variations?

As I commented on each line, every line has a meaning and a job to do.  If you can reduce them to 2 or 3 lines, that's great.  But I don't see a way to do.
With the four steps, you can create either `1` or `0` from a defined value or `0` from _undefined_ one.  Currently we have to make sure a define is defined with either `#ifndef` or `#if !defined`.  With this macro, you don't need them; undefined one will automatically turn into `0`.

> we could just change the behavior of the configuration system

Sure.  If we make sure all variables are defined, that'd be fine.  One situation I can think of is that sub-feature.  Say, we have `CSP_DEBUG` and `CSP_USE_ASSERT`, and `CSP_USE_ASSERT` depends on `CSP_DEBUG`.  We have to make sure to define  `CSP_USE_ASSERT` when we disable `CSP_DEBUG`.  This can easily happen If a parent option control `subdir()` in `meson.build`, you miss defining a variable, for example.  And this is a feature.

Hope I answer your questions.

johandc:
I think the best solution is to never use IFDEF at all. Everything IFDEF can do can also be done with just IF.

We should change the configuration system so that all compile time parameters are always defined. That solves any case, and doesn't need a macro. I cannot see any disadvantages to this approach.

Meson already defines all disabled options as 0. So it's just up to if WAF can do the same.

Issue #221, opened 2021-08-01
Simplify assert and stdout from library


yashi:
It seems to me that the current code have mixed debugging feature and logging feature.  i.e) In order to print error log with `csp_log_error()`, we must enable `CSP_DEBUG`.  The function to set log level is called `csp_debug_set_level()`, instead of `csp_log_set_level()`.  It might be a good idea to separate them.

Actually, it'd be just a bunch of renames to do `s/csp_debug_/csp_log_/ig`.  We only have `csp_assert()` for debugging.

johandc:
So if I understand you correctly, you want to have CSP_DEBUG just control the CSP_ASSERT, and rename debug_set_level to log_set_level?

Your point about the confusion is valid. But i think i have a better idea for the solution.
It's been a while since this systems was made, and if I should make it again I would do the following:

1. Use native assert.h I don't know why this should have ever been it's own function. Maybe for portability. But i have yet to find one that didn't have assert.h

2. Delete the entire log system. This is filled with nasty macros and just a right mess :)

3. Introduce a set of variables corresponding to each of the 7 so-called log levels we have today. I believe we already have the state of the debugging stored in a uint8_t array. However, this is quite limiting, and it's not easy to make a new debug "level" for lets say a new interface file under test or whatever.

4. Management of runtime variables. We have a few (RDP options, and these log levels) should be handled with a convenient helper function, but the variables themselves should be exported though a header file too. This is important because, it is not CSP's task to handle configuration management. And for users who already have a system for that (see https://github.com/spaceinventor/libparam) it is much easier when the parameters are just exported for memory mapping.

5. Everywhere logging is needed, instead of calling csp_log_error("some text") i would do the following:

```
if (csp_log_error) { printf("some text"); }

or

if (csp_log_error) {
    printf("some text");
}
```

It takes up slightly more room, but the code is readable and macro free. Much simpler to actually understand what happens. it also opens up to actual log levels. So the variable csp_log_error, could take multiple values, not just 0 or 1. For example when i'm writing drivers today, each individual driver / module or subsystem of the code has its own uint8_t log or debug parameter. This opens up for much much finer grained debugging. The other advantage is that when writing the driver i just declare my own log/debug parameter in my own local scope, meaning that its ultra simple and allows for infinite expansion, without a centralised log og debugging system.

yashi:
> Use native assert.h

You mean, `<assert.h>` from C standard library?  I'm all for it.

> Delete the entire log system.

[...]

PR #222, opened 2021-08-01
arch/zephyr: Add Zephyr RTOS support

yashi:
~This` is a RFC PR.  DO NOT MERGE.~

This commit adds a support for a RTOS called [Zephyr][1].  This PR is not ready for merge.  It compiles and works on a real board with Zephyr with the `lo` interface.  No hardware devices are included.

I'm creating this PR to ask everybody to check what I'm working on.
Some things I'd like to highlight are:

- Zephyr uses a CMake based build system with modules support.  This PR makes libcsp as a [Zephyr module][2].  In order to do so, I've created `zephyr/` directory at top.  All source code is located under `arch/zephyr/`, though.  I'd like to ask it's OK or not.  I can move the files under the top directory but that'll clutter the top directory for sure.
- libcsp currently is using dynamic memory allocations.  Zephyr, instead, prefers a static allocation over dynamic one.  I've made the Queues and the Semaphores with static allocations with hacks.
- `csp_thread_create()` is created with C preprocessor macro in order to use statically allocated memory locations.  Because it's a macro, some limitations applies.
- samples under `zephyr/samples/` are, obviously, taken from `examples/`, to show what we need to modify in order to run on another arch.  You can diff, i.e., `zephyr/samples/arch/main.c` against `examples/csp_arch.c`.  These files under `zephyr/samples/` will be removed before merge.
- files under `zephyr/tests/` are just for me.  They will be removed or refined before merge.

I'm sure I still miss some points I should make but without any further ado....  Please review :smiley: 

[1]: https://zephyrproject.org/
[2]: https://docs.zephyrproject.org/latest/guides/modules.html


include/csp/csp_types.h
```diff
@@ -37,7 +37,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 extern "C" {
 #endif
 
-#if (CSP_BIG_ENDIAN && CSP_LITTLE_ENDIAN)
+#if ((CSP_BIG_ENDIAN && CSP_LITTLE_ENDIAN) || !(CSP_BIG_ENDIAN || CSP_LITTLE_ENDIAN))
```
johandc: Mental note to self.

I believe we only use the endianess for two things:
1) in csp_types.h we had the csp_packet_t structure which used bitfields for the header fields. With the introduction of  csp 2.0, all header fields have become native size, and there is a function (in csp_id.c) that takes care of the packing.

2) in csp_endian.c within arch/ - I don't know how many exotic compilers csp is built on. but i would rather rely on the now almost mandatory endian.h or machine/endian.h i havent checked which is the correct one. But this works on picolibc, linux and macosx. I don't know if zephyr has one?

include/csp/arch/csp_thread.h
```diff
@@ -102,6 +102,46 @@ typedef csp_thread_return_t (* csp_thread_func_t)(void *);
...
+	csp_thread_return_t _csp_thread_entry_##task_name(void *arg1, void *arg2, void *arg3)
+
+static inline void set_handle(csp_thread_handle_t *handle, k_tid_t tid)
+{
+	if (handle) {
+		handle = tid;
+	}
+}
+
+#define csp_thread_create(entry_point, name, stack_size, arg, prio, handle)	\
```
johandc: Static memory allocation is something i have long wanted in CSP.

My suggestion is to make a new API for static stuff providing the memory for the stack and the OS specific task handle.
So something in the lines of. https://www.freertos.org/Static_Vs_Dynamic_Memory_Allocation.html

```
static char stack[1024 * sizeof(int)]; //! We define stacksize in bytes
static csp_task_t handle;
csp_create_task_static(function, arg_ptr, stack, sizeof(stack), handle, sizeof(handle)); 
```

The same logic applies to semaphores and queues

include/csp/arch/csp_thread.h
```diff
@@ -113,7 +153,9 @@ typedef csp_thread_return_t (* csp_thread_func_t)(void *);
    @param[out] handle reference to created thread.
    @return #CSP_ERR_NONE on success, otherwise an error code.
 */
+#if (!CSP_ZEPHYR)
```
johandc: I think it's okay for a platform to NOT define the dynamic memory API functions. As long as the static ones are there. CSP should be moved to static allocation internally. The only places where CSP is really using anything dynamic is already covered by memory pools, so the change should not be that difficult.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+if(CONFIG_LIBCSP)
+
+  zephyr_library_named(libcsp)
+
+  set(CSP_VERSION 1)
+
+  set(CSP_DEBUG ${CONFIG_LIBCSP_DEBUG})
```
johandc: All of this would disappear with the introduction of #221

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+    endif()
+    if (${CONFIG_LIBCSP_LOG_LEVEL} GREATER_EQUAL 3)
+      set(CSP_LOG_LEVEL_INFO 1)
+    endif()
+    if (${CONFIG_LIBCSP_LOG_LEVEL} GREATER_EQUAL 4)
+      set(CSP_LOG_LEVEL_DEBUG 1)
+    endif()
+  endif()
+
+  set(CSP_BIG_ENDIAN ${CONFIG_BIG_ENDIAN})
```
johandc: All of this would disappear with the introduction of #223

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+if(CONFIG_LIBCSP)
+
+  zephyr_library_named(libcsp)
```
johandc: Is there any way we could put this in either contrib/zephyr/CMakelists.txt or write the CMakelists.txt file generic, and move it to to the top folder to officially support CMake modules.
I can see theres a bit of zephyr specific stuff going on in here, and i'm not too familir with CMake. But isn't it bad to make a library use build macros defined outside the library itself?
could zephyr_library_named() and the other zephyr functions be reduces to native CMake syntax?

zephyr/Kconfig
```diff
@@ -0,0 +1,31 @@
+# libcsp Cubesat Space Protocol
+
+# Copyright (c) 2021 Space Cubics, LLC.
+# SPDX-License-Identifier: Apache-2.0
+
```
johandc: All these files are acceptable, but i would like to put them under contrib/zephyr to make room for other project contributions without cluttering the top directory too much.

johandc (changes requested):
This is generally accepted.

I have added a few suggestions for improvements, but there are more to the whole of CSP, instead of the zephyr stuff.

I haven't reviewed all the details, but am happy to host it inside a contrib/zephyr folder. I actually thought i would add that with example drivers for different chips too.

zephyr/CMakeLists.txt
yashi: I think we can put it anywhere.
But unfortunately we can't use generic CMakeLists.txt, or at least we can't directly use the generic CMakeLists.txt from Zephyr's build system.  That is because CMake doesn't allow to have multiple [`project()`](https://cmake.org/cmake/help/latest/command/project.html).
Instead, we might be able to have a top level CMakeLists.txt dedicated for libcsp with `project()` and a shared CMakeLists.txt which will be included from both the top level CMakeLists.txt and Zephyr.  I'll test a bit to see whether we can or not.

yashi: I've investigated this a bit today.  The results is:
- It was easy to replace all `zephyr_*` variant in the CMakeLists.txt, but
- Zephyr module system [assume that there is `zephyr/module.yaml`](https://docs.zephyrproject.org/latest/guides/modules.html#module-repositories)
- With CMake, we usually descend from top level CMakeLists.txt to sub-directories.  [It's possible for a sub-directory to have `project()` and work with top level CMakeLists.txt](https://stackoverflow.com/questions/7980784), but
  1. It's not recommenced
  2. We can't mix bottom up (Zephyr) and top down (libcsp proper) because the top level CMakeLists.txt always descend into sub-directories.
  3. It's much more complicated.

So, is it OK to have `zephyr/` at the top level and make all files zephyr specific?

zephyr/Kconfig
yashi: I agree.  But Zephyr module system assume that `zephyr/` exist...

johandc:
@yashi 

Hmm, okay. I have two comments.

1) First of, maybe we can fix the zephyr module (assuming zephyr modules are github repos) with a symlink. So maybe after the initial clone, you could have a small script, that sets up a symlink between contrib/zephyr and /zephyr, and then we can add /zephyr to the .gitignore file, so it won't show up as a change. This avoids cluttering the base dir, but also helps the zephyr build system to grab csp as a module.

2) I think we should consider adding a CMakeLists.txt file to the base of libcsp for those who use CMake. Personally i use Meson, GomSpace uses waf, and others use CMake. So as long as there are hands to maintain the build file, i think we can support all three systems. I even think Meson has ways to interoperate with CMake. Would you be interested in adding a CMakeLists.txt file for libcsp? 

Note: The very latest version of CSP on spaceinventor/libcsp/2.0 contains static allocation only. And requires a few more defines compile time.

yashi:
I'll search a bit more how not to clutter the top dir just for Zephyr module.

I use CMake just because Zephyr build system is based on it.  I prefer Meson as well for my personal uses.  But I can add basic support for it.  Even, I might be able to find a good way to integrate libcsp with Zephyr in the course of a cmakanization.  Created #229.

My concern is: Do you really want to maintain three efferent build system?  We should at least have CI for each.

Just for the reference for other visitors, Meson does have [CMake module](https://mesonbuild.com/CMake-module.html#cmake-subprojects), but it's not tier 1.  See [Meson's policy on mixing multiple build systems in one build directory](https://mesonbuild.com/Mixing-build-systems.html)

yashi:
> Note: The very latest version of CSP on spaceinventor/libcsp/2.0 contains static allocation only. And requires a few more defines compile time.

I'll convert Zephyr port to the static version.  Thanks :smile:

johandc:
> 
> My concern is: Do you really want to maintain three efferent build system? We should at least have CI for each.
> 

Not sure how much work it would be. I have a feeling that CMake and Meson are pretty similar. But no. I would rather support only a single, but if you want CMake, you can go and do that :)

johandc:
Do you still need a contrib/zephyr folder now that there's CMake support native?

yashi:
I still need `zephyr/` directroy.  I've checked code and also asked a maintainer about it and it is hard coded in the source code.  So, I'm planning to make a fork of libcsp, or a branch on this repo if you are ok with it, to have just small portion of zephyr glue code.
There will be at least three files:
- `module.yml`: this is the one I can't move
- CMakeLists.txt: converts Zephyr config to libcsp configuration
- `Kconfig`: Zephyr configuration system
- some Zephyr specific samples if applicable

What do you think?

yashi:
I just pushed another RFC.  I know I haven't fixed things you have pointed out the last.  But, this batch is to show you 
- how Zephyr port is now integrated with the CMake support, and
- static implementation of Zephyr thread, especially about stack initialization in `zephyr/samples/arch/main.c`.
```c
K_THREAD_STACK_DEFINE(stack, STACKSIZE);
```

I'll fix the things you pointed out in the next push.

include/csp/arch/csp_thread.h
```diff
@@ -1,7 +1,7 @@
 /*
 Cubesat Space Protocol - A small network-layer protocol designed for Cubesats
 Copyright (C) 2012 Gomspace ApS (http://www.gomspace.com)
-Copyright (C) 2012 AAUSAT3 Project (http://aausat3.space.aau.dk)
```
yashi: white space change.  will remove.

include/csp/arch/csp_thread.h
```diff
@@ -80,7 +80,7 @@ typedef HANDLE csp_thread_handle_t;
 typedef unsigned int csp_thread_return_t;
 typedef csp_thread_return_t (* csp_thread_func_t)(void *) __attribute__((stdcall));
 
-#define CSP_DEFINE_TASK(task_name) csp_thread_return_t __attribute__((stdcall)) task_name(void * param)
```
yashi: white space change.  will remove.

src/csp_service_handler.c
```diff
@@ -21,7 +21,13 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/csp.h>
 
 #include <stdio.h>
+#ifdef __GNUC__
+#ifndef alloca
+#define alloca __builtin_alloca
+#endif
+#else
 #include <alloca.h>
+#endif
```
yashi, line 30: MISRA doesn't like alloca and Zephyr toolchain doesn't directly support it.  AFAIK, TI toolchain, [Code Composer Studio](https://www.ti.com/tool/CCSTUDIO), doesn't support it neither.

zephyr/samples/arch/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.13.1)
```
yashi: will bump to the version 3.20.  (I think I used some feature from 3.20)

yashi (commented):
I just added some comments to myself.

src/csp_service_handler.c
yashi: I personally think that `alloca()` is same as large object on the stack, and love to use it.  But, ....

johandc:
What about the idea of having a pre build script setup a symlink from contrib/zephyr to /zephyr?

yashi:
**Short answer**:
No

**Long answer**:
The thing is that Zephyr's build is managed by a tool called [west](https://docs.zephyrproject.org/latest/guides/west/index.html) and it goes line this:
```shell
$ west init ~/myproject
$ cd ~/myproject
$ west update
$ west build --board qemu_cortex_m3 zephyr/samples/basic/blinky
$ west flash
```
where `west update` pulls all modules that the project depends on.  The first three steps are one time only.  That means that while in a development, You
1. modify your code
2. `west build` and test
3. Go back to 1

So, Yes, I can insert a new step (`ln -s ..../libcsp/contrib/zephyr ..../libcsp/zephyr`) after `west update` and before `west build`, Yes it's a persistent and one time only procedure for users, No it's not like a script-to-fix-them-all, because the path a module is placed in is arbitrary, depending on a project to a project. So,  I'd rather fork a repo for users so that they can follow what we have in [the Zephyr documents](https://docs.zephyrproject.org/latest/getting_started/index.html).

That said, if this is only me and my team, it's rather simple and take the symbolic link idea.  But we, Space Cubics, are planning to release the whole software as a SDK for cubesat.  That means that "users" will be someone who's new to the whole build process and I want it keep it simple and standard.

The unfortunate thing is that West doesn't have a hook which a module can use for whatever it needs, and, as said, the `/zephyr/module.yml` path is fixed right now. But we are discussing about enhancing module handing in Zephyr community.  Hope something comes up and fix this issue.  But, I'm sure it'll take a while.

So, I really appreciate your feedback and suggestion. But I'm thinking about placing `module.yml` in some other way if not allowed in the top directory.  (Don't take me wrong.  I understand that you don't want to clutter the top directory.  It's totally understandable.  I mean, many, if not all, modules have a dedicated repo forked from the upstreams. It's not just libcsp.)

ps. This is how I build and run libcsp `arch` sample
```shell
$ west build -t flash --board myboard modules/lib/libcsp/zephyr/sample/arch
```

johandc:
Okay, so here's the compromise:

1) You put the files into contrib/zephyr: This means their git location is where i want them. We also have a pull request for RIOT os, and i also want to make an extensive library of contributed network drivers, for CAN on different Chip's and so on. So the contrib directory is something i that i expect to grow.

2) We commit the symlink for /zephyr -> /crontrib/zephyr into the root of the main libcsp. This will of course clutter the root dir a bit, but let's not make it a bigger problem than it is at the moment. If we had 10 symlinks it would be annoying, but i think that a single is fine. I also would like to acknowledge your work and contribution. So i am happy to do that.

yashi:
Thank you for your kind words!

But, does it work under Windows / MacOS?

I know git handles symlinks OK on Linux but not sure others
https://github.com/git-for-windows/git/wiki/Symbolic-Links

I'm not currently planing to support Windows as a build env for our product, but both libcsp and Zephyr support them.

johandc:
Symlinks work on Mac.

Who cares about native windows build, even windows users are using Linux nowadays. There is no one left to maintain osx and windows support. So i think we should remove it with the 2.0 release.

I already removed windows build and Mac OS X from the CI.

include/csp/csp_types.h
yashi: I have created #250 to address this.

include/csp/arch/csp_thread.h
yashi: This one should be addressed by #247.

zephyr/CMakeLists.txt
yashi: This issue has been discussed in #221.  And I'd like to address it after we merged Zephyr port.

zephyr/CMakeLists.txt
yashi: Unfortunately csp_rdp.c is still using it. https://github.com/libcsp/libcsp/blob/53d740adbdf6c269a33ab68c2f548b03590dad8f/src/transport/csp_rdp.c#L78-L94

yashi: #250 has a commit to make sure that both defines are defined.

zephyr/CMakeLists.txt
yashi: It turns out Zephyr module can use a plain CMake build, and it's already merged #231.

For the top level `zephyr/`, we decided to go with a symlink https://github.com/libcsp/libcsp/pull/222#issuecomment-927958792

zephyr/Kconfig
yashi: Will be in `contrib/zephyr/`.

include/csp/arch/csp_thread.h
yashi: I've removed dynamic memory API from Zephyr.

yashi:
OK, I think I've address everything we've talked about.  As mentioned in the commit log, rtable doesn't work right now due to an empty `sscanf()`, some warnings are generated, or even `<endian.h` is not availabe yet.  But it's just matter of time.  Hope it's OK to merge as is.  So that we can work on other parts.

PR #227, opened 2021-09-04
src/bindings/python: Fix destructor for sfp_recv

IlievIliya92:
In order to deallocate the memory for the "dataout" pointer it is required to
use "csp_free". Since "dataout" is not a "csp_packet_t" pointer the
"pycsp_free_csp_buffer" destructor fails leading to memory leak.

Signed-off-by: Iliya Iliev <iliya@endurosat.com>

### DIFF
--- a/src/csp_sfp.c
+++ b/src/csp_sfp.c
@@ -71,6 +71,9 @@ int csp_sfp_send_own_memcpy(csp_conn_t * conn, const void * data, unsigned int t
 		return CSP_ERR_INVAL;
 	}
 
+	/* Add the csp packet to the total size */
+	totalsize += sizeof(csp_packet_t);
+
 	unsigned int count = 0;
 	while(count < totalsize) {
 
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -27,6 +27,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 #include <csp/drivers/usart.h>
 #include <csp/drivers/can_socketcan.h>
 #include <csp/csp_endian.h>
+#include <csp/arch/csp_malloc.h>
 
 #define SOCKET_CAPSULE      "csp_socket_t"
 #define CONNECTION_CAPSULE  "csp_conn_t"
@@ -72,6 +73,15 @@ static csp_socket_t * get_obj_as_socket(PyObject* obj, bool allow_null) {
 static PyObject* PyErr_Error(const char * message, int error) {
     PyErr_Format(Error, "%s, result/error: %d", message, error); // should set error as member
     return NULL;
+}
+
+static void pycsp_free_csp_sfp_buffer(PyObject *obj) {
+    void * sfp_packet = get_obj_as_packet(obj, true);
+    if (sfp_packet) {
+        csp_free(sfp_packet);
+    }
+
+    PyCapsule_SetPointer(obj, &CSP_POINTER_HAS_BEEN_FREED);
 }
 
 static void pycsp_free_csp_buffer(PyObject *obj) {
@@ -300,7 +310,7 @@ static PyObject* pycsp_sfp_recv(PyObject *self, PyObject *args) {
         return PyErr_Error("sfp_recv()", res);
     }
 
-    return PyCapsule_New(dataout, PACKET_CAPSULE, pycsp_free_csp_buffer);
+    return PyCapsule_New(dataout, PACKET_CAPSULE, pycsp_free_csp_sfp_buffer);
 }
 
 static PyObject* pycsp_transaction(PyObject *self, PyObject *args) {


yashi:
Hi there. 
It looks like a good catch.
This code was from @arrooney, I should ping.

src/csp_sfp.c
```diff
@@ -71,6 +71,9 @@ int csp_sfp_send_own_memcpy(csp_conn_t * conn, const void * data, unsigned int t
 		return CSP_ERR_INVAL;
 	}
 
+	/* Add the csp packet to the total size */
```
IlievIliya92, line 74: #### Sending a request message via the SF protocol *csp_sfp_send* function

The following is an example that illustrates sending a csp packet via the 
SF protocol using the *csp_sfp_send* function. 

```C
/************** Send a request message (Client side) ******************/
    csp_packet_t *csp_request = NULL;
    csp_request = csp_buffer_get(500);
    if (csp_request == NULL)
    {
        csp_log_warn("%s: Failed to get CSP buffer!", __FUNCTION__);
        /* Do some error handling */
    }

    /*
     * Fill in the csp_request message data accordingly to the application
     */


    unsigned int mtu = 256;
    uint32_t send_timeout = 1000;

    /* Send a request message */
    if (CSP_ERR_NONE != csp_sfp_send(conn,
                                     csp_request,
                                     csp_request->length,
                                     mtu, send_timeout))
    {
        csp_log_warn("%s: Failed to send CSP request!", __FUNCTION__);
        /* Do some error handling */
    }

```

**NOTE: Not adding the size of the 'csp_packet_t' structure to the total size
of the message to be send via the SF protocol  will result in losing
the last 48 bytes of the data stored in the csp_request->data pointer.**

#### REceiving a request message via the SF protocol *csp_sfp_recv* function

The following is an example that illustrates receiving csp packet via the 
SF protocol using the *csp_recv_send* function (assuming that there is an already
accepted connection *(csp_conn_t *conn)* via *csp_accept*).


```C
    /************** Receive request message (Server side) ******************/
    csp_packet_t *csp_request = NULL;
    void *dataout = NULL;
    int return_datasize = 0;
    uint32_t recv_timeout = 1000;

    csp_request = csp_buffer_get(500);
    if (csp_request == NULL)
    {
        csp_log_warn("%s: Failed to get CSP buffer!", __FUNCTION__);
        /* Do some error handling */
    }
    else
    {
        if (CSP_ERR_NONE != csp_sfp_recv(conn, &dataout, &return_datasize, 
                                        recv_timeout))
        {
            csp_log_warn("%s: Failed to recv CSP request!", __FUNCTION__);
            /* Do some error handling */
        }
        else
        {
            /* Get the request message */
            csp_request->length = return_datasize;
            memcpy(csp_request->data, (uint8_t *)dataout, csp_request->length);             
        }
    }

```

src/csp_sfp.c
```diff
@@ -71,6 +71,9 @@ int csp_sfp_send_own_memcpy(csp_conn_t * conn, const void * data, unsigned int t
...
+	totalsize += sizeof(csp_packet_t);
+
```
IlievIliya92, line 76: #### Sending a request message via the SF protocol *csp_sfp_send* function

The following is an example that illustrates sending a csp packet via the 
SF protocol using the *csp_sfp_send* function. 

```C
/************** Send a request message (Client side) ******************/
    csp_packet_t *csp_request = NULL;
    csp_request = csp_buffer_get(500);
    if (csp_request == NULL)
    {
        csp_log_warn("%s: Failed to get CSP buffer!", __FUNCTION__);
        /* Do some error handling */
    }

    /*
     * Fill in the csp_request message data accordingly to the application
     */


    unsigned int mtu = 256;
    uint32_t send_timeout = 1000;

    /* Send a request message */
    if (CSP_ERR_NONE != csp_sfp_send(conn,
                                     csp_request,
                                     csp_request->length,
                                     mtu, send_timeout))
    {
        csp_log_warn("%s: Failed to send CSP request!", __FUNCTION__);
        /* Do some error handling */
    }

```

**NOTE: Not adding the size of the 'csp_packet_t' structure to the total size
of the message to be send via the SF protocol  will result in losing
the last 48 bytes of the data stored in the csp_request->data pointer.**

#### REceiving a request message via the SF protocol *csp_sfp_recv* function

The following is an example that illustrates receiving csp packet via the 
SF protocol using the *csp_recv_send* function (assuming that there is an already
accepted connection *(csp_conn_t *conn)* via *csp_accept*).


```C
    /************** Receive request message (Server side) ******************/
    csp_packet_t *csp_request = NULL;
    void *dataout = NULL;
    int return_datasize = 0;
    uint32_t recv_timeout = 1000;

    csp_request = csp_buffer_get(500);
    if (csp_request == NULL)
    {
        csp_log_warn("%s: Failed to get CSP buffer!", __FUNCTION__);
        /* Do some error handling */
    }
    else
    {
        if (CSP_ERR_NONE != csp_sfp_recv(conn, &dataout, &return_datasize, 
                                        recv_timeout))
        {
            csp_log_warn("%s: Failed to recv CSP request!", __FUNCTION__);
            /* Do some error handling */
        }
        else
        {
            /* Get the request message */
            csp_request->length = return_datasize;
            memcpy(csp_request->data, (uint8_t *)dataout, csp_request->length);             
        }
    }

```

johandc: I think this concept has been slightly misunderstood. SFP was meant to be used to transfer any type of data, given by the pointer and the length. It is correct that you might use it to encapsulate a larger CSP packet. But that's not an effective way to do things. You should ensure that the maximum MTU is not exceeded by any of the end-nodes. If that is not possible, then try to increase the MTU of the unit with a smaller MTU in the middle (for example a radio).

Added the length of the csp_packet_t structure is a mistake, because it would break the API for anyone else who is not using it for sending that partiruclar data type. If you want to send csp packets this way, you should add the additional length to the csp_sfp_send call. (csp_request->length + sizeof(header)) - BUT. this is also a mistake. Since CSP 1 and 2 uses different header lengths. Furthermore the csp_packet_t structure is different sizes on different architectures. (it's a non packed struct, which contains padding an other stuff)..

The correct way to prepare a CSP packet for transmission is to use csp_id_prepend and then using the frame * and frame_len.

```

csp_id_prepend(packet);

whatever_interface_send(packet->frame_begin, packet->frame_length)
```

IlievIliya92: Thank you for the clarification on the SFP usage!

PR #228, opened 2021-09-19
2.0 Static memory allocation

johandc:
Static memory allocation


johandc:
Excuse the rather clumsy commit log, and lack of rebasing. Since i'm pretty much the only developer on 2.0 at the moment, i'm taking the freedom to just commit linearly as i go along.

Issue #229, opened 2021-09-19
Support CMake


yashi:
Add CMake support.

This came up in the discussion https://github.com/libcsp/libcsp/pull/222#issuecomment-922434719

> I think we should consider adding a CMakeLists.txt file to the base of libcsp for those who use CMake. Personally i use Meson, GomSpace uses waf, and others use CMake. So as long as there are hands to maintain the build file, i think we can support all three systems. I even think Meson has ways to interoperate with CMake.

PR #231, opened 2021-09-23
[RFC] Add CMake support

yashi:
This commit add CMake suport for libcsp. Addressing #229.

This commit is only for posix target.  No other target has been tested yet.  I haven't tested with Zephyr at all.  So this is really a RFC PR.  Not only for CMake but I'd like to integrate freertos into the build, especially on our CI.

Please give me any feedback.

- [X] Test to see Zephyr module works with it
- [X] Add proper commit log
- [ ] Document about CMake
- [X] Integrate with `buildall.py`
- [X] Add CI support

Thanks,


johandc:
Due to the similarities between meson and cmake I don't think it will be difficult to support both fully.

We need gomspace to make a decision on supporting WAF or changing to either meson or cmake.

yashi:
yeah, that's what I thought, too.  I tried to follow what we have in `meson.build`.  And that turns out to be a quite OK.  If we decided to have them both, we'd better keep them synced.

If Zephyr can work with this CMake, it will be a win.  I'll let you know how it turns out in a few days..

yashi:
I've done working on CMake integration and this works with Zephyr module as well!! :tada:  So please review and if you like please merge it.  I'll maintain CMake support as well as Zephyr port.

johandc (approved):

PR #239, opened 2021-09-27
Fix ci build

yashi:
Let's see how it goes.

### DIFF
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -15,7 +15,7 @@ def build_with_meson():
                     'zmqproxy']
     builddir = 'build'
 
-    meson_setup = ['meson', 'configure', builddir]
+    meson_setup = ['meson', 'setup', builddir]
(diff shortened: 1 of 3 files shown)


johandc (approved):

yashi:
Oops, thank you for merging it but the PR was still WIP and a draft. 😁

I couldn't finish it by the last train but t seems we still have a execution problem.

I'll finish up tomorrow.

johandc:
I was hoping the build would work after the merge. Because you had added examples to the target names.
At least CSP now compiles it's just a matter of fixing the examples. I don't think you need the examples on the target names for the ninja build.

johandc:
I'm just trying a few changes directly on the development branch from the GitHub online editor from my phone here. Excuse the mess if it doesn't work.

johandc:
Finally fixed it. Thanks for your work

PR #245, opened 2021-09-27
Fix CMake build

yashi:
@johandc, I missed one change for CMake build.  Please merge.


yashi:
actually, there are a few things I missed.

PR #247, opened 2021-09-28
[RFC] static: Add csp_thread_create_static() for POSIX

yashi:
This is a RFC.

I came up with this implementation of csp_thread_create_static() for POSIX in the course of Zephyr port.

**Empty struct for unused type**:

I came up with `empty{}` for unused type.  Is it acceptable?

**Use of abort()**:

As we discussed we don't want to fail from static version of functions, thus I just put it abort() in it.  Do you think it's acceptable?

**Order of Function arguments**:

For argument order, I think this version is cleaner but we should follow the non static version for compatibility.

### DIFF
--- a/include/csp/arch/csp_thread.h
+++ b/include/csp/arch/csp_thread.h
@@ -44,6 +44,8 @@ extern "C" {
    Platform specific thread handle.
 */
 typedef pthread_t csp_thread_handle_t;
+typedef pthread_t csp_thread_t;
+typedef struct empty{} csp_stack_t;
 
 /**
    Platform specific thread return type.
--- a/src/arch/posix/csp_thread.c
+++ b/src/arch/posix/csp_thread.c
@@ -22,6 +22,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 #include <limits.h>
 #include <unistd.h>
+#include <stdlib.h>
 #include <time.h>
 #include <errno.h>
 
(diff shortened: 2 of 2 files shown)


johandc:
I am willing to go even further on this one.

Let's remove csp_thread.

Here's the rationale. CSP was never designed to be an OS abstraction layer. In fact all the includes in the /arch dir was just recently (as in a couple of years ago) made public. The internal arch directory only contained what's necessary for CSP to build and run.

I just searched the code, and the only place csp thread create is used is for the router/bridge task. In order to support barebone (no-os) the working part of the router task has long ago been split into a router_work() function. With only a tiny dummy task calling that work function in a loop.

So here's the deal. We do not provide csp_thread.h, but instead require the user to call csp_init, and then csp_router_work() (non blocking), or csp_router_loop() (blocking mode, takes over the calling task)

It could be argued that without csp_thread is is now more difficult to write portable code that uses the CSP stack. For example we have a parameter service (libparam) that listens on a certain port number. However for such tasks again it should be easy to split into functions like the above.

This also lends to easier handling of coroutines when all work functions can be serialised.

It would remove some of the hassle of maintaining portability across different OS

yashi:
I'm all for it!

And your post answers my question at https://github.com/libcsp/libcsp/issues/221#issuecomment-927641486 as well.

Now, from a practical point of view, how do you want to run a thread in examples? For real applications, I think it's fine because they are mostly targeting a single platform. Even an application targeting multiple platform, we can just have conditional compilation with multiple files.  How they manage target specific code is up to them.

But for example apps in libcsp, what do you want to do?

One option is to have a weak hook function, say, csp_arch_init(), in libcsp and call it from csp_init().  The default csp_arch_init() behavior can be either a) do nothing and return, or b) warn about it and exit.  An application can define the same function for each target.

Interface drivers, if_zmqhub and if_udp, also define tasks. (but nobody calling it?)

yashi:
@johandc, do you want to remove `csp_thread*` before the Zephyr merge or after?

johandc:
I think after. There is no rush. But no-one forces you to use those functions even today.

johandc:
Merging this, it will only be temporary anyways.

Issue #230, opened 2021-09-20
Move Travis CI to other CI


yashi:
Travis CI [does not offer free tier](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) any more.  And said open source offer is [already blocked](https://twitter.com/james_hilliard/status/1336081776691843072).

So, we must find some CI services with
- No cost
- Linux, Windows, Mac must be supported

This is related to #212.

johandc:
I got a mail from Travis they will be giving us some OSS credits.

yashi:
That's great.  Hope it will last until we migrate to another one.  I'll keep posting if I get something new.

PR #248, opened 2021-09-28
Add CMake to buildall.py (and fix waf)

yashi:
This PR is to add CMake support to buildall.py.

In order to do so, I have to fix waf build which is left behind.  Is it intentional?  We don't test waf on CI any more.  If we don't support waf in the `develop` branch, we can just drop it entirely. If the credit on Travis CI is the issue, I actually planning to add Github Actions to libcsp.  This allows us to use much more build resources.

For python binding, I just didn't have enough motivation to fix `csp_init()` change in the binding because I don't use libcsp with python at all.  And other build systems don't build with buildall.py anyway. I'll let you decide what to do with it.

### DIFF
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -42,7 +42,6 @@ def build_with_waf(options):
     waf = ['./waf']
     if target_os in ['posix']:
         options += [
-            '--enable-python3-bindings',
             '--enable-can-socketcan',
             '--with-driver-usart=linux',
             '--enable-if-zmqhub',
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -11,8 +11,8 @@ DEFAULT_BUILD_SYSTEM = 'waf'
 
 def build_with_meson():
     targets = ['examples/csp_server_client',
-                    'examples/csp_arch',
-                    'examples/zmqproxy']
+               'examples/csp_arch',
+               'examples/zmqproxy']
     builddir = 'build'
 
     meson_setup = ['meson', 'setup', builddir]
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -21,6 +21,18 @@ def build_with_meson():
     subprocess.check_call(meson_compile + targets)
 
 
+def build_with_cmake():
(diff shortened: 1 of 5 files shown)


yashi:
This is partially related to #246 and #230.

How do you want `buildall.py` to be?

1. simple and default build for all build systems
2. extensible handing of all combination of user customizable parameters.

I mean, in #246, we failed to detect !CSP_USE_RDP case.  If we want to check this kind of build case, we can either
1. tell CI to handle parameter matrix
2. tell `buildall.py` to handle all combinations in one go

For the case 1, we can just execute the build system directly.  No `buildall.py` involvement is required.
For the case 2, we must tell `buildall.py` to either a) handle parameter matrix internally, or b) let it take option parameters and convert to specific build system.
- a)
  - PROs: CI doesn't have to care about parameter matrix.  Just run `buildall.py` and done with it.
  - CONs: CI can't tell which combination failed.
- b)
  - PROs: CI can tell which combination failed
  - CONs: `buildall.py` gets cluttered with parameter conversion code

I assume that `buildall.py` was created with an intention to make build easier.  But it doesn't really work well with CI and it puts us more work, IMHO.

So I propose "1"
- keep `buildall.py` simple with the default "build all" pattern, and
- use each build system on CI if we want to test build options.

WDYT?

examples/buildall.py
```diff
@@ -21,6 +21,18 @@ def build_with_meson():
...
+def build_with_cmake():
+    targets = ['examples/csp_server_client',
+               'examples/csp_arch',
+               'examples/zmqproxy']
+    builddir = 'build'
+
+    cmake_setup = ['cmake', '-GNinja', '-B' + builddir]
+    cmake_compile = ['ninja', '-C', builddir]
+    subprocess.check_call(meson_setup)
+    subprocess.check_call(meson_compile + targets)
```
yashi: Wrong variables.  Must be fixed.

yashi (commented):
This is a bug. Will fix.

PR #250, opened 2021-09-29
Remove the endian check from csp_types.h

yashi:
This PR is for https://github.com/libcsp/libcsp/pull/222#discussion_r680510244 and also the result of discussion from #219.

Hope you like it.


johandc:
After the removal of <csp_endian.h> we do not need these defines.
The only place left that it is used is in csp_rdp.c line 65. For packing of bitfields in the RDP header. 
This should be rewritten in the same form as was done with the csp header in csp_id.c

PR #269, opened 2021-10-19
Windows USART issue fix

domantascibas:
- fixes in `usart_windows.c`:
    - fixes the `fd` COM port handle in the `csp_usart_open()` function, which was lost
    - reimplements the `csp_usart_write()` and the `usart_rx_thread()` functions to allow overlapped data transmission. in other words - allows simultaneous data RX and TX
- fixes in `csp_if_kiss.c`:
    - reimplements the `csp_kiss_tx()` function to increase the data TX speeds. slow data transmission speeds were caused by sending the data a single byte at a time, which means that the `csp_usart_write()` function is sending one byte at a time and adds a lot of overhead. now the `csp_kiss_tx()` function is sending data in larger (up to 2048B) chunks. a temporary data buffer is used to store data and add the special ESC characters. it takes a bit more resources to prepare the data packet to be sent, but reduces the overall transmission time

### DIFF
--- a/src/drivers/usart/usart_windows.c
+++ b/src/drivers/usart/usart_windows.c
@@ -100,7 +100,6 @@ static int setPortTimeouts(csp_usart_fd_t fd) {
 
 static unsigned WINAPI usart_rx_thread(void* params) {
     usart_context_t * ctx = params;
-    DWORD eventStatus;
     uint8_t cbuf[100];
     DWORD bytesRead;
     DWORD dwErrors;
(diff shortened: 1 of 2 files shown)


johandc:
Hello. Thanks for the contribution.

I have two comments. Number one is the requirement of 2048 byte stack is too high. Not many embedded system are running with that large stacks.

On a Windows PC I don't understand why there would be an issue with the uart interface writing one byte at a time, since the output driver should be buffered. Just like it is on Linux. Furthermore, a PC should have no trouble keeping up with the slow speed of s serial port.

Adding the buffered output to CSP kiss interface is not really the place. It might be done later in the uart driver layer or the OS if not already.

Issue #243, opened 2021-09-27
atomics


johandc:
Consider using atomics for simple locks (such as the source port lock)

- [x] sport lock
- [x] connection table lock

johandc:
If we can add an interface that uses atomics when availabe, and reverts to using binary semaphore's when not. It would be nice.

yashi:
Does it mean we'll be using C11 instead of C99?

johandc:
I completely removed the source port lock by pre-assigning source ports to each connection. I think it's pretty clever, because it actually took a O(n * m) search down to O(1).

For this to work, we assume that number of connections is lower than the number of port numbers available for outgoing (typically 48). There is a compile time check for this.

johandc:
I added C11 atomics for the connection table lock using the:

  `atomic_compare_exchange`

Function. The compiler typically knows what to do. But in the case of Cortex-M0 it's not supported, and it makes an external function call to a user implementation. I therefore added an example in 'contrib/atomics/atomics_freertos_gcc.c' to show how the user may implement such thing on Cortex-M0 and FreeRTOS. It's typically just deactivating interrrupts while performing the compare and exchange.

On more advanced ARM systems such as cortex M3, M4, M7, A9... and up, as well as X86, and ARM64, the compiler has assembler intrinsics that performs this using dedicated instructions.

The removal of the sport and conn locks has reduced a M0 image size by 300 bytes .text and 100 bytes .bss.

PR #274, opened 2021-10-28
build: Enable C11

yashi:
As #243, we'll be using atomics which is a feature in C11.  Enable C11 in the build systems.


johandc:
Yes. We will use c11

PR #284, opened 2021-11-02
Remove csp_thread support

yashi:
This is a RFC PR to show how we can gracefully remove csp_thread support from libcsp.  It's not complete yet, but hope this shows some idea.  ~~After this PR, only user is the UDP interface driver.  I'd be nice if you can help me removing csp_thread from it.~~

After removing csp_thread, we can either remove `csp_<arch>_thread_create()` or leave simplified version in-tree, especially for POSIX.  Because we need it anyway for Python bindings as well as arch dependent drivers.

WDYT?

### DIFF
--- a/contrib/zephyr/samples/arch/main.c
+++ b/contrib/zephyr/samples/arch/main.c
@@ -47,7 +47,7 @@ int main(int argc, char * argv[]) {
     csp_thread_handle_t tid;
     csp_thread_t new_thread;
 
-    tid = csp_thread_create_static(&new_thread, "thread",
+    tid = csp_zephyr_thread_create_static(&new_thread, "thread",
 				   stack, K_THREAD_STACK_SIZEOF(stack),
 				   thread_func, NULL, 0);
     //csp_assert(res == CSP_ERR_NONE);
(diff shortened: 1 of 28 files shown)


yashi:
BTW, this PR includes #283.

yashi:
Added csp_if_udp refactoring. csp_thread_create() is not touched yet and csp_if_udp_rx_loop() and csp_if_udp_init() are still in the file.  But, it's much easier to see how it goes.  I wanted to show this in order to ask you for feedbacks.

johandc:
I like what i see :)

1) The UDP interface is POSIX only for now. So we can do the same as with zmqhub.
2) You added a arch specific csp_bridge_start() into /arch. Should we do the same with csp_router_start() in order to be consistent? - Also perhaps i think these should be put into examples, and not be part of the API. To "force" users to copy them into their project, and actually use the real _work() API. What do you think? For example, with FreeRTOS the user has to decide on a stack size. But on Posix that's not necessary.
3) The timeout on work() functions i think we should keep, and then pass it along to the qfifo instead of the define. So it could be used in non-blocking mode.

yashi:
That's good to hear :smile:. I'll keep working on it.

1. OK
2. We can certainly move `_start()` to examples, yes.  And I agree that for the real apps, we call `_work()` from platform specific way.
3. you want to modify `csp_qfifo_read()` by adding timeout as a new param, right?  If so, I'll do that for the last commit of this series, or after the merge (because that's nothing to do with csp_thread removal.  How does that sound?

johandc:
3. Yes, exactly.

yashi:
There is no user of `csp_bridge_start()` in examples.  Do you have any plan to implement a new sample using it?

johandc:
Just did a checkout and build of our software, and it works for me :+1:

Issue #259, opened 2021-10-02
Remove csp_thread


yashi:
This came up in https://github.com/libcsp/libcsp/pull/247#issuecomment-929391729
It's easier to remove but we need to decide how we run threads uniformly ​on different platforms.

yashi:
> So here's the deal. We do not provide csp_thread.h, but instead require the user to call csp_init, and then csp_router_work() (non blocking), or csp_router_loop() (blocking mode, takes over the calling task)

We have following `CSP_DEFINE_TASK()`
- `examples/csp_arch.c`
- `examples/csp_server_client.c`
- `src/csp_route.c` :: `csp_task_router`
- `src/csp_bridge.c` :: `csp_bridge`
- `src/interfaces/csp_if_zmqhub.c` :: `csp_zmqhub_task`
- `src/interfaces/csp_if_udp.c` :: `csp_if_udp_rx_task`

If we provide `csp_router_work()` and `csp_router_loop()`, we should also provide

- `csp_bridge_work()` and `csp_bridge_loop()`
- `csp_if_udp_rx_work()` and `and csp_if_udp_rx_loop()`
- `csp_zmqhub_work()` and `csp_zmqhub_loop()`

For the sample code under `examples/`, do you think it's OK to move each code under its own directory? So that we can have arch dependent files for each arch?

For example,

```
├── csp_arch
│   ├── csp_arch.c
│   ├── freertos.c
│   ├── linux.c
│   ├── win.c
│   └── zephyr.c
├── csp_server_client
│   ├── csp_server_client.c
│   ├── freertos.c
│   ├── linux.c
│   ├── win.c
│   └── zephyr.c
└── zmqproxy
    ├── zmqproxy.c
    ├── freertos.c
    ├── linux.c
    ├── win.c
    └── zephyr.c
```

In this case, we might want to rename the main file to `main.c`. Also would be nice to have `arch_` prefix for arch dependent code, ie `arch_linux.c`.  (`arch_posix.c`?)

johandc:
Yes, Either that or:

```
 csp_arch.c
 csp_arch_linux.c
 csp_arch_freertos.c
 zmqproxy_linux.c
```

But i'm a little in doubt about how the FreeRTOS and Zephyr examples would run. Normally they should be included in another project in order for the OS to run, and be linked for the target. I think we may consider using an additional test repository to house some example projects which could be linked to some common target boards, like the SamExplained series.

yashi:
[...]

PR #285, opened 2021-11-04
Remove csp_thread support part 2

yashi:
This is part 2 of csp_thread removal series(#259).  The part 1 is #284. This PR removes most of `csp_thread.h` including all typedefs and POSIX and Zephyr version of `csp_*_thread_create()`.  FreeRTOS and Window  still have `*_thread_create()` in csp_thread.h.

This PR is to show how I'm removing the functions and types from csp_thread.h and get feedback.

csp_thread_create is gone but the semantics should be the same.  All pthreads are created detached. Returns to the caller with error value if possible.  No extra check is done to keep the same behavior of the original csp_thread_create().

One thing I've changed is the stack size.  Because we don't care about the stack size on POSIX, we are now using the default size by ignoring the given one.  We'd later improve emulating a small stack size on POSIX if we are willing to I suppose.

I haven't moved arch specific functions to examples, just yet. Will do once all csp_*_thread_create() are replaced by the native functions

Please review and test it with your applications.  And give me any feedback.


yashi:
Rebased on top of #286

PR #286, opened 2021-11-04
Make windows port work

yashi:
Here is my attempt to make libcsp buildable for Windows.  I don't have Windows, so this is only tested on Linux with MinGW cross compile and wine to run it.

I know windows is not tier 1 and I don't personally use it at all.  But because it was quite easy to revive just for examples, I just did it.  Oh, please don't ask me why I didn't incorporate #269.  Again I don't care about Windows and I don't know anything about it.

We don't have to merge this. I just did this to remove csp_thread from windows port.

If anyone seeing this with Windows knowledge and interests, please review.

### DIFF
--- a/examples/csp_server_client_windows.c
+++ b/examples/csp_server_client_windows.c
@@ -1,0 +1,48 @@
+#include <csp/csp.h>
+#include <windows.h>
+#include <process.h>
+
+void server(void);
+void client(void);
+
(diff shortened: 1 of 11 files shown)


yashi:
I've dropped "windows: wscript: Make sys/socket.h optional" in favor of #287.  Actually this PR is rebased on top of it.

PR #287, opened 2021-11-05
waf: avoid exception if header check fails.

bfol:
Do not raise exception of the header files to be checked for does not exist

### DIFF
--- a/wscript
+++ b/wscript
@@ -116,7 +116,7 @@ def configure(ctx):
                                         'src/rtable/csp_rtable_{0}.c'.format(ctx.options.with_rtable)])
 
     # Add if UDP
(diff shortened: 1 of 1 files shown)


yashi:
hello!

I also have similar fix in #286.  And yes I think both should be optional.  Would you mind if I ask which platform are you building on?  If you are on Windows, would you mind to take a look at #286?  Thanks.

bfol:
I build on linux - for an avr32 based embeded platform - so thankfully no windows involved.

yashi:
Thanks!

BTW, if you can provide us how you build libcsp for avr32, we can check it in GitHub Actions. If you are interested in, please open an issue with steps to setup toolchains and build.

PR #288, opened 2021-11-05
Cmp route set compat

bfol:
Restore on the wire compatibility of cmp_route_set with nodes built with version 1.6 or earlier


johandc:
Sure. 👌

PR #289, opened 2021-11-05
Remove csp_thread support part 3

yashi:
This is part 3 of csp_thread removal series(#259). The part 1 is #284, the part 2 is #285. This PR removes Windows version of `csp_*_thread_create()`. FreeRTOS is the only remaining one in libcsp.

Without `csp_thread_create()` wrapper, it's very simple one-liner to create a thread under Windows:

    uintptr_t ret = _beginthreadex(NULL, 0, usart_rx_thread, NULL, 0, NULL);

I use `uintptr_t` for the return variable because that's what `_beginthreadex()` returns.  We don't have anything to pass to the function except the thread entry function.

The wrapper was using `0` for the last parameter but it's actually a pointer to an unsigned int.  So I've changed it to NULL.  That's what the doc says:

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex

    thrdaddr
    Points to a 32-bit variable that receives the thread identifier. If it's NULL, it's not used.


johandc:
I like everything I see.

Overall very high quality work, and a very well disciplined git commit history that is easy to follow. 

The only thing i think we need to change is to remove csp_bridge_start() from the API and thus from the arch layer too.

PR #290, opened 2021-11-06
src/bindings/python: Fix python binding for sfp_recv

IlievIliya92:
src/bindings/python: Fix python binding for sfp_recv
       
 The encapsulated object is returned as raw bytes using
 the data from 'dataout' with length 'return_datasize'

### DIFF
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -204,7 +204,6 @@ static PyObject * pycsp_send(PyObject * self, PyObject * args) {
 	Py_RETURN_NONE;
 }
 
-#if 0
 static PyObject* pycsp_sfp_send(PyObject *self, PyObject *args) {
     PyObject* conn_capsule;
     Py_buffer data;
@@ -255,9 +254,8 @@ static PyObject* pycsp_sfp_recv(PyObject *self, PyObject *args) {
         return PyErr_Error("sfp_recv()", res);
     }
 
-    return PyCapsule_New(dataout, PACKET_CAPSULE, pycsp_free_csp_buffer);
+    return Py_BuildValue("y#", dataout, return_datasize);
 }
-#endif
 
 static PyObject * pycsp_transaction(PyObject * self, PyObject * args) {
 	uint8_t prio;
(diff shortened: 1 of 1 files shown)


src/bindings/python/pycsp.c
```diff
@@ -255,9 +254,8 @@ static PyObject* pycsp_sfp_recv(PyObject *self, PyObject *args) {
         return PyErr_Error("sfp_recv()", res);
     }
 
-    return PyCapsule_New(dataout, PACKET_CAPSULE, pycsp_free_csp_buffer);
+    return Py_BuildValue("y#", dataout, return_datasize);
```
yashi, line 257: You are leaking `dataout` here.  `csp_sfp_recv()`, or `csp_sfp_recv_fp()` to be exact,  `malloc()`s a memory region and copy received packets in it, then returns as `dataout`.  `Py_BuildValue()` will `memcpy()` from the given `dataout` to the newly created bytes object.

> In order to deallocate the memory for the "dataout" pointer it is required to use "csp_free".

As said above, `dataout` is `malloc()`ed.  Thus, this is false.  You can just `free()` it.  But, in order to avoid the extra `memcpy()`, why not keep using `PyCapsule_New()` with `free()`?  Doesn't it work?  I don't have sfp python code so I can't test.

Also, it used to return a Capsule object but now return a bytes object.  But this should be OK, I assume.

yashi (changes requested):

src/bindings/python/pycsp.c
yashi: BTW, we are removing dynamic allocation from libcsp.  it'd be nice if we can remove malloc() from sfp.

johandc:
Hello.

I commented out SFP because I wanted a malloc free (no pun intended) CSP implementation. Do you really use SFP? It's quite inefficient and error prone due to the lack of retransmission.

CSP now supports packets size of 2kB. So I suggest increasing MTU size og switching up to RDP for larger data transfers.

IlievIliya92:
Hi, 

Sounds, great! Will use a larger MTU size instead. Thank you for the information!

PR #292, opened 2021-11-12
Fix null pointer dereference

avinashraja98:
NULL was being passed in as arg for the usart_rx_thread which when dereferenced caused a segmentation fault. This error can be seen when testing a kiss interface using socat.

### DIFF
--- a/src/drivers/usart/usart_linux.c
+++ b/src/drivers/usart/usart_linux.c
@@ -281,7 +281,7 @@ int csp_usart_open(const csp_usart_conf_t * conf, csp_usart_callback_t rx_callba
 			return CSP_ERR_NOMEM;
 		}
 		pthread_attr_setdetachstate(&attributes, PTHREAD_CREATE_DETACHED);
-		ret = pthread_create(&ctx->rx_thread, &attributes, usart_rx_thread, NULL);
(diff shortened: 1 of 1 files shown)


yashi (approved):
Thanks.

This bug was introduced in d306789b8417539.

yashi:
I've checked all other `pthread_create()` call sites:
- examples/zmqproxy.c
- examples/csp_server_client_posix.c
- src/arch/posix/csp_bridge.c
- src/drivers/can/can_socketcan.c
- src/bindings/python/pycsp.c
- src/interfaces/csp_if_zmqhub.c
- src/interfaces/csp_if_udp.c

Among these callers, the followings are passing `NULL`

In src/bindings/python/pycsp.c:
----
```c
ret = pthread_create(&handle, &attributes, csp_task_router, NULL);
```
`csp_task_router()` doesn't access the `param`.

In src/arch/posix/csp_bridge.c:
----
```c
ret = pthread_create(&handle, &attributes, csp_bridge, NULL);
```
`csp_bridge()` doesn't access the `param`

In examples/csp_server_client_posix.c:
----
```c
ret = pthread_create(&handle, &attributes, routine, NULL);
```
All functions passed as `routine` (`task_router`, `task_server`, and `task_client`) doesn't access the `param`.

johandc:
Thanks

Issue #233, opened 2021-09-24
Build libcsp with FreeRTOS on CI


yashi:
I think it's a good idea to at least build libcsp against FreeRTOS on CI.

@johandc, I'm not familiar with FreeRTOS but I'd like to, first, test it locally and deply it to CI.  If you have a simple build steps, would you mind to share it with me?  It doesn't have to be a comprehensive at all.

johandc:
I think there is some kind of posix port of freertos available, that could be used to run basic tests.

yashi:
csp_arch.c on FreeRTOS for POSIX is done.  I'd like to host it under libcsp if you are OK with it.

I wanted to cross build for ARM  so that we could test FreeRTOS for ARM on QEMU.   The blinky worked on QEMU but I couldn't integrate picolibc with it.   We can migrate to Non-POSIX later I assume.

https://github.com/spacecubics/libcsp-freertos

johandc:
Very nice indeed. Do you think it can run automatically with GitHub actions too?

I will setup the repo and provide you with write access

yashi:
Yes, I think I can.

johandc:
Done
https://github.com/libcsp/libcsp-freertos

PR #296, opened 2021-11-14
subprojet: picolibc: Disable multilib

yashi:
picolibc has the following block in its meson.build

    if not enable_multilib and meson.is_subproject()
      picolibc_dep = declare_dependency(include_directories: inc, link_with: [lib_c, lib_m])
    endif

We must disable multilib if we want to use `picolibc_dep`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -97,7 +97,8 @@ csp_c_args = ['-Wshadow',
 
 clib = meson.get_compiler('c').find_library('c', required: false)
 if not clib.found()
-	clib = dependency('libc', fallback: ['picolibc', 'picolibc_dep'], required: true)
+	clib = dependency('libc', fallback: ['picolibc', 'picolibc_dep'], required: true,
+			  default_options : ['multilib=false'])
 endif
 
 csp_deps += clib


johandc:
Yep i know.

I wrote those lines 2 years ago in picolibc :)
https://github.com/picolibc/picolibc/commit/6d3d8726bde2504bfda377882f7f0efad7443942#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4

A standalone CSP build with picolibc, is hard to envision. So picolibc is always part of the main project, which is responsible for configuring picolibc. The reason why we still need the dependency inside the CSP meson.build, is in order for the headers to be available.

Setting 'default_options' on the dependency, should normally does nothing. But i assume you have a specific test case, where it is needed, since you came across this? I don't mind merging, but i would like to know the reason.

yashi:
I was working on #233 and first wanted to cross compile to ARM with picolibc.  I thought I can't give `default_options` from top project.  But I could be wrong.  It didn't work out well and I've ditched it.  I only need it to test against API after all. So, I switched to POSIX version.

https://github.com/spacecubics/libcsp-freertos

johandc:
Here is the correct place to put options.

Put the subproject name in front so:
picolibc:multilib=false

johandc:
Forgot the link
https://github.com/spacecubics/libcsp-freertos/blob/341d291124507d4cc091ddb1de053f33c355f1bf/meson.build#L1

johandc:
Did you forget to add the .gitmodules file to that repo?

yashi:
Do we need it?  If we keep track which revisions to build against, it'd be useful. Otherwise, we can just shallow clone whatever we need.   If you like `git submodule`, I can certainly add it.

And thanks for the `default_options`.  I think I did that but... We'll see when we need it again.

yashi:
Let's leave it alone for now. I don't need it for #233.

yashi:
It looks like this PR has been merged.  Thanks :-)

johandc:
Actually I didn't merge it. I just pointed out how I would rather put the options in the top meson file in libcsp-freertos

PR #300, opened 2021-11-21
Broadcast WIP

johandc:
We wish to be able to send packets to all nodes on a particular subnet. For this purpose, we are going to reserve the highest address within each subnet to broadcast, just like IP does. This would work for any BUS where it's possible to listen for multiple addresses. That means CAN and ZMQhub currently.

- [x] Put nodeid on each interface
- [x] Put netmask on each interface
- [x] Make router respond (incoming) to multiple addresses
- [x] Add additional CAN filter for the broadcast address
- [x] Automatically add link local routes to the routing table
- [x] Get rid of global CSP address
- [x] Outgoing connections select source address based on matching routing entry
- [x] Loopback interface is used if destination matches the loopback address (which one? 0, comes to mind as it's available on both csp 1 and 2.)
- [x] Not all interfaces supports loopback on the wire (zmq does, can does not) so we could consider sending those to the loopback interface too

### DIFF
--- a/doc/basic.md
+++ b/doc/basic.md
@@ -267,23 +267,25 @@ no knowledge of any specific interface , protocol or driver:
    CSP interface.
 */
 struct csp_iface_s {
-    const char *name;          //!< Name, max compare length is #CSP_IFLIST_NAME_MAX
-    void * interface_data;     //!< Interface data, only known/used by the interface layer, e.g. state information.
-    void * driver_data;        //!< Driver data, only known/used by the driver layer, e.g. device/channel references.
-    nexthop_t nexthop;         //!< Next hop (Tx) function
-    uint16_t mtu;              //!< Maximum Transmission Unit of interface
-    uint8_t split_horizon_off; //!< Disable the route-loop prevention
-    uint32_t tx;               //!< Successfully transmitted packets
-    uint32_t rx;               //!< Successfully received packets
-    uint32_t tx_error;         //!< Transmit errors (packets)
-    uint32_t rx_error;         //!< Receive errors, e.g. too large message
-    uint32_t drop;             //!< Dropped packets
-    uint32_t autherr;          //!< Authentication errors (packets)
-    uint32_t frame;            //!< Frame format errors (packets)
-    uint32_t txbytes;          //!< Transmitted bytes
-    uint32_t rxbytes;          //!< Received bytes
-    uint32_t irq;              //!< Interrupts
-    struct csp_iface_s *next;  //!< Internal, interfaces are stored in a linked list
+	uint16_t addr;              // Host address on this subnet
+	uint16_t netmask;           // Subnet mask
+	const char * name;          // Name, max compare length is #CSP_IFLIST_NAME_MAX
+	void * interface_data;      // Interface data, only known/used by the interface layer, e.g. state information.
+	void * driver_data;         // Driver data, only known/used by the driver layer, e.g. device/channel references.
+	nexthop_t nexthop;          // Next hop (Tx) function
+	uint16_t mtu;               // Maximum Transmission Unit of interface
+	uint8_t split_horizon_off;  // Disable the route-loop prevention
+	uint32_t tx;                // Successfully transmitted packets
+	uint32_t rx;                // Successfully received packets
+	uint32_t tx_error;          // Transmit errors (packets)
+	uint32_t rx_error;          // Receive errors, e.g. too large message
+	uint32_t drop;              // Dropped packets
+	uint32_t autherr;           // Authentication errors (packets)
+	uint32_t frame;             // Frame format errors (packets)
+	uint32_t txbytes;           // Transmitted bytes
+	uint32_t rxbytes;           // Received bytes
+	uint32_t irq;               // Interrupts
+	struct csp_iface_s * next;  // Internal, interfaces are stored in a linked list
 };
 ```
 
--- a/src/csp_iflist.c
+++ b/src/csp_iflist.c
@@ -11,17 +11,10 @@
 /* Interfaces are stored in a linked list */
 static csp_iface_t * interfaces = NULL;
 
-csp_iface_t * csp_iflist_get_by_addr(uint16_t addr, int subnet_match) {
+csp_iface_t * csp_iflist_get_by_subnet(uint16_t addr) {
+
 	csp_iface_t * ifc = interfaces;
 	while (ifc) {
-
-		//printf("search if %s %u == %u ?\n", ifc->name, ifc->addr, addr);
-
-		/* Direct match on host address */
-		if (ifc->addr == addr) {
-			//printf("addr match\n");
-			return ifc;
-		}
 
 		/* Reject searches involving subnets, if the netmask is invalud */
 		if (ifc->netmask == 0) {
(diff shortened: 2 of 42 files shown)


johandc:
@yashi got anything to sha about this one?

yashi:
I'll do a quick review now.

yashi:
I'd love to have broadcast capability in CSP, BTW.

doc/basic.md
```diff
@@ -267,23 +267,25 @@ no knowledge of any specific interface , protocol or driver:
    CSP interface.
 */
 struct csp_iface_s {
-    const char *name;          //!< Name, max compare length is #CSP_IFLIST_NAME_MAX
-    void * interface_data;     //!< Interface data, only known/used by the interface layer, e.g. state information.
-    void * driver_data;        //!< Driver data, only known/used by the driver layer, e.g. device/channel references.
```
yashi, line 272: It's quite tough to find what really changed if you mix non white space changes.

src/csp_iflist.c
```diff
@@ -10,15 +11,53 @@
...
 	csp_iface_t * ifc = interfaces;
 	while (ifc) {
 		if (strncmp(ifc->name, name, CSP_IFLIST_NAME_MAX) == 0) {
-			break;
+			return ifc;
 		}
 		ifc = ifc->next;
 	}
-	return ifc;
+	return NULL;
```
yashi, line 62: why do you need to change this?  It should work as is, no?

src/csp_io.c
```diff
@@ -122,15 +122,15 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
 	target->flags = source->flags;
 }
 
-int csp_send_direct(csp_id_t idout, csp_packet_t * packet, const csp_route_t * ifroute) {
+int csp_send_direct(csp_id_t idout, csp_packet_t * packet, const csp_route_t * ifroute, int from_me) {
```
yashi: must be `bool` instead of `int from_me`.

doc/basic.md
```diff
@@ -267,23 +267,25 @@ no knowledge of any specific interface , protocol or driver:
...
-    uint32_t tx_error;         //!< Transmit errors (packets)
-    uint32_t rx_error;         //!< Receive errors, e.g. too large message
-    uint32_t drop;             //!< Dropped packets
-    uint32_t autherr;          //!< Authentication errors (packets)
-    uint32_t frame;            //!< Frame format errors (packets)
-    uint32_t txbytes;          //!< Transmitted bytes
-    uint32_t rxbytes;          //!< Received bytes
-    uint32_t irq;              //!< Interrupts
-    struct csp_iface_s *next;  //!< Internal, interfaces are stored in a linked list
+	uint16_t addr;              // Host address on this subnet
```
yashi, line 270: I was gonna say that `address` is used in other structures.  But, some uses `addr` as well.  oh well.

src/csp_iflist.c
```diff
@@ -10,15 +11,53 @@
 /* Interfaces are stored in a linked list */
 static csp_iface_t * interfaces = NULL;
 
+csp_iface_t * csp_iflist_get_by_addr(uint16_t addr) {
```
yashi: Some idea:
```c
#define CSP_NETWORK(addr) (addr & csp_id_get_netmask())

static bool is_same_network(uint16_t addr1, uint16_t addr2)
{
	return CSP_NETWORK(addr1) == CSP_NETWORK(addr2);
}

static bool is_broadcast(uint16_t addr)
{
	return (addr & csp_id_get_hostmask()) == csp_id_get_hostmask();
}

csp_iface_t * csp_iflist_get_by_addr(uint16_t addr) {
	csp_iface_t * ifc = interfaces;

	while (ifc) {
		if (addr == ifc->addr) {
			break;
		}
		if (is_same_network(addr, ifc->addr) &&
		    is_broadcast(addr)) {
			break;
		}
		ifc = ifc->next;
	} 
		
	return ifc;
}
```
Totally untested.  I think we can work on macros and inline functions after the merge alone with `csp_rtable_find_route()`, which does pretty match the same thing, no?

yashi (commented):
Some questions

1. Is it OK to have multiple entries for the same interface?  That is, interface A for network 1 and network 2.  This is related to the next question.
2. how does the address specified by `via` on `csp_rtable_set()` handled?
3. I feel we should move encryption routines out from `csp_send_direct()`.  It's fine that `csp_send()` encrypts the packet for users.  But for `csp_send_direct()` or any internal function should do one thing and only one thing.  If we have encryption a function, `csp_send()` can encript the packet and send it via `csp_send_direct()`, routers can use `csp_send_direct()` without `from_me`.

yashi:
a bit more on 3.  If `csp_send_direct()` is also user facing API and want it to handle encryption, we should create a new internal api and make `csp_send_direct()`:
```c
int csp_send_direct(...) {
    encript();
    __csp_send_direct(...);
}
```
Then, make routers to use `__csp_send_direct()`.

yashi:
One big different from IP broadcast is that CSP routers, by default, forward broadcast packets to the destination subnet, right?  We might need a big fat warning about it in the doc.  A configuration nob on each router whether or not to route broadcast packet might help?  Otherwise we have to allocate a subnet for GS all the time?

src/csp_iflist.c
johandc: Yes, but it's better code practice to return immediately with the value. Instead of breaking, and assuming the return later on will work. For example, if anyone were to put code in between, they wouldn't necessary know not to modify the ifc.

src/csp_io.c
johandc: I don't like bool's. because the extra hassle of including stdbool everywhere. bool and int have same size, and is more portable. (reduces header clutter).

johandc: The day bool becomes a first class type we can talk about it. But it never will :)

src/csp_iflist.c
johandc: I like what you did for readability, however. csp_id_get_hostmask() isn't really a global. So it needs to be calculated from the interface configuration
```c
uint16_t netmask = (1 << (csp_id_get_host_bits() - ifc->netmask)) - 1;
```

I just added an update to the function, so it can match on the entire subnet, for outgoing traffic, and on just the node or the broadcast addr for incoming traffic.

johandc:
> One big different from IP broadcast is that CSP routers, by default, forward broadcast packets to the destination subnet, right? We might need a big fat warning about it in the doc. A configuration nob on each router whether or not to route broadcast packet might help? Otherwise we have to allocate a subnet for GS all the time?

Excellent observation skills! With CSP we will allow broadcast traffic to be routed. This means that if i want to ping any node on the satellite bus, i can ping the broadcast address of that subnet. Which is a huge feature :)

Regarding the configuration knob, i think you are mistaking this from layer-2 broadcast, which we definitely will not want to broadcast across the router. So this til always be rejected.

An example of layer-2 broadcast is a special packet that all nodes on CAN will receive, regardless of their CSP configuration. Something that i want to add, in order to recover mis-configured nodes. But it will never reach the router. Layer-2 broadcast should be rewritten so that the incoming address matches that of the interface it was received on, and then passed to the router queue. It will therefore always be received as "to_me", and thus also never possible to route.

doc/basic.md
johandc: Yes, i know. I had to fix some documentation too

doc/basic.md
johandc: Yes i noticed. I like iface and addr for shortness. I may just refactor the code a bit later on.

johandc:
> 1. Is it OK to have multiple entries for the same interface?  That is, interface A for network 1 and network 2.  This is related to the next question.
> 2. how does the address specified by `via` on `csp_rtable_set()` handled?

I'm a little confused by these two. But if you are referring to routing table entries then yes. Multiple routes can point to the same interface.

The 'via' address is used for the layer-2 to rewrite the destination address on the MAC layer. For example on I2C, where a node can only listen on once address, the router's cannot just go into promiscious mode to feth all traffic. So each node needs to rewrite and know the gateway MAC address. 

This is actually only needed for I2C at the moment. But one could imaging it being used for Ethernet and even CAN. But on CAN i have decided it's so low speed, that i will just use promiscious mode on the radios and filter by routing table. This also removes the configuration of redundant radios into the radios, and not spread out on the network nodes to chante the address of the 'via'. or 'gw addr'.


> 3. I feel we should move encryption routines out from `csp_send_direct()`.  It's fine that `csp_send()` encrypts the packet for users.  But for `csp_send_direct()` or any internal function should do one thing and only one thing.  If we have encryption a function, `csp_send()` can encript the packet and send it via `csp_send_direct()`, routers can use `csp_send_direct()` without `from_me`.

Yes, encryption should just be removed. And i actually just did a split for send_direct, because some packets should be sent based on a routing entry, while others do not have a routing entry and should be sent based on just the interface. A little messy, but it works :)

a final note is that send_direct is only available within CSP not outside.

PR #301, opened 2021-11-21
Yaml

johandc:
Added yaml configuration format

For our Posix/C based tools (satctl), we have a plethora of command line options to try and configure various interfaces. It doesn't get better now that each interface should have its own address and netmask. So in order to ease configuration i have added a small yaml file.

I wanted to add this to CSP because i belive other tools may benefit from this too.

It depends on libyaml, and automatically byilds the functions if it's available on the system. It also depends on malloc.

oh and btw. the implementation is lossy, i haven't bothered cleaning up string's and such, because it's an init time tool only and mostly running on PC. I should probably add a note to the source file about this, before someone comes yelling at me :)

### DIFF
--- a/include/csp/csp_iflist.h
+++ b/include/csp/csp_iflist.h
@@ -1,18 +1,6 @@
 #pragma once
 
-/**
-   @file
-
-   Interface list.
-
-   Linked-list of interfaces in the system.
-
-   This API is not thread-safe.
-*/
-
 #include <csp/csp_interface.h>
-
-
 
 /**
    Add interface to the list.
@@ -22,27 +10,13 @@
 */
 int csp_iflist_add(csp_iface_t * iface);
 
-/**
-   Get interface by name.
-
-   @param[in] name interface name.
-   @return Interface or NULL if not found.
-*/
 csp_iface_t * csp_iflist_get_by_name(const char *name);
 
-/**
-   Print list of interfaces to stdout.
-*/
+/* Includes match on broadcast address */
+csp_iface_t * csp_iflist_get_by_addr(uint16_t addr);
+
 void csp_iflist_print(void);
-
-/**
-   Return list of interfaces.
-
-   @return First interface or NULL, if no interfaces added.
-*/
 csp_iface_t * csp_iflist_get(void);
 
-/**
-   Convert bytes to readable string.
-*/
+/* Convert bytes to readable string */
 int csp_bytesize(char *buffer, int buffer_len, unsigned long int bytes);
--- a/include/csp/csp_iflist.h
+++ b/include/csp/csp_iflist.h
@@ -11,9 +11,8 @@
 int csp_iflist_add(csp_iface_t * iface);
 
 csp_iface_t * csp_iflist_get_by_name(const char *name);
-
-/* Includes match on broadcast address */
-csp_iface_t * csp_iflist_get_by_addr(uint16_t addr, int subnet_match);
+csp_iface_t * csp_iflist_get_by_addr(uint16_t addr);
+csp_iface_t * csp_iflist_get_by_subnet(uint16_t addr);
 
 void csp_iflist_print(void);
 csp_iface_t * csp_iflist_get(void);
--- a/include/csp/csp_iflist.h
+++ b/include/csp/csp_iflist.h
@@ -20,3 +20,6 @@ csp_iface_t * csp_iflist_get(void);
 
 /* Convert bytes to readable string */
 int csp_bytesize(char *buffer, int buffer_len, unsigned long int bytes);
+
+/* From iflist_yaml.c */
+void iflist_yaml_init(void);
--- a/include/csp/csp_iflist.h
+++ b/include/csp/csp_iflist.h
@@ -13,7 +13,7 @@ int csp_iflist_add(csp_iface_t * iface);
 csp_iface_t * csp_iflist_get_by_name(const char *name);
 
 /* Includes match on broadcast address */
-csp_iface_t * csp_iflist_get_by_addr(uint16_t addr);
+csp_iface_t * csp_iflist_get_by_addr(uint16_t addr, int subnet_match);
 
 void csp_iflist_print(void);
 csp_iface_t * csp_iflist_get(void);
--- a/include/csp/csp_iflist.h
+++ b/include/csp/csp_iflist.h
@@ -22,4 +22,4 @@ csp_iface_t * csp_iflist_get(void);
 int csp_bytesize(char *buffer, int buffer_len, unsigned long int bytes);
 
 /* From iflist_yaml.c */
-void iflist_yaml_init(void);
+void iflist_yaml_init(char * filename);
--- a/meson.build
+++ b/meson.build
@@ -45,22 +45,9 @@ conf.set10('CSP_USE_XTEA', get_option('use_xtea'))
 conf.set10('CSP_USE_PROMISC', get_option('use_promisc'))
 conf.set10('CSP_USE_DEDUP', get_option('use_dedup'))
 
-csp_sources = []
 csp_deps = []
 
-zmq_dep = dependency('libzmq', version: '>4.0', required: false)
-if zmq_dep.found()
-	conf.set('CSP_HAVE_LIBZMQ', 1)
-endif
-
-yaml_dep = dependency('yaml-0.1', required : false)
-if yaml_dep.found()
-	conf.set('CSP_HAVE_LIBYAML', 1)
-	csp_sources += files(['src/yaml/iflist_yaml.c'])
-	csp_deps += yaml_dep
-endif
-
-csp_sources += files([
+csp_sources = files([
 	'src/crypto/csp_hmac.c',
 	'src/crypto/csp_sha1.c',
 	'src/crypto/csp_xtea.c',
@@ -86,6 +73,7 @@ csp_sources += files([
 	'src/csp_sfp.c',
 ])
 
+subdir('src/yaml')
 subdir('src/rtable')
 subdir('src/arch')
 subdir('src/drivers')
--- a/meson.build
+++ b/meson.build
@@ -45,12 +45,22 @@ conf.set10('CSP_USE_XTEA', get_option('use_xtea'))
 conf.set10('CSP_USE_PROMISC', get_option('use_promisc'))
 conf.set10('CSP_USE_DEDUP', get_option('use_dedup'))
 
+csp_sources = []
+csp_deps = []
+
 zmq_dep = dependency('libzmq', version: '>4.0', required: false)
 if zmq_dep.found()
 	conf.set('CSP_HAVE_LIBZMQ', 1)
 endif
 
-csp_sources = files([
+yaml_dep = dependency('yaml-0.1', required : false)
+if yaml_dep.found()
+	conf.set('CSP_HAVE_LIBYAML', 1)
+	csp_sources += files(['src/yaml/iflist_yaml.c'])
+	csp_deps += yaml_dep
+endif
+
+csp_sources += files([
 	'src/crypto/csp_hmac.c',
 	'src/crypto/csp_sha1.c',
 	'src/crypto/csp_xtea.c',
@@ -76,7 +86,6 @@ csp_sources = files([
 	'src/csp_sfp.c',
 ])
 
-csp_deps = []
 subdir('src/rtable')
 subdir('src/arch')
 subdir('src/drivers')
--- a/src/yaml/iflist_yaml.c
+++ b/src/yaml/iflist_yaml.c
@@ -1,7 +1,6 @@
 
 #include "iflist_yaml.h"
 
-#include <assert.h>
 #include <csp/csp_iflist.h>
 #include <csp/csp_interface.h>
 #include <csp/csp_rtable.h>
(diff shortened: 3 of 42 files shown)


src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+#include <csp/interfaces/csp_if_zmqhub.h>
+#include <csp/interfaces/csp_if_can.h>
+#include <csp/interfaces/csp_if_lo.h>
+#include <csp/drivers/can_socketcan.h>
+#include <csp/drivers/usart.h>
+#include <stdio.h>
+#include <yaml.h>
+
+struct {
+    char *name;
```
yashi: perhaps * at the middle?
I guess we need to `clang-format` this file?

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+    FILE *file = fopen("iflist.yaml", "rb");
+    if (file == NULL)
+        return;
+
+    yaml_parser_t parser;
+    yaml_parser_initialize(&parser);
+    yaml_parser_set_input_file(&parser, file);
+    yaml_event_t event;
+
+    assert(yaml_parser_parse(&parser, &event));
```
yashi: Please move the functions out of `assert()`.  It should be tested with and without assert.

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+                break;
+            } else {
+                value = event.data.scalar.value;
+            }
+
+            iflist_yaml_key_value((char *)key, (char *)value);
+            continue;
+        }
+    }
+}
```
yashi: no newline at the end.

meson.build
```diff
@@ -45,12 +45,22 @@ conf.set10('CSP_USE_XTEA', get_option('use_xtea'))
...
+csp_sources = []
+csp_deps = []
+
 zmq_dep = dependency('libzmq', version: '>4.0', required: false)
 if zmq_dep.found()
 	conf.set('CSP_HAVE_LIBZMQ', 1)
 endif
 
-csp_sources = files([
+yaml_dep = dependency('yaml-0.1', required : false)
```
yashi: I'd write
```meson
yaml_dep = dependency('yaml-0.1', required : false)
conf.set('CSP_HAVE_LIBYAML', yaml_dep.found())
csp_deps += yaml_dep
```
And move `csp_sources += files(['src/yaml/iflist_yaml.c'])` down to `src/yaml/`.  I wanted to other `csp_sources` to where the source files are, as well.

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+            printf("failed to add CAN interface [%s], error: %d", data.device, error);
+            return;
+        }
+    }
+
+    iface->addr = atoi(data.addr);
+    iface->netmask = atoi(data.netmask);
+
+    csp_rtable_set(atoi(data.addr), atoi(data.netmask), iface, CSP_NO_VIA_ADDRESS);
+	csp_rtable_set(atoi(data.addr), csp_id_get_host_bits(), &csp_if_lo, CSP_NO_VIA_ADDRESS);
```
yashi: Indent.

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+
+struct {
+    char *name;
+    char *driver;
+    char *device;
+    char *addr;
+    char *netmask;
+    char *server;
+    char *is_dfl;
+} data;
```
yashi: I wouldn't use a global variable.  pass it to the functions accessing it.  We can create it as a local variable in `iflist_yaml_init()`

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+static void iflist_yaml_end_if(void) {
+    /* Sanity checks */
+    if ((!data.name) || (!data.driver) || (!data.addr)) {
+        printf("interface is missing, name, driver or addr\n");
+        return;
+    }
+
+    csp_iface_t *iface;
+
+    /* ZMQ */
```
yashi: `#if (CSP_HAVE_LIBZMQ)`

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+
+        printf("server: %s\n", data.server);
+
+        csp_zmqhub_init(atoi(data.addr), data.server, 0, &iface);
+        strncpy((char*) iface->name, data.name, CSP_IFLIST_NAME_MAX);
+
+        csp_iflist_add(iface);
+    }
+
+    /* CAN */
```
yashi: `#if (CSP_HAVE_LIBSOCKETCAN)`

yashi: Or we can perhaps move this block to libsocketcan driver.  Make it so that in the header file

```c
#if (CSP_HAVE_LIBSOCKETCAN)
extern int csp_can_socketcan_set_if(...);  // or what ever the name is
#else
static inline int csp_can_socketcan_set_if() {}  // make it noop when disabled.
#endif
```

If we have these interfaces installed. this part of the file will be

```c
if (strcmp(data.driver, "zmq") == 0) {
    csp_zmphub_set_if(...);
} else if (strcmp(data.driver, "can") == 0) {
    csp_can_socketcan_set_if(...);
} else {
    log_error("Unsupported interface");
}
```

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+            continue;
+        }
+
+        if (event.type == YAML_SCALAR_EVENT) {
+            yaml_char_t *key = event.data.scalar.value;
+
+            yaml_char_t *value;
+            yaml_parser_parse(&parser, &event);
+            if (event.type != YAML_SCALAR_EVENT) {
+                printf("Extected value\n");
```
yashi: typo?

src/yaml/iflist_yaml.c
```diff
@@ -0,0 +1,166 @@
...
+            iflist_yaml_end_if();
+            continue;
+        }
+
+        if (event.type == YAML_SCALAR_EVENT) {
+            yaml_char_t *key = event.data.scalar.value;
+
+            yaml_char_t *value;
+            yaml_parser_parse(&parser, &event);
+            if (event.type != YAML_SCALAR_EVENT) {
```
yashi: I couldn't reproduce this case.  Do you have a data file for this?

yashi (changes requested):
Is there public repo for satctl?  It'd help if I can test it when I review.

I'd make a simple state machine for `iflist_yaml_init()`.  The current implementation takes any scalar event.  The state machine is very simple:

1. document start
2. sequence
3. mapping
4. scalar (as key)
5. another scalar (as value) (go back to 4)
6. many end sections

But we can fix it when we need it.

Everything else is nitpicks.

src/yaml/iflist_yaml.c
johandc: Right, that came directly from an online example 😆

meson.build
johandc: Okay, i agree

src/yaml/iflist_yaml.c
johandc: Good Idea

src/yaml/iflist_yaml.c
johandc: Oh, yes.

src/yaml/iflist_yaml.c
johandc: I think it would be best to produce a warning, when trying to initialise an unsupported interface. I actually think it's nice of the header would simply take care of that, instead of clobbering the code with checks.

src/yaml/iflist_yaml.c
johandc: Yes

johandc:
Thanks for a very good review. I have been focusing on making stuff work today. Your help, in getting the details right is much appreciated.

> Is there public repo for satctl? It'd help if I can test it when I review.

Yes. It's open source and on: 
github.com/spaceinventor/satctl

> 
> I'd make a simple state machine for `iflist_yaml_init()`. The current implementation takes any scalar event. 

Haha, i actually threw away the state machine from the reference implementation and reduced number of line from 300 to 100 this way. This is much easier for the eyes, and works just as well when the schema is predefined like here. If it sees anything not as it expects it just quits. 

So in my opinion, unless we need to parse arbitrary trees into objects, tjen a state machine isn't needed.

What do you think about the idea of the yaml file in general? I think the network configuration just got more difficult, in that we have more parameters, and yet again it was a mess with all the old runtime flags.

I actually want to add routing table, and other CSP init parameters there too. And have a flag to select which config file to start satctl with.

As for the embedded use, none of this yaml nonsense hokus pokus is required. Since the amount of network interfaces is typically fixed. And parameters can be used to initialise them in static ram.

yashi:
> What do you think about the idea of the yaml file in general?

To tell the truth, I don't know.  I haven't setup a complex CSP network yet.  But given that CSP 2.0 supports 16k addresses, I'm sure we need some mechanism to setup everything right.  I don't know how you guys manage the network settings for many cubesats with many components.  I'd love to learn it from you.

My current idea is to have a single file to describe the whole network of cubesat including GS (emulated by PC).

Say you have five components in a cubesat, two of them are purchased from other companies, three are your own building from source.  How would you make sure that network configuration on all components are right?  Should we extend this yaml capability to setup everything?  Or is it totally different issue?

I think we can generate binary blob from yaml describing routing tables and link it against the final image for embedded systems.  We can also extend `satctl` to update network setting of nodes, if it doesn't have the capability yet.  (or is it handle by [spaceboot](https://github.com/spaceinventor/spaceboot)?) But I'm not sure the single file for cubesat network is a good way to go.

If this yaml conf is only for PC, we can go wild :wink: 

Oh, it might be a good idea to pass a file name from a tool, such as `satctl`, to the API.

> Yes. It's open source and on: [github.com/spaceinventor/satctl](https://github.com/spaceinventor/satctl)

Thanks :+1:

johandc:
> To tell the truth, I don't know. I haven't setup a complex CSP network yet. But given that CSP 2.0 supports 16k addresses, I'm sure we need some mechanism to setup everything right. I don't know how you guys manage the network settings for many cubesats with many components. I'd love to learn it from you.

We go from the top down.
0) Select CSP version. If you have legacy products and are forced go with v1. otherwise select v2.

1.1) How many satellites will be in the constellation = min number of subnets
1.2) How many nodes will be on each satellite = min number of host addresses.
1.3) Select a subnet mask that falls in between 1.1 and 1.2
note: We typically use 8 as netmask, allowing for 256 subnets, with 64 nodes per satellite.

2.1) For each satellite-bus assign a subnet
2.2) For each point-to-point linik assign a subnet (these can have much higher netmask 12 for example)
2.3) For RF links, decide wether they operate in a BUS (all with same RX/TX freq. and conf) or point-to-point, each link individual. Assign subnet of the size that matches to a bus or point to point.
2.4) For encrypted links assign tunnel endpoints (csp_if_tun) (more advanced than this)
2.5) Make a topology diagram showing only the subnets and how they are inter-connected, possibly the gateway and their addresses. Take special notice of cases where redundant gateways exist. For example having a satellite with multiple receivers, or a ground station network with multiple stations.

3.1) Within each subnet assign individual node addresses. These are static for the entire mission duration and can be input into the nodes by config parameters. The gorund segment can be more loose with its client addressing, since mission control clients can typically spawn at will.

 
> My current idea is to have a single file to describe the whole network of cubesat including GS (emulated by PC).

That would be awesomme if combined with a tool to create graphs.
https://graph-tool.skewed.de/static/doc/draw.html

 
> Say you have five components in a cubesat, two of them are purchased from other companies, three are your own building from source. How would you make sure that network configuration on all components are right? 

I would make it a requirement that:
1) The systems support either CSP 1 or 2.
2) The systems come with a means of changing the node-id-
3) If 2 is not possible. They must be ordered with the correct address. So make very sure it's the final decision before ordering.

Should we extend this yaml capability to setup everything? Or is it totally different issue?

I don't think the topology description belongs to the satellite, or configuration format. Each node should have their separate config (on PC via YAML, on satellite via fixed parameters)

 
> I think we can generate binary blob from yaml describing routing tables and link it against the final image for embedded systems. 

CSP already supports reconfigurable routing table through the `csp_rtable` parameter. It's a text string which are parsed by CSP in the CSV format "node/mask interface, ...."

> We can also extend `satctl` to update network setting of nodes, if it doesn't have the capability yet. 

`satctl` can modify parameters on nodes, and therefore also modify the CSP configuration. Not that it's dangerous to mess with while in orbit. Therefore a special recovery procedure is proposed to become part of CSP. (i haven't written the spec yet, but it relies on layer-2 broadcast, which i'm getting to)

> (or is it handle by [spaceboot](https://github.com/spaceinventor/spaceboot)?) 

No, spaceboot is just for image transfer.

> If this yaml conf is only for PC, we can go wild wink

If the libyaml wasnt a big monster, we could use it for the satellite. But i think it's a waste of energy, since most subsystems are pretty static.
 
> Oh, it might be a good idea to pass a file name from a tool, such as `satctl`, to the API.

Yes.

benner:
> 1.1) How many satellites will be in the constellation = min number of subnets
> 1.2) How many nodes will be on each satellite = min number of host addresses.
> 1.3) Select a subnet mask that falls in between 1.1 and 1.2
> note: We typically use 8 as netmask, allowing for 256 subnets, with 64 nodes per satellite.
> 
> 2.1) For each satellite-bus assign a subnet
> 2.2) For each point-to-point linik assign a subnet (these can have much higher netmask 12 for example)
> 2.3) For RF links, decide wether they operate in a BUS (all with same RX/TX freq. and conf) or point-to-point, each link individual. Assign subnet of the size that matches to a bus or point to point.
> 2.4) For encrypted links assign tunnel endpoints (csp_if_tun) (more advanced than this)
> 2.5) Make a topology diagram showing only the subnets and how they are inter-connected, possibly the gateway and their addresses. Take special notice of cases where redundant gateways exist. For example having a satellite with multiple receivers, or a ground station network with multiple stations.
> 
> 3.1) Within each subnet assign individual node addresses. These are static for the entire mission duration and can be input into the nodes by config parameters. The gorund segment can be more loose with its client addressing, since mission control clients can typically spawn at will.

Awesome addressing/network planing strategy.  This can be placed outside PR/issue and put into docs or other findable place.

src/yaml/iflist_yaml.c
johandc: Hmm, yes it seems that if no value is given, libyaml will automatically create a scalar value of 0. Which means it's impossible to create a parse error. So i'm removing the check for cleaner code.

johandc:
All suggestions is implemented, but i'm waiting with the merge because this was based on the broadcast branch. And i'm expecting the broadcast branch to be merged soon.

include/csp/csp_iflist.h
```diff
@@ -22,4 +22,4 @@ csp_iface_t * csp_iflist_get(void);
 int csp_bytesize(char *buffer, int buffer_len, unsigned long int bytes);
 
 /* From iflist_yaml.c */
-void iflist_yaml_init(void);
\ No newline at end of file
+void iflist_yaml_init(char * filename);
```
yashi, line 24: ending newline is missing.  I searched setting for clang-format but couldn't find it.  Do you know whether it supports it or not?

meson.build
```diff
@@ -45,22 +45,9 @@ conf.set10('CSP_USE_XTEA', get_option('use_xtea'))
 conf.set10('CSP_USE_PROMISC', get_option('use_promisc'))
 conf.set10('CSP_USE_DEDUP', get_option('use_dedup'))
 
-csp_sources = []
 csp_deps = []
 
-zmq_dep = dependency('libzmq', version: '>4.0', required: false)
-if zmq_dep.found()
-	conf.set('CSP_HAVE_LIBZMQ', 1)
-endif
```
yashi, line 51: We should note about this in the commit log that this has been removed because we already have it under interface.

src/yaml/iflist_yaml.c
```diff
@@ -14,153 +13,165 @@
...
+		return;
+	}
 
-    csp_iface_t *iface;
+	csp_iface_t * iface;
 
-    /* ZMQ */
-    if (strcmp(data.driver, "zmq") == 0) {
-        printf("Adding ZMQ interface: %s node %s\n", data.name, data.addr);
+    if (0) {
```
yashi, line 39: I'd add
```c
    /* We must have this empty if for the last else block */
```

src/yaml/iflist_yaml.c
```diff
@@ -14,153 +13,165 @@
...
 
-        /* Check for valid server */
-        if (!data.server) {
-            printf("no server configured\n");
-            return;
-        }
+#if (CSP_HAVE_LIBZMQ)
+	/* ZMQ */
+    } else if (strcmp(data->driver, "zmq") == 0) {
+		printf("Adding ZMQ interface: %s node %s\n", data->name, data->addr);
```
yashi, line 44: how about
```c
	/* We must have this empty if for the last else block */
	if (0) {
	}
#if (CSP_HAVE_LIBZMQ)
	/* ZMQ */
	else if (strcmp(data->driver, "zmq") == 0) {
	    :
	}
#endif
#if (CSP_HAVE_LIBSOCKETCAN)
	/* CAN */
	else if (strcmp(data->driver, "can") == 0) {
	    :
	}
#endif
	/* Unsupported interface */
	else {
	    :
	}
````
This is more consistent.

File: src/yaml/iflist_yaml.h [hunk 15]
```diff
@@ -1,3 +1,3 @@
 #pragma once
 
-void iflist_yaml_init(void);
\ No newline at end of file
+void iflist_yaml_init(char * filename);
```
yashi, line 3: missing end-of-file newline.

src/yaml/iflist_yaml.c
```diff
@@ -14,153 +13,165 @@
...
+			printf("failed to add CAN interface [%s], error: %d", data->device, error);
+			return;
+		}
 
-    if (iface && data.is_dfl) {
-        printf("Setting default route to %s\n", iface->name);
-        csp_rtable_set(0, 0, iface, CSP_NO_VIA_ADDRESS);
+#endif
+
+    /* Unsupported interface */
```
yashi, line 82: I think you missed running clang-format again after you edited?  It has spaces instead of a tab.

yashi (approved):
Looks good to me. Thanks.

I've added a few comments for me asking how you want the style to be.  Nothing blocking.

BTW, does your editor support squashing this commit into the commit "Added yaml configuration format" (31d5223cfb1517519) ?  Or do you like to have a separate commit for review response?

johandc:
I normally don't do squashing, it's mainly used by larger projects to keep each pull request to single commit messages in the log. Which is nicer or course. But on a day to day basis on a project I mainly maintain myself, o don't care so much about the commit log. I even do regular merge commits too. Faster, gets the job done. Next issue.

I would like to get my discipline better some day, but I'm lacking the personal incentive to spend time learning the squash. Furthermore I feel like it rewrites history a little too much for a little project like this.

Your other comments are duly noted, at some point I feel like a script to apply clang format to the entire tree will be useful.

johandc:
@yashi  @jledet  
I apologise, that i mentioned 'satctl' as an open source program. We didn't have the right to that name so it has been changed to https://github.com/spaceinventor/spacepromt instead.
Best, Johan.

Issue #242, opened 2021-09-27
lightweight listener sockets


johandc:
For listeners only:

Use small statically allocated listener sockets. They are bound to CSP using the ports[] array, and therefore does NOT need to come from the CONN table.

A listener socket can be even smaller when it's for a connection less socket (basically a direct callback)
This allows for a huge memory saving.

Issue #241, opened 2021-09-27
RDP retransmission queue


johandc:
RDP transmissions should have a single global TX queue for easier memory management. Typically only a single transfer is in progress at a time, and there is a lot of memory overhead on having each connection with full RDP support.

For the RX queue for not yet delivered out-of-order packets, RDP could also use a single queue. It costs a little extra to scan, but the memory footprint improvement is quite large.

Issue #234, opened 2021-09-25
csp_mutex_create_static() and csp_bin_sem_create_static() should return int


yashi:
Many csp function return values indicating success for failure. New functions, `csp_mutex_create_static()` and `csp_bin_sem_create_static()` should also follow the semantic.  I know both function has an out parameter, which indicate success for failure, but that's only for FreeRTOS.  POSIX and Zephyr won't need the second argument but nice to return `int`.

If you agree, I can post an PR.

yashi:
I might have sent too early.  Thread creation also returns a handle.  Let me think about it a bit more.

johandc:
I don't agree.

The reason is this. For static allocation there is no way a an allocation can fail. So there is no reason to make any error checking, which is actually the primary reason for using static allocation in the first place.

Using void where possible is much nicer and makes for cleaner and smaller code.

johandc:
As an example I want to make CSP send a void function too. The concept that once CSP send has been called that packet is out of here is much nicer than having to check for the case where an outgoing buffer was full, on every single call to send. 

Full output buffers should be handled by proper flow control instead. There is also a bug with CSP send having an inconsistent return type. So both can be solved.

yashi:
I agree that we want them to be allocation free.  But there are other restriction in RTOS.  For example,  [RTEMS sem_init()](https://docs.rtems.org/branches/master/posix-users/semaphore.html#sem-init) returns status code and has the following cases:
- **EINVAL**: The value argument exceeds SEM_VALUE_MAX (this is not a case for CSP because of binary nature)
- **ENOSPC**: A resource required to initialize the semaphore has been exhausted The limit on semaphores (SEM_VALUE_MAX) has been reached
- **ENOSYS**: The function sem_init is not supported by this implementation (this is not a case for CSP because all arch supports it)
- **EPERM**: The process lacks appropriate privileges to initialize the semaphore

QNX: http://www.qnx.com/developers/docs/qnxcar2/index.jsp?topic=%2Fcom.qnx.doc.neutrino.sys_arch%2Ftopic%2Fkernel_Semaphores.html

In practice it doesn't really fail as documented, so yes I agree them can return void.  I'll reopen it when I encounter a real problem.  Thank you for your time.

johandc:
ENOSPC: It seems like a poor implementation in RTEMS that it's not possible to pass the static memory along to the sem_init call. 

[...]

Issue #209, opened 2021-07-24
Unify return value and errors


yashi:
As #23 suggested, we should use unified scheme for error numbers and return values.  We have `CSP_ERROR_NONE` defined as `0` in `csp_error.h` but `csp_queue`, `csp_semaphore`, and `csp_mutex` uses `1`, or `pdTRUE` / `pdPASS` as OK.

Because this is API breaking changes, it should only be for 2.0, as #23 suggested.

I'm creating this new issue becuase #23 is already closed.

johandc:
This will have to wait until a 2.1 or subsequent release. Since 2.0 is already in release candidate.

Theres a lot of cleanup that I would like to perform if I had the change to do over. 😊

yashi:
I didn't know 2.0 is that close.  There is no RC tag yet, right?

Would you give v2.1, or v3.0, tag to this?  Or did you mean we won't fix this?  I'm sure it's a huge change.

johandc:
I think we want to go with zero as OK and negative values as error.

This means the change needs to be in csp/arch which is used much more seldom than csp io. But yes it IS somewhat of a bigger change.

I don't want to ignore it, because it IS an inconvenient inconsistency. So I would like to address some of these internal stuff in a major release. 

The next API break will be in 2.1

johandc:
I fixed the csp_send return value issue now

johandc:
csp_mutex is gone now too

johandc:
The CSP_QUEUE_OK changed was non breaking, since only the macros was used it was only a matter of changing the value, and fixing the return of the FreeRTOS queue functions.

PR #304, opened 2021-11-27
Space Inventor Master

### LINKED ISSUE #221: Simplify assert and stdout from library

johandc:
Fixes #221, #242, #241

The CSP core never emits calls to printf, instead it uses error counters for typical errors. And a global errno with codes for rare errors. The packet and protocol logs have been converted into flags, which enable the printf to stdout of these two. lock debugging have been wiped, i didn't feel it was necessary or used ever.

The new error counters will be able to catch many more overflow, and underruns than the old string based error log system. While the errno should simply always be 0. If it's not something is not setup correctly or there is a bug in the code.

The can interface gets it's own errno, because an error on the CAN bus isn't necessarily a bug within CSP. But rather on the bus.

The way to access these variables is to include <csp/csp_debug.h> from your application.

One final note: All posix stuff, like socketcan, uart_linux, and zmq just uses printf directly as printf is assumed to be available on posix.

PART 2:

**do not use csp_conn objects for listener sockets**
We now demand the user allocate a csp_socket_t, then calls csp_bind, csp_listen and csp_accept on that socket as per usual.

a call to csp_socket() is not longer necessary.
sock->opts can still be set manually.

Rationale: csp_conn is a very  heavy object, namely because of the rx queue and the RDP rx and tx queues. For listener sockets, all that is really neaded is a single queue to accept new connections on. This can be somewhat smaller than the conn rx_queue, and also there is no need for RDP, idin, idout and all that jazz.

Callbacks will now be bound directly to the port number, instead of through a socket object. Saving even more memory.

PART 3:
Uses global RDP retransmission queue.
It was not neceassary for each connection to have it's own RDP queue. So we flipped it around, so there is one RX and one TX queue for RDP, and it serches for the conn id on each queued packet. This saves A LOT of memory too.


bfol (changes requested):
From Gomspace point of view, we have 2 main concerns. 
1) The pollution of stdout with printf. We prefer to have printouts done via a macro, so that we can override these and print to a different stream if wanted. 
2) Loosing the ability to override the printouts from CSP (CSP_USE_EXTERNAL_DEBUG)

Otherwise it looks good

johandc:
> From Gomspace point of view, we have 2 main concerns.
> 
> 1. The pollution of stdout with printf. We prefer to have printouts done via a macro, so that we can override these and print to a different stream if wanted.
> 2. Loosing the ability to override the printouts from CSP (CSP_USE_EXTERNAL_DEBUG)

@bfol 
Okay to me it sounds a little like the same thing. But I get your point. This is why there is no printf or stdout of any kind left inside CSP.

Regarding those few cases that remain, the old macro system was a little cumbersome and convoluted. How about we make a macro like this:

```c
#ifndef CSP_PRINTF
#define CSP_PRINTF printf
#endif
```

And the the build script could replace `CSP_PRINTF` with any function from your log system as long as it keeps the format, vaargs arguments.

johandc:
Actually it would be even cooler if we could get error codes for the last interfaces too, so there would be no need for stdout at all.

Regarding the very useful packet printout. It think we should make that as a user hook on incoming and outgoing. In fact that could even be used for stuff other than just the printout.

And the RDP debugging is something that is only rarely used, but I want to leave it in place with a toggle, and perhaps a compilation option to disable it, to save the large amount of strings inside.

johandc:
Another thing is that I might want to do is add weak linked implementations of many of the hooks. How do you like that idea?

bfol:
Converting the remaining printf to error codes and weak linked hooks would be the best option in my view. Especially a hook on packet input / output would be really usefull

johandc:
Okay, great I will do that.

yashi:
### Global error number vs return values

I don't see the point of having a global variable, `csp_dbg_errno`, instead of just a simple return value.  It's racy under multi-threads env we are all in.

Despite the name, it's not even a debugging facility.  It's actually the obligation of application authors to check how functions behaved.

This `csp_dbg_errno` thing is a big NAK from me.

This is related #234

### Subsystem error counters

I like the idea.  I'm not sure we should have fixed `csp_dbg_` prefix though.  But the name is rather minor issue.

If it's not atomic, we should use mutex to protect it. Yes, with ARM instruction, integer increment is atomic if and only if one thread is writing.  But, it should be good practice and relates to the next point.

We should allow variable sizes for the counter.  Some mission might want to use 64 bits for the counter.  If 64 bits are chosen, we need a lock or the counter must be atomic.

I don't know it's a good idea to make the variables global.  If the value is read-only for applications, it might be better to have a read accessor for them.

### Debugging printf

As have already mentioned, we, Space Cubics, wanted to replace them with our own custom functions.  But with subsystem error counters installed, and good error indication from function return values, we might not need it at all.  I haven't checked it fully just yet.

stdio is huge and we don't want it in our spacecraft.  But I'm sure it's nice to have while debugging.

### Other comments

I've started review this PR but honestly it's quite hard to see what's going on.

As noted by you, the PR consists 3 different issues with 9 commits, without telling which one does which.  I'll just list them here for me and any one visiting this PR in future.  No other intention is associated with it.

- **Remove assert from csp**
  - This should be the one for #221 
  
- **merge master**
  - It says "merge master" but it's not every merge commit.
  - This has the new debug APIs in it.  New symbols, `csp_dbg_init_errno`, `csp_dbg_inval_reply` and many other `csp_dbg_*`; new defines `CSP_DBG_*` in `csp_debug.h`
  - This commit replaces `csp_log_*()` with `printf()` by introducing (temporarily?) `stdio.h`.
  - So this is definitely for #221, but two big parts listed above. (nice to split)
  - But I might have missed something from "merge" part.
    
- **Cleanup in debug header**
  - This moves newly introduced symbols in `csp_debug.c`, and
  - Remove unused csp_debug related macros.
  - Rename a few newly introduced symbols.
  - This should be part of the former part of the previous commit (squashed to one after the split)
    
- **Fix examples and add color to packet and rdp print**
  - This actually has four parts:

  1. fixing code under `examples/` by removing old symbols
  2. introducing terminal colors in macros in `csp_debug.h`
  3. `csp_print_packet()` fix for newline handling
  4. removing redundant debug part from the build systems

- **do not use csp_conn objects for listener sockets**
  - `csp_port.c` indeed introduce two new callbacks, `csp_port_get_callback()` and `csp_bind_callback()`.
  - And, `csp_listen()` now call `csp_queue_create_static()`
  - So this is #242
  - But, many other parts are debug macro renames, which hide the important changes in `csp_port.c`
  - somehow csp_types.h now includes `csp_queue.h` and `csp_semaphore.h`, along with `struct csp_socket_s` from `csp.h`.
    
- **Fix build on freertos and Fix example and cmake build**
  - fix regression caused by other commits
  
- **Global RDP retransmission queue**
  - #241
  
- **Updated changelog**
  - just as it says.

johandc:
First of all, thank a lot for a very good review. I have been looking forwards to this one.

First of all, once again I can only apologise for the messy commits. The thing is that I couldn't implement these things without eachother. And the way GitHub handles merges, it automatically takes every into one pull request.

Also the merge commit was a mistake, it should have been a rebase, but after fixing a conflict it turned into a merge. 

Your dissertation of the individual changes is perfect. So no comments on that. I will address each of the major points in separate replies below

johandc:
> Global error number vs return values

Okay, i know that it's racy, but it may be the best solution and here is why.

1. It does not replace any return values. And the user is never required to check the error number for anything. Period.
2. It is only used for very rare error events that previously had a printf. For example if a null pointer is passed to csp_close or stuff like that. Stuff that shouldn't ever happen.
3. All the print outs that actually can happen, under normal operation like buffer underrun, and connection flooded are now in counters. A counter is much better than a printf.
4. Removing the printfs saves code size
5. Including the very rare stuff as special return values, just leaves the problem to the user to handle. And this adds to the code size and leads to more confusing API.
6. For many functions all that matters is wether it was successful or not. And like with the new csp_send() it is fire and forget. So CSP handles any errors internally.
7. The value of the CSP errno should always be zero. All the errors going in here is a sign of a malfunctioning system. So controlling that it's zero is only required if any network issues appear.

I think as a solution to the global errno, I will add a hook together with a weak implementation, so the user can catch the error codes themselves and do what they want with it. So CSP will default to a global.

johandc:
> Subsystem error counters

Each interface already has error counters, so I just expanded it to cover other parts of the stack.

Again it's not something that requires to be checked by any means. But if there is an issue, and you are loosing packets. Reading the debugging counters can point to the cause.

Suggestions for better names?

I want the memory address to be accessible from the outside. In order to map the variables to our parameter system. For remote read access.

Also I don't want any runtime setup, the symbols should be available to the linker. Because the mapping happens compile time.

Adding acessor functions just adds more code to do runtime what can be done compile time. 

The csp_conf struct is also a global, on order to reduce namespace clutter i could put the counters inside a struct.

I think this is the simplest solution to just have them being accessible as a global. And they are prefixed with csp_ so it's not like it should conflict with other variables.

Regarding atomic access, we should let the compiler handle that and use <stdatomic.h>. We don't want to introduce locks inside of a debugging function that can happen inside an isr. So maybe we just use 'atomic_uint_8'

The compiler actually reduces this to a single atomic instruction on most platforms. Which is pretty nice.

I chose uint8 for the space savings, and because it doesn't matter if an error counter wraps around. It's not the absolute number that's interesting. It's the fact that there was a change between two events for example. So usually while debugging you read out the value first, then perform an action and read out again. So no need for uint64.

johandc:
> Debugging printf

We have only two places where CSP would make calls to printfs left. And that is the packet printout, and the protocol debugging, which i still actively use when working with RDP.

I think I just agreed with bfol, to make the packet in/out event into a hook, with a weak implementation defaulting to printf when that flag is set.

And the protocol debug i think will be changed into a compile time enable, because it's actually only meant for the developer. 

We do have some printfs remaining on Linux only stuff like zmq and socketcan. But let's try to remove those too.

Finally stdio.h isn't that bad, if you use picolibc. It's tiny, low stack usage and does not use malloc().

johandc:
Final comment. Globals aren't always bad. Especially on Embedded. Because it can be a tedious task to pass down and cary around a CSP instance pointer through each and every function call. Especially because CSP is not designed to be multiple instances. It will always be the same pointer anyways.

And making the variables static means we would need a function call to get a pointer, which produces larger code and runs way slower. And if that function always returns the same pointer it might as well be available as a symbol.

yashi:
> Globals aren't always bad.

No, of course not.  Passing a parameter to a deep function chain pollute CPU registers and slow things down.  So if you have to run a tight loop with a value under a deep calls, global variables can save you.  But, we don't have a deep call chain in CSP.  We don't want a deep call chain because of limited stack size.

> Adding acessor functions just adds more code to do runtime what can be done compile time.

That's true! If the accessor code for each rarely used variables adds up, that's bad for restricted devicees.  And this is a good and a valid reason to use global variables.  So I agree on using global variables.  ​A bit more on below.

But,

> Because it can be a tedious task to pass down and cary around a CSP instance pointer through each and every function call.

I don't think this is a good reason.  tedious to write?  embedded software _is_ tedious to write because all nominal pathes are about 20 % of the total system.  80 % is error checking and abdominal code pathes we don't want to happen in the field.  OK you didn't mean this. :stuck_out_tongue_winking_eye:  But, we write reusable code once because we don't want to repeat tedious work again.  But someone must write that code for this messy unpredictable real-world so that all others can lazily use the code.  (Being a lazy is a good thing for a programmer, right :wink:)

> CSP is not designed to be multiple instances

If so, I can understand and I think it's a valid reason to have one single global csp instance and all function referencing to it.  That will reduce register pollution as well.

> making the variables static means we would need a function call to get a pointer

This is a good point to use global variables in embedded system until LTO is widely used!  Both GCC and Clang already supports `-flto`.  But Zephyr doesn't not support it yet.  How about FreeRTOS?

> i could put the counters inside a struct.

A good pre defined prefix should be enough, IMHO.  Or is there any merit of using struct that I don't see?

johandc:
> But someone must write that code for this messy unpredictable real-world so that all others can lazily use the code. (Being a lazy is a good thing for a programmer, right 😉)

I cannot agree more on this. Best example is csp_send. CSP should be so easy to use, that it is not even required to check the return value of the send function, in order to avoid a buffer leak.

(If we want the old behaviour back, it will be a different function, which allows for retrying with the same buffer.)

The debug variables are really meant as lazy mans measure, to get the cause, if something is not looking right. The normal use cases, should not be polluted with extremely rare return values, that the user won't check for anyways.

In general it was a mistake to carry along FreeRTOS convention of returning pdTRUE which is a define set to 1. I believe it's nicer to use the zero good everything else bad approach. But that is another issue #209

src/csp_debug.c
```diff
@@ -8,4 +10,22 @@ uint8_t csp_dbg_conn_noroute;
...
+    va_start(args, fmt);
+    vfprintf(stdout, fmt, args);
+    va_end(args);
+}
+
+#else
+
+__attribute__((weak)) void csp_print(const char * fmt, ...)) {
+    return;
+}
```
yashi: Make this an inline function in csp_debug.h so that we don't even have to branch to this empty function if want to disable calling `csp_print()` when `!CSP_DEBUG`.  Compilers will optimized this out.

Also, you don't need weak here.

yashi (changes requested):

src/csp_debug.c
johandc: It cannot be a preprocessor macro, if the function should be weak linked and overridden at link time. By a user supplied implementation.

I originally suggested at simple define, but I'm not sure it was enough to satisfy the need of gomspace. As the old system allowed to override.

yashi: I said _inline function_, not a preprocessor macro.

Do you want user to override `csp_print()` when `!CSP_DEBUG`? And do you want to supply non-debug version of `csp_print()`, which is just empty function?  If that's your intention, can we do:

```c
#if (CSP_DEBUG)
#include <stdio.h>
#endif

__attribute__((weak)) void csp_print(const char * fmt, ...) {
#if (CSP_DEBUG)
    va_list args;
    va_start(args, fmt);
    vfprintf(stdout, fmt, args);
    va_end(args);
#else
    return;
#endif
}
```
This tells to reader that our intentions are
- We always provide weak version of `csp_print()`.
- When `CSP_DEBUG`, we provides `vfrpintf()` as the default, which to be overridden.
- When `!CSP_DEBUG`. we provide an empty function as the default, which to be overridden.

If you don't want user to override `csp_print()` when `!CSP_DEBUG`, you can do the following in the _header file_.  This i what I suggested in the previous comment.

**in the .c source file**
```c
#if (CSP_DEBUG)
#include <stdio.h>

__attribute__((weak)) void csp_print(const char * fmt, ...) {
    va_list args;
    va_start(args, fmt);
    vfprintf(stdout, fmt, args);
    va_end(args);
}
#endif
```
**in the header file**
```c
#if (CSP_DEBUG)
void csp_print(const char * fmt, ...);
#else
inline void csp_print(const char * fmt, ...) {}
#endif
```

This can tell compiler to optimize out the branch instruction to the empty funciton, which is not possible without LTO because the function body is in the other compile uint.

johandc: Oh, wish i saw this before. 

I ended up with something very similar, using the preprocessor to remove the printout alltogether.

johandc: I ended up renaming CSP_DEBUG to CSP_HAVE_STDIO (inline with some of the other header HAVE_LIB_SOCKETCAN and HAVE_LIB_ZMQ)

This means we can rather easily remove stuff that depends on stdio, such as snprintf, and scanf.

I realize that there could be cases where the user does NOT want to include STDIO in CSP, but the DO want to override the function. But i'm also in doubt if adding an extra option such as CSP_ENABLE_PRINT is really necessary.

johandc:
Final call,

I took care of a bunch of stuff and wiped the last few issues i had remaining. It's a little all or nothing with this commit. If there's something you don't like we can still change it on develop after the merge.

include/csp/csp_hooks.h
```diff
@@ -0,0 +1,17 @@
+#include <csp/csp_types.h>
```
bfol, line 1: Missing #pragma once

src/csp_service_handler.c
```diff
@@ -232,61 +226,14 @@ void csp_service_handler(csp_packet_t * packet) {
 
 		case CSP_PING:
 			/* A ping means, just echo the packet, so no changes */
-			csp_log_info("SERVICE: Ping received");
 			break;
 
-		case CSP_PS: {
```
bfol, line 238: We still use this functionality. If its remove due to removed arch support, it would be find to add a HOOK that can be implemented externally

johandc: I think it should be moved outside CSP. Because it has nothing to do with the network stack. I also wanted to move memfree, but that was to tiny I left it.

What is your opinion, should task list be part of CSP?

include/csp/csp_hooks.h
johandc: Thanks

src/csp_service_handler.c
johandc: @bfol 

We just moved the tasklisk to a parameter called os_print_dbg, when that is set to 1, FreeRTOS prints it's debug information including the task list in ASCII to our stdout where we collect it from.

I know you have an external log system, so it should be easy for GS to make such a parameter.

POSIX systems has other ways of debugging tasks, so I don't think it really belongs in CSP

bfol: removing CPS_PS completely is not backwards compatible. This is my main concern. As I wrote in my original comment, I am perfectly fine with changing the implementation to a hook instead, so the user can choose to implement the functionality the way he want's

johandc: I have created a new thread on this and other things i would like to deprecate.

PR #305, opened 2021-12-01
Remove missing CSP_NODE_MAC in pycsp.c

avinashraja98:
Seems like CSP_NODE_MAC was removed in csp_rtable.h but not in the python bindings file in commit e39c97cfa6adf77249f0494414cc3bb5e803e701. Was causing build errors.


johandc:
Thanks

PR #315, opened 2021-12-08
Re-introduce csp_conn_get_array

bfol:
We use this one in a lot of our unit tests


johandc:
Are those unit tests related to libcsp? If so, could they be put into github actions somehow?

bfol:
The unit tests where we use these are to check that our server processes closes all their connections - so not really useful in libcsp

PR #316, opened 2021-12-14
Update i2c interface to new code.

bfol:
The increased address space of CSP 2.0 is handled by just the lower 7 bits of
the destination address as the address of the i2c interface. The user must
handle this in the address and routing setup of the satellite.


johandc:
Looks good

PR #320, opened 2021-12-21
Make Zephyr port buildable again

yashi:
This is the first batch to fix and adopt new libcsp API in Zephyr port.  Each commit should be trivial and easy to follow.  Nothing technical.

With all the changes done to libcsp, we can now build examples/csp_arch.c as-is against Zephyr.  Yay!


johandc:
Well done sir

PR #321, opened 2021-12-21
meson.build: Remove redundant csp_shlib and csp_shdep

yashi:
Meson allows you to specify either static or shared library at build
time with `default_library`.

From a parent project, you can either:

    meson setup builddir

or

    meson setup builddir -Ddefault_library=static

If you have newer Meson than libcsp requires, you can also specify it
in `dependency()`.

Thus, we don't need two distinct deps for each library.

ref)
https://mesonbuild.com/Release-notes-for-0-54-0.html#per-subproject-default_library-and-werror-options
https://mesonbuild.com/Release-notes-for-0-60-0.html#dependency-sets-default_library-on-fallback-subproject

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -101,7 +101,7 @@ endif
 
 csp_deps += clib
 
-csp_lib = static_library('csp',
+csp_lib = library('csp',
 	sources: [csp_sources, csp_config_h],
 	include_directories : csp_inc,
 	dependencies : csp_deps,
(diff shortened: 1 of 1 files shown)


eli-schwartz:
https://github.com/libcsp/libcsp/blob/7bb0527ab068508826c3ba2fb7ddbc0e29f9d859/examples/meson.build#L8-L22

You probably also want to remove the second of these. It's both useless and references an undefined variable.

And then you can remove the unused option from meson_options.txt :)

yashi:
@eli-schwartz, thanks.  #323

eli-schwartz:
You still didn't remove the unused option :) https://github.com/libcsp/libcsp/blob/1f20246765990f92e2caa4aa000729310350b58e/meson_options.txt#L8

yashi:
oops... updated.  Thanks.  Hope everything is there now.

PR #323, opened 2021-12-21
meson.build: Remove csp_arch_shared

yashi:
This is a leftover from the previous change.  We don't need the
dedicated csp_arch_shared.  We can `-Ddefault_library=shared`, which
is by default right now.

If you need static version of it, do:

    meson setup builddir -Ddefault_library=static && meson compile -C builddir examples/csp_arch


yashi:
Thanks @eli-schwartz for the reviews in #321.

Issue #318, opened 2021-12-18
sem: Why do we need alignment hack for all arches


yashi:
https://github.com/libcsp/libcsp/blob/de88960c9502c80931cef5272d13b49921586a8e/include/csp/arch/csp_semaphore.h#L48-L57

@johandc, would you mind elaborating me why we needed this alignment thing for _all_ arch?

johandc:
Yes. I thought it would be nice for all semaphores to be statically allocated. This was previously done with a macro that resolved to some arch native type. So a csp_sem_t was not an actual type, it was an OS thing that put a requirement on an external header. Like FreeRTOS.h or zephyr.h for anybody who uses the semaphores. So i guess it helps reduce the number of headers that gets pulled in.

I don't know which is better. To use the union to perform the alignment or the GCC attribute align.

yashi:
> I thought it would be nice for all semaphores to be statically allocated.

I agree.

> So i guess it helps reduce the number of headers that gets pulled in.

Why did you want to reduce the number of header files to be included?

Header files are there to be used.  Let's use it when we need it.

I can keep this tricked typdef for FreeRTOS like the following or convert to simple `typedef StaticSemaphore_t csp_bin_sem_t;`.  What do you think?

```diff
diff --git a/include/csp/arch/csp_semaphore.h b/include/csp/arch/csp_semaphore.h
index 4b47616c..e605ecd3 100644
--- a/include/csp/arch/csp_semaphore.h
+++ b/include/csp/arch/csp_semaphore.h
@@ -22,9 +22,8 @@
  */
 
 #if (CSP_POSIX || __DOXYGEN__)
-
-    /* POSIX interface */
-    #define CSP_SIZEOF_SEM_T 32
+#include <semaphore.h>
+typedef sem_t csp_bin_sem_t;
 
 #elif (CSP_FREERTOS)
 
@@ -35,16 +34,6 @@
 #define CSP_SIZEOF_SEM_T sizeof(StaticSemaphore_t)
 #else
 #define CSP_SIZEOF_SEM_T sizeof(void *)
-#endif
-
-#elif (CSP_ZEPHYR)
-
-    /* Zephyr RTOS Interface */
-    #include <zephyr.h>
-    #define CSP_SIZEOF_SEM_T sizeof(struct k_sem);
-
-#endif
-
 /**
  * This definition is borrowed from POSIX sem_t
  * It ensures the proper amount of memory to hold a static semaphore
[...]

PR #326, opened 2021-12-22
csp_semaphore: Make csp_semaphore internal

yashi:
We need an architecture independent semaphore implementation to be used in libcs.  But we don't want to expose it to be a part of our API.  Applications, OTOH, can directly use its target OS services.

Our goal is to provide a small, efficient, and easy to use network stack for CubeSat.  Not providing yet another OS abstraction layer.

In fact, there are plenty of OS abstraction libraries out there if applications need it.


yashi:
#318

PR #330, opened 2021-12-26
Link layer redundancy

johandc:
multiple can interfaces. The old PBUF implementation
used global state, meaning there was a shared input queue
if two CAN interfaces is created, and the same message is seen
on both interfaces. The pbuf system would not work.

The previous rewrite of the PBUF system went from a static pool
to a linked list. This commit goes the last bit of the way by
moving the head of the linked list from a static variable into
each interface data structure.

This means fully independant PBUFs for CAN -> CSP packet assembly
allowing for redundant CAN interfaces with the same data.

Future CSP router will support "copy_to", so CAN0 can be setup to
copy all packets to CAN1 for redundancy. Together with proper
deduplication this allows for full link_layer redundancy over CAN bus.

Signed-off-by: Johan de Claville Christiansen <johan@space-inventor.com>

### DIFF
--- a/src/csp_dedup.c
+++ b/src/csp_dedup.c
@@ -3,9 +3,11 @@
 #include "csp_dedup.h"
 
 #include <stdlib.h>
+#include <stdio.h>
 
 #include <csp/arch/csp_time.h>
 #include <csp/csp_crc32.h>
+#include <csp/csp_id.h>
 
 /* Check the last CSP_DEDUP_COUNT packets for duplicates */
 #define CSP_DEDUP_COUNT 16
@@ -20,7 +22,8 @@ static int csp_dedup_in = 0;
 
 bool csp_dedup_is_duplicate(csp_packet_t * packet) {
 	/* Calculate CRC32 for packet */
-	uint32_t crc = csp_crc32_memory((const uint8_t *)&packet->id, packet->length + sizeof(packet->id));
+	csp_id_prepend(packet);
+	uint32_t crc = csp_crc32_memory(packet->frame_begin, packet->frame_length);
 
 	/* Check if we have received this packet before */
 	for (int i = 0; i < CSP_DEDUP_COUNT; i++) {
@@ -29,8 +32,9 @@ bool csp_dedup_is_duplicate(csp_packet_t * packet) {
 		if (crc == csp_dedup_array[i]) {
 
 			/* Check the timestamp */
-			if (csp_get_ms() < csp_dedup_timestamp[i] + CSP_DEDUP_WINDOW_MS)
+			if (csp_get_ms() < csp_dedup_timestamp[i] + CSP_DEDUP_WINDOW_MS) {
 				return true;
+			}
 		}
 	}
 
--- a/src/csp_dedup.c
+++ b/src/csp_dedup.c
@@ -3,7 +3,6 @@
 #include "csp_dedup.h"
 
 #include <stdlib.h>
-#include <stdio.h>
 
 #include <csp/arch/csp_time.h>
 #include <csp/csp_crc32.h>
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -238,14 +238,6 @@ int csp_send_direct_iface(csp_id_t idout, csp_packet_t * packet, csp_iface_t * i
 	iface->tx++;
 	iface->txbytes += bytes;
 
-
-	/* COPY TO */
-	if (iface->copy_to != NULL) {
-
-		mtu = iface->copy_to->mtu;
-		if (mtu > 0 && bytes > mtu)
-			goto tx_err;
-	}
 	return CSP_ERR_NONE;
 
 tx_err:
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -133,6 +133,9 @@ int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 		/* Apply outgoing interface address to packet */
 		idout.src = iface->addr;
 		
+		/* Todo: Find an elegant way to avoid making a copy when only a single destination interface
+		 * is found. But without looping the list twice. And without using stack memory.
+		 * Is this even possible? */
 		copy = csp_buffer_clone(packet);
 		if (csp_send_direct_iface(idout, copy, iface, via, from_me) != CSP_ERR_NONE) {
 			csp_buffer_free(copy);
@@ -156,7 +159,7 @@ int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 		return CSP_ERR_NONE;
 	}
 
-	if (csp_send_direct_iface(idout, copy, route->iface, route->via, from_me) != CSP_ERR_NONE) {
+	if (csp_send_direct_iface(idout, packet, route->iface, route->via, from_me) != CSP_ERR_NONE) {
 		csp_buffer_free(packet);
 	}
 
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -134,7 +134,7 @@ int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 		idout.src = iface->addr;
 		
 		copy = csp_buffer_clone(packet);
-		if (csp_send_direct_iface(idout, copy, iface, via, 1) != CSP_ERR_NONE) {
+		if (csp_send_direct_iface(idout, copy, iface, via, from_me) != CSP_ERR_NONE) {
 			csp_buffer_free(copy);
 		}
 
@@ -156,7 +156,7 @@ int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 		return CSP_ERR_NONE;
 	}
 
-	if (csp_send_direct_iface(idout, copy, route->iface, route->via, 1) != CSP_ERR_NONE) {
+	if (csp_send_direct_iface(idout, copy, route->iface, route->via, from_me) != CSP_ERR_NONE) {
 		csp_buffer_free(packet);
 	}
 
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -125,21 +125,37 @@ int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 	int ret;
 
 	/* Try to find the destination on any local subnets */
-	csp_iface_t * local_interface = csp_iflist_get_by_subnet(idout.dst);
-	if (local_interface) {
-		idout.src = local_interface->addr;
-		ret = csp_send_direct_iface(idout, packet, local_interface, CSP_NO_VIA_ADDRESS, 1);
+	int via = CSP_NO_VIA_ADDRESS;
+	csp_iface_t * iface = csp_iflist_get_by_subnet(idout.dst);
 
 	/* Otherwise, resort to the routing table for help */		
-	} else {
+	if (iface == NULL) {
 		csp_route_t * route = csp_rtable_find_route(idout.dst);
 		if (route == NULL) {
 			csp_dbg_conn_noroute++;
 			return CSP_ERR_TX;
 		}
-		idout.src = route->iface->addr;
-		ret = csp_send_direct_iface(idout, packet, route->iface, route->via, 1);
+		iface = route->iface;
+		via = route->via;
 	}
+
+	/* Apply outgoing interface address to packet */
+	idout.src = iface->addr;
+
+	/* Perform copy of packet, before sending */
+	csp_packet_t * copy = NULL;
+	if (iface->copy_to) {
+		copy = csp_buffer_clone(packet);
+	}
+
+	/* Send */
+	ret = csp_send_direct_iface(idout, packet, iface, via, 1);
+	
+	/* Send copy */
+	if (iface->copy_to) {
+		csp_send_direct_iface(idout, copy, iface->copy_to, via, 1);
+	}
+
 	return ret;
 
 }
@@ -221,6 +237,15 @@ int csp_send_direct_iface(csp_id_t idout, csp_packet_t * packet, csp_iface_t * i
 
 	iface->tx++;
 	iface->txbytes += bytes;
+
+
+	/* COPY TO */
+	if (iface->copy_to != NULL) {
+
+		mtu = iface->copy_to->mtu;
+		if (mtu > 0 && bytes > mtu)
+			goto tx_err;
+	}
 	return CSP_ERR_NONE;
 
 tx_err:
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
 
 int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 
-	int ret;
-
 	/* Try to find the destination on any local subnets */
 	int via = CSP_NO_VIA_ADDRESS;
-	csp_iface_t * iface = csp_iflist_get_by_subnet(idout.dst);
-
-	/* Otherwise, resort to the routing table for help */		
-	if (iface == NULL) {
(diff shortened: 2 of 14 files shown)


johandc:
Went ahead and added the fix for #331 this commit too.

I want to propose another solution, where the link layer duplication happens by sending to all interfaces that at link local to the destination. In that regard if both interfaces are setup with similar id and netmask, both of them will be used for transmission. This requires the function csp_iflist_get_by_subnet to be able to return multiple results. In C this is best handled by a callback pattern or an iterator pattern. I'm a bit undecided which one to go with for now. But this 'copy_of' is not likely to make it into develop branch.

johandc:
Okay, i have working link-layer redundancy here now.

It works by looping over all available matching interfaces and sending out the packet to each one of them. Here is an example yaml configuration for csh to test it:

```yaml
# a.yaml
- name: "ZMQ"
  driver: "zmq"
  server: "localhost"
  addr: 1
  netmask: 8
  default: true

- name: "CAN"
  driver: "can"
  device: "vcan0"
  addr: 1
  netmask: 8

```

```yaml
# b.yaml
- name: "ZMQ"
  driver: "zmq"
  server: "localhost"
  addr: 1
  netmask: 8
  default: true

- name: "CAN"
  driver: "can"
  device: "vcan0"
  addr: 1
  netmask: 8

```

```bash
./builddir/csh -f a.yaml 
```

```bash
./builddir/csh -f b.yaml 

  ***********************
  **     CSP   Shell   **
  ***********************

  Copyright (c) 2016-2022 Space Inventor ApS <info@space-inventor.com>

  Reading config from b.yaml
  ZMQ addr: 2 netmask 8
  Setting default route to ZMQ
  CAN addr: 2 netmask 8

% 
% ident 63

IDENT 1
  kontor
  #22-Ubuntu SMP Fri Nov 5 13:2
  5.13.0-22-generic
  Dec 27 2021 00:52:07

IDENT 2
  kontor
  #22-Ubuntu SMP Fri Nov 5 13:2
  5.13.0-22-generic
  Dec 27 2021 00:59:13


% info
ZMQ        addr: 2 netmask: 8 mtu: 1024
           tx: 00002 rx: 00003 txe: 00000 rxe: 00000
           drop: 00002 autherr: 00000 frame: 00000
           txb: 194 (194.0B) rxb: 291 (291.0B) 

CAN        addr: 2 netmask: 8 mtu: 2040
           tx: 00002 rx: 00002 txe: 00000 rxe: 00000
           drop: 00000 autherr: 00000 frame: 00001
           txb: 194 (194.0B) rxb: 194 (194.0B) 

```

I have confirmed this actually works by running a continous ping between the nodes, and sucessively opening and closing either the zmq or the vcan interface. Also note in the above example that the "drop" counter = 2 on the ZMQ interface.

src/csp_dedup.c
```diff
@@ -3,9 +3,11 @@
 #include "csp_dedup.h"
 
 #include <stdlib.h>
+#include <stdio.h>
```
yashi: Why did you need `stdio.h`?

src/csp_io.c
```diff
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
...
+		return CSP_ERR_NONE;
 	}
 
-	return ret;
+	/* Try to send via routing table */
+	csp_route_t * route = csp_rtable_find_route(idout.dst);
+	if (route == NULL) {
+		csp_dbg_conn_noroute++;
+		csp_buffer_free(packet);
+		return CSP_ERR_NONE;
```
yashi, line 159: If we don't find a local interface nor a route for the given packet, we are dropping the packet, right?  Is it still OK to return `CSP_ERR_NONE`?

src/csp_io.c
```diff
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
...
 int csp_send_direct(csp_id_t idout, csp_packet_t * packet, int from_me) {
 
-	int ret;
-
 	/* Try to find the destination on any local subnets */
 	int via = CSP_NO_VIA_ADDRESS;
-	csp_iface_t * iface = csp_iflist_get_by_subnet(idout.dst);
+	csp_iface_t * iface = NULL;
+	csp_packet_t * copy = NULL;
+	int local_found = 0;
```
yashi, line 129: Should be `bool`.

src/csp_io.c
```diff
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
...
-	return ret;
+	/* Try to send via routing table */
+	csp_route_t * route = csp_rtable_find_route(idout.dst);
+	if (route == NULL) {
+		csp_dbg_conn_noroute++;
+		csp_buffer_free(packet);
+		return CSP_ERR_NONE;
+	}
+
+	if (csp_send_direct_iface(idout, copy, route->iface, route->via, 1) != CSP_ERR_NONE) {
```
yashi: Should be `packet` instead of `copy`?

src/csp_io.c
```diff
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
...
-			csp_dbg_conn_noroute++;
-			return CSP_ERR_TX;
+	while ((iface = csp_iflist_get_by_subnet(idout.dst, iface)) != NULL) {
+		
+		/* Apply outgoing interface address to packet */
+		idout.src = iface->addr;
+		
+		copy = csp_buffer_clone(packet);
+		if (csp_send_direct_iface(idout, copy, iface, via, 1) != CSP_ERR_NONE) {
+			csp_buffer_free(copy);
```
yashi: When you failed to send, don't you wanna tell caller what happened?

src/csp_io.c
```diff
@@ -122,42 +122,46 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
...
+		return CSP_ERR_NONE;
+	}
+
+	if (csp_send_direct_iface(idout, copy, route->iface, route->via, 1) != CSP_ERR_NONE) {
+		csp_buffer_free(packet);
+	}
 
+	return CSP_ERR_NONE;
+	
 }
```
yashi: Just some ideas:

**single return with indentations**
```c
{
	while ();
	
	ret = CSP_ERR_NONE;
	
	/* If the above worked, we don't want to look at the routing table */
	if (!local_found) {
		/* Try to send via routing table */
		csp_route_t * route = csp_rtable_find_route(idout.dst);
		if (route == NULL) {
			csp_dbg_conn_noroute++;
			ret = CSP_ERR_TX;
		} else {
			ret = csp_send_direct_iface(idout, packet, route->iface, route->via, 1);
			if (ret == CSP_ERR_NONE) {
				goto end;
			}
		}
	}
	csp_buffer_free(packet);
	
end:	
	return ret;
}
```


**single return with goto**
```c
{
	while ();
	
	/* If the above worked, we don't want to look at the routing table */
	if (local_found) {
		ret = CSP_ERR_NONE;
		goto free;
	}
	
	/* Try to send via routing table */ 
	csp_route_t * route = csp_rtable_find_route(idout.dst);
	if (route == NULL) {
		csp_dbg_conn_noroute++;
		ret = CSP_ERR_TX;
		goto free;
	}

	ret = csp_send_direct_iface(idout, packet, route->iface, route->via, 1);
	if (ret == CSP_ERR_NONE) {
		goto end;
	}
	
free:
	csp_buffer_free(packet);

end:	
	return ret;
}
```

src/csp_dedup.c
johandc: I don't. It was on for a brief period.to print some debugging. Should be removed

src/csp_io.c
johandc: Yes. I have come to terms with the fact that csp_send should not return a result. This is further exacerbated by the fact that one call to send can now send out multiple packets over multiple interfaces, so which call to csp_send_direct_iface should be the one setting the return value.

We should refactor most of the outgoing chain to use void return value. Haven't gotten to that part of the cleanup yet.

How to make blocking writes then? Well use RDP that waits on the acknowledgement. Don't use interface level flow control.

src/csp_io.c
johandc: I don't like bool. It's messy to include stdbool everywhere. And does not have any benefits to int. Are there any benefits that I don't know of? Haven't really investigated tbh

src/csp_io.c
johandc: Yes. Definitely.

Originally I removed the routing table from the send call just to test this function. The routing table lookup isn't really compatible with multiple outgoing interfaces yet. It was using the copy_to, which would have worked both for local interfaces and for routing table lookups. That however was an extra piece of configuration I didn't like.

I'm thinking that maybe if there are two routing entries with the same netmask. For example 0/0 and another 0/0 it would use both.

I'm in doubt on where to go with routing tables. But that's maybe a topic for a separate issue.

johandc: The doubt is based on that the link configuration now sorts our 95% of all routing out of the box. I can really only see this being relevant in something like a radio or edge router.

src/csp_io.c
johandc: Nope. See above.

The sender reacting to errors is opening a large can of trouble to the sender. And it won't actually ensure any better performance. Because the packet could be silently dropped anywhere in the network anyway.

src/csp_io.c
johandc: I think the correct solution is to have csp_send_direct_iface being a void and let the free happen at the lower layer.

johandc:
Thanks for your, as always, very helpful review.

src/csp_io.c
yashi: Well, it tells you that the variable is a bool by just looking at it, not something you count or calculate.  Lowering our cognitive load is always a good thing.

There is a saying that "The code will be read many more time than writing."  So, we should always take a bit more time while writing for future readers (including ourselves). :smiley:   It's true for the source code as well as our commit log.

Some MISRA guys use `bool` instead of `int` because `if (a)` is a violation if `int a`, but not if `bool a`.  But, I don't really care this.

src/csp_io.c
yashi: Or we can have another function, `csp_send_direct_iface_free()`, which also consumes the given packet when failed.

BTW, are you OK with `goto`?  Multiple `return`s? Or deep nested blocks?  Any preferable style you can share with us?

BTW2, I still don't like when send functions don't tell anything.  Applications can't make sure the packet is safely sent.  I think it's crucial for applications to know what happened to the packet they sent.  When the packet is a telemetry to a GS, that's a life line for a satellite I suppose.  I know your reasoning of having internal error counters but...  Might be nice if you can share your experience on this (when you have time).

Well, this is a separate issue, I guess.

src/csp_io.c
johandc: Thanks for your input on this. I think we could make a separate issue to "bool'ify" the entire library. Identify all usages of `if (int a) ` in the code and address them separately. I wont fix for this PR. but i see the issue if MISRA guys says it. Do you know where to lookup the MISRA guideline for this, and perhaps you can make a new issue with a link to that where we can continue the talk.

src/csp_io.c
johandc:
```c
/* Todo: Find an elegant way to avoid making a copy when only a single destination interface
* is found. But without looping the list twice. And without using stack memory.
* Is this even possible? */
		 
```

src/csp_io.c
johandc: > BTW, are you OK with `goto`? Multiple `return`s? Or deep nested blocks? Any preferable style you can share with us?

`goto` : Yes. Very nice for exiting functions with nested cleanup required. In this respect i use it in the linux kernel manner.

Multiple `returns` : Yes. I like to exit functions early, if possible. This avoids the usage of `else` and deep nesting. One exception is if there is a simple return in the else statement. I prefer 

```c
int max(int a, int b)
{
    if (a > b)
        return a;
    else
        return b;
}
```

over the mozilla 

```c
int max(int a, int b)
{
    if (a > b)
        return a;
    return b;
}
```

but then of course theres this. That closes the argument entirely. 

```c
int max(int a, int b)
{
    return (a > b ? a : b);
}
```

But generally i consider `else` to be a warning indicator. Not directly harmfull as some would say. 
https://blog.mozilla.org/nnethercote/2009/08/31/no-else-after-return-considered-harmful/
:)

**Deep nested blocks : No, Just no. This is C not javascript.**

```js
app.get('/test', function (req, res) {
  http.get('some/place', function (req1, res1) {
    if (res1.statusCode == 200) {
      res1.on('data', function (data) {
        http.get('other/place?q=' + data, function (req2, res2) {
          if (res2.statusCode == 200) {
            res2.on('data', function (data) {
              db.query(data).on('data', function (rows) {
                res.writeHead(200)
                res.end(JSON.stringify(rows))
              })
            })
          }
        })
      })
    }
  })
})
```

yashi: > This is C not javascript.

:rofl:

johandc: > BTW2, I still don't like when send functions don't tell anything. Applications can't make sure the packet is safely sent. I think it's crucial for applications to know what happened to the packet they sent. When the packet is a telemetry to a GS, that's a life line for a satellite I suppose. I know your reasoning of having internal error counters but... Might be nice if you can share your experience on this (when you have time).

_Applications can't make sure the packet is safely sent._

Correct! And they also cannot make sure the packet is safely delivered. csp_send has been reporting a status for over a decade now. And i have never ever used it, not even once. The idea was that csp_send could report if the outgoing buffer was full, and cause the sender to wait and retransmit, this would serve as a flow control mechanism. However as i also commented above. Outgoing interface buffer level, is hardly a flow control mechanism at all. Because it's almost guaranteed to no be the actual bottleneck. Proper flow control is usually only required for stream or file transfers which uses RDP anyways. For everything else, the transaction style, request/response, usually is enough.

_When the packet is a telemetry to a GS, that's a life line for a satellite I suppose_

That is also absolutely correct. Some data are very critical. But consider your alternative? You already did "your best" by calling csp_send(), if that didn't work what's the alternative? If the user had an alternative, better or more safe way of transmitting that critical data they should be using both csp_send and whatever_send for a backup.

Also. There are only two reasons why the csp_send can fail:

Outgoing interface buffer full: The link is saturated. This should not happen in the first place, and the only way to clear the saturation is to start dropping packets. Which is the intended purpose. Stalling tasks and output can lead to more issues.
Outgoing interface down or misconfigured: In this case repeatedly calling csp_send(), would not yield any better result than the first call. Dropping the packet would allow the application to continue and not stall, when the interface goes down. Usually this is what we want to do.

Again, i have never ever had a usage for the csp_send return value. It has only been a liability for the user, and a cause of buffer leaks, for the user to be required to check and free.

src/csp_io.c
yashi: Rule 14.4 of MISRA-C 2012: The controlling expression of an if statement and the controlling expression of an iteration- statement shall have essentially Boolean type

yashi: Additional info: https://forum.misra.org.uk/thread-1344.html

> MISRA-C introduces a strong typing model to help users avoid unexpected issues which occur in the C typing model.
> The rule prevents the accidental use of integer expression when a boolean expression was intended.
> 
> To be compliant with MISRA rule 14.4 you should write your test as a boolean expression:
>` if((udtEvents & udtSTOP_EVENT) != 0U  )`

johandc:
@bfol  hi Brian, can I have your take on all this? How does gomspace configure and setup routes today and which impact does it have that CSP now becomes an interface first kind of routing. And for most systems the routing table is now unneeded. I know you are using I2C with the via address for the default route today. I was considering adding a default flag to one of the interfaces to completely remove the need for the routing table.

We used to configure All systems using the routing table string. But it is kind of heavy to include sscanf just for that purpose when simpler means can do the job better.

Is it time we had a meeting to align about this?

bfol:
@johandc I will look into this later this week, when i am done with the post vacation panic tasks

PR #333, opened 2021-12-27
Make csp_print configurable

### LINKED ISSUE #319: thoughts on csp_print()
`csp_print()` is a macro but `csp_print_func()` is a function.  `csp_print()` doesn't pass any file related information to `csp_print_func()`. 
 That means that we can't get any file related info, such as `__LINE__` or `__file__` in `csp_print_func()`.

There is no way to replace `csp_print()`.

Also, some OS can print without stdio.  ie) having simple UART driver without full `stdio.h`.  But currently we are either
- having `stdio.h` and `csp_print_func()`, which uses `stdio.h`, or
- `csp_print()` become noop.  no print.

We need third option.  Enable `csp_print()` without `stdio.h`.

To combine the two, I propose: Make csp_print() replaceable so that custom csp_print() can take context info if needed.

yashi:
This PR makes `csp_print()` toggleable by `CSP_ENABLE_CSP_PRINT` option.  The main commit is "csp_print: Add option to toggle csp_print".  The others are minor fixes and changes.

One thing I'd also like to fix, but it's not done yet, is to have yet another option for `csp_print_func()` with `stdio.h`.  Because I think there is a situation where one want all `sscanf()` and `snprintf()` related functions in libcsp but don't really want to use `vprintf()` version of `csp_print_func()`.

I'm fine having an empty `csp_print_func()` for all arch except for the POSIX arch.  Because for POSIX, we always have `stdio.h` anyway.  For other arches, I also think it's OK to have arch specific `csp_print_func()`. But I don't think @johandc likes the idea, because it will pull arch specific headers in our public header.

WDYT?

This should close #319.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ endif()
 
 set(CSP_HAVE_STDIO 1)
 set(CSP_ENABLE_CSP_PRINT 1)
+set(CSP_PRINT_STDIO 0)
 set(CSP_QFIFO_LEN 15 CACHE STRING "Length of incoming queue for router task.")
 set(CSP_PORT_MAX_BIND 16 CACHE STRING "Length of incoming queue for router task")
 set(CSP_CONN_RXQUEUE_LEN 16 CACHE STRING "Number of packets in connection queue")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 endif()
 
 set(CSP_HAVE_STDIO 1)
+set(CSP_ENABLE_CSP_PRINT 1)
 set(CSP_QFIFO_LEN 15 CACHE STRING "Length of incoming queue for router task.")
 set(CSP_PORT_MAX_BIND 16 CACHE STRING "Length of incoming queue for router task")
 set(CSP_CONN_RXQUEUE_LEN 16 CACHE STRING "Number of packets in connection queue")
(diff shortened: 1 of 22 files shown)


johandc:
> One thing I'd also like to fix, but it's not done yet, is to have yet another option for `csp_print_func()` with `stdio.h`. Because I think there is a situation where one want all `sscanf()` and `snprintf()` related functions in libcsp but don't really want to use `vprintf()` version of `csp_print_func()`.

I have reviewed the usages of snprintf / sscanf:

- [ ] csp_conn.c: Here it is used for a helper printout `csp_conn_print_table` i actually might just throw out that function completely. I don't see it as being a part of the CSP public API at all. It should only ever be used internally for debugging while developing the stack.
- [ ] csp_iflist.c: Here it is used in `csp_iflist_print` this is a bit more useable to the user to check the configuration of CSP is correct. But i am very tempted to just make an interator function the user can calll to loop over all interfaces and put the debugging output in the users hands. This removes the complexity of snprintf and keeps the library small and tight. We could add helper functions in contrib for users to pull from as a quick resource.
- [ ] csp_rtable_stdio.c: Passing configuration to CSP as a string, and using sscanf, snprintf was a bad solution. I think i prefer the new method of passing a pointer to an integer array instead. See #334 

Doing the above would yield a library completely without stdio.

yashi:
@johandc, what do you think if I add `CSP_PRINT_WITH_VPRINTF` option to control the current `csp_print_func()` implementation?  The option is default `yes`.  With `no`, we get an empty `csp_print_func()` so that users can implement however they want.

yashi:
Or, would `CSP_PRINT_WITH_STDIO` be better?

johandc:
Maybe shorten the name and skip the preposition. So maybe just CSP_PRINT_STDOUT or something?

johandc (approved):

PR #348, opened 2022-03-15
Fix master

sf314:
This is an incredibly old project, and managed to be 180 commits behind master. The new commits will be pulled in.

There was only one file affected by the update that ended up having merge conflicts: `wscript`. For each case, the conflict occurred because I had modified the script to add my own custom interfaces, so in the conflict UI I kept both changes.


sf314:
Edit: did this the wrong way round, meant to update a different project

johandc:
Master is old on purpose because CSP 2 isn't released officially yet.

sf314:
Yes sorry! By `master`, I meant my fork `sf314:master` (where I keep some custom modifications), as opposed to `libcsp:master`. I accidentally created a PR in the reverse direction, when what I really wanted to do was [update my fork](https://github.com/sf314/libcsp/pull/2) from the upstream.

johandc:
Okay, no problems. Happy that you are using it.

PR #349, opened 2022-03-30
CSP endian-ness fixes

rcunrau:
CSP allows you define the endian-ness of header fields, and honours that decision - for the most part. This changeset adds runtime network-order to host-order conversions wherever the packet header data is referenced.


johandc:
Endiannes conversion is done in csp_id when the interfaces gets ready to send the data.

At this point where your are making changes it is always in host order.

PR #355, opened 2022-04-15
include/csp/arch: add random functions header, use where possible

fjmolinas:
libc `rand` functions will not always be available, in a lot of cases embedded OS will have their own API for random number generation, so instead of using the `rand` functions, this PR proposes adding a `csp_rand.h` to be implemented by the different architectures.

I'm currently working with libCSP in RIOT, but I have not tested this PR for zephyr.

### DIFF
--- a/include/csp/arch/csp_rand.h
+++ b/include/csp/arch/csp_rand.h
@@ -1,0 +1,6 @@
+#pragma once
+
+#include <inttypes.h>
+
+uint8_t csp_rand8(void);
+uint16_t csp_rand16(void);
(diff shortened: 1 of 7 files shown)


johandc:
Actually both uses of rand can be avoided. So I will consider that instead.

johandc:
I suggest the block og code for adding random packet loss inside the can driver should simply be completely commented out.

And for random iss on RDP connect. This is really not needed since all RDP connections are running over UDP with ephemeral source ports. So only in the rare case of a source port wrapped and there is a connection clash, would the RDP random iss make a difference.

I suggest either setting it to 0. Or keeping a static variable and setting to the previously given value minus some large offset. But this is a waste of memory.

Generally getting rid on the reliance on random numbers is a good thing.

PR #356, opened 2022-04-15
src/csp_semaphore: fix function signature

fjmolinas:
This PR aligns the function signature so that all users `uint32_t` avoid functions of different signatures warning for some archs.

### DIFF
--- a/src/arch/freertos/csp_semaphore.c
+++ b/src/arch/freertos/csp_semaphore.c
@@ -14,7 +14,7 @@ void csp_bin_sem_init(csp_bin_sem_t * sem) {
 	return;
 }
 
-int csp_bin_sem_wait(csp_bin_sem_t * sem, unsigned int timeout) {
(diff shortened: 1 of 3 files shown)


johandc:
The timeout parameter can always fit into unsigned int regardless of the architecture.

In general I am a fan of using the most basic type necessary. Which is why this was changes into unsigned int in the first place.

The other reason is because all header files must be self sufficient, so all headers must include inttypes.h in order to build correctly. This just clutters the header files with includes. So I like the builtin types better where possible.

fjmolinas:
> The timeout parameter can always fit into unsigned int regardless of the architecture.
> 
> In general I am a fan of using the most basic type necessary. Which is why this was changes into unsigned int in the first place.
> 
> The other reason is because all header files must be self sufficient, so all headers must include inttypes.h in order to build correctly. This just clutters the header files with includes. So I like the builtin types better where possible.

Ok! make sense, closing this one then.

PR #357, opened 2022-04-15
src/csp_io: set idout->src when sending thruogh routing table

fjmolinas:
When using `csp_sendto` the src address is not filled in since its bound to be filled by `csp_send_direct`. But if it if did not find an iface matching the subnet then it will send via the routing table and the `src` address field will remain unset.

https://github.com/libcsp/libcsp/blob/08a983d8fab649900c70c55673f7ec5fd7925e02/src/csp_io.c#L318

I'm unsure if it could have been set somewhere else, so I added the check on wether it was still 0 before setting it.


johandc:
Thanks for fixing this one. 
I fixed it already on spaceinventor/libcsp master where I have been working lately
I am going to merge with develop soon

fjmolinas:
Thanks for the review!

PR #358, opened 2022-04-15
src/csp_dedup: fix empty initializers

fjmolinas:
This simply avoids the `-Wmissing-field-initializers` warning.


fjmolinas:
Thanks!

PR #359, opened 2022-04-15
src/csp_io.h: send direct pass idout struct by reference

fjmolinas:
This changes the `send_direct` api to pass the structure pointer to `id` by reference instead of by value. I'm not sure if this was done with the intention of having no side effects. In my case I actually want to know what was set as the destination address for messages, passing it by reference addresses this issue.

### DIFF
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -79,7 +79,7 @@ void csp_id_copy(csp_id_t * target, csp_id_t * source) {
 	target->flags = source->flags;
 }
 
-void csp_send_direct(csp_id_t idout, csp_packet_t * packet, csp_iface_t * routed_from) {
+void csp_send_direct(csp_id_t* idout, csp_packet_t * packet, csp_iface_t * routed_from) {
 
 	int from_me = (routed_from == NULL ? 1 : 0);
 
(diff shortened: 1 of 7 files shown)


src/csp_io.c
```diff
@@ -319,7 +322,9 @@ void csp_sendto(uint8_t prio, uint16_t dest, uint8_t dport, uint8_t src_port, ui
 	packet->id.sport = src_port;
 	packet->id.pri = prio;
 
-	csp_send_direct(packet->id, packet, NULL);
+	printf("set outgoing address %"PRIx16":%d\n", packet->id.src, packet->id.sport);
```
johandc: Stray printf should be cleaned before merge

johandc (changes requested):
I like it.

The reason for this is legacy from when a csp_id_t was a uint32, in a packed bitfield. So this improvement will avoid a lot of stack usage for this call.

src/csp_io.c
fjmolinas: ups

fjmolinas: addressed and squashed

fjmolinas:
Rebased to fix conflicts

johandc (approved):

fjmolinas:
Thanks for the review and merging!

PR #362, opened 2022-04-19
src/csp_rtable_stdio: set _DEFAULT_SOURCE to 1

fjmolinas:
This was required for me to build with newlib-nano:

```
/home/francisco/workspace/RIOT/build/pkg/libcsp/src/csp_rtable_stdio.c:19:25: error: implicit declaration of function 'strnlen'; did you mean 'strlen'? [-Werror=implicit-function-declaration]
   19 |  const size_t str_len = strnlen(rtable, 100);
      |                         ^~~~~~~
      |                         strlen
/home/francisco/workspace/RIOT/build/pkg/libcsp/src/csp_rtable_stdio.c:26:15: error: implicit declaration of function 'strtok_r'; did you mean 'strtok'? [-Werror=implicit-function-declaration]
   26 |  char * str = strtok_r(rtable_copy, ",", &saveptr);
      |               ^~~~~~~~
      |               strtok
/home/francisco/workspace/RIOT/build/pkg/libcsp/src/csp_rtable_stdio.c:26:15: error: initialization of 'char *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
/home/francisco/workspace/RIOT/build/pkg/libcsp/src/csp_rtable_stdio.c:59:7: error: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
   59 |   str = strtok_r(NULL, ",", &saveptr);
```

### DIFF
--- a/src/csp_rtable_stdio.c
+++ b/src/csp_rtable_stdio.c
@@ -1,3 +1,7 @@
+/* Required for strnlen, strtok_r in string.h, when building with -std=c99 */
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200809L
+#endif
 #include <stdio.h>
 #include <inttypes.h>
 
@@ -41,7 +45,7 @@ static int csp_rtable_parse(const char * rtable, int dry_run) {
 
 		csp_iface_t * ifc = csp_iflist_get_by_name(name);
 		if ((address > csp_id_get_max_nodeid()) || (netmask > (int)csp_id_get_host_bits()) || (ifc == NULL)) {
-			csp_dbg_errno = CSP_DBG_ERR_INVALID_RTABLE_ENTRY; 
+			csp_dbg_errno = CSP_DBG_ERR_INVALID_RTABLE_ENTRY;
 			return CSP_ERR_INVAL;
 		}
 


johandc (changes requested):
This is not the proper way to solve this. It Essentialy overrides the c library settings. 

I feel like the solution should be to include string.h explicitly.

fjmolinas:
> I feel like the solution should be to include string.h explicitly.

That was the first thing I tried, but sadly it did not work, the same error persisted. I'll see if I find another solution.

fjmolinas:
> > I feel like the solution should be to include string.h explicitly.
> 
> That was the first thing I tried, but sadly it did not work, the same error persisted. I'll see if I find another solution.

AFAIK `strnlen` and `strtok_r` are not part of the C standard, so when compiling with `-std=99` you don't get them by default, in `c11` you get `strnlen_s` and `strtok_s` but it's not exactly the same.

So the question is whether the used functions should be compliant with ISO C or POSIX. If the first then a  possible fix would be to simply replace those used functions by `strlen` or `strtok`, if POSIX, then I can define `_POSIX_C_SOURCE` instead of `_DEFAULT_SOURCE` to get exactly what is missing.

This is quite contained anyway, so if the preferred status is what is preferred, I don't think it will be too much of an issue for me to simply patch this locally.

fjmolinas:
> So the question is whether the used functions should be compliant with ISO C or POSIX. If the first then a possible fix would be to simply replace those used functions by `strlen` or `strtok`, if POSIX, then I can define `_POSIX_C_SOURCE` instead of `_DEFAULT_SOURCE` to get exactly what is missing.

Would in this case need to be `#define _POSIX_C_SOURCE=200809L` https://man7.org/linux/man-pages/man3/strnlen.3.html

PR #363, opened 2022-04-19
src/csp_buffer: buffer/skbuf statically defined data size

fjmolinas:
When compiling I get the following error/warning:

```
/home/francisco/workspace/RIOT/build/pkg/libcsp/include/csp/csp_types.h:161:25: error: ISO C forbids zero-size array ‘data’ [-Werror=pedantic]
  161 |                 uint8_t data[0];
      |                         ^~~~
/home/francisco/workspace/RIOT/build/pkg/libcsp/include/csp/csp_types.h:163:26: error: ISO C forbids zero-size array ‘data16’ [-Werror=pedantic]
  163 |                 uint16_t data16[0];
      |                          ^~~~~~
/home/francisco/workspace/RIOT/build/pkg/libcsp/include/csp/csp_types.h:165:26: error: ISO C forbids zero-size array ‘data32’ [-Werror=pedantic]
  165 |                 uint32_t data32[0];
```

I think that is from when the `buffer` API could be dynamically allocated, but maybe now that it is statically allocated this is no longer needed? If that is the case then maybe this PR can be useful by simplifying the size declaration.

### DIFF
--- a/include/csp/csp_types.h
+++ b/include/csp/csp_types.h
@@ -155,12 +155,9 @@ typedef struct csp_packet_s {
 	 * csp_buffer_init(), and can later be accessed by csp_buffer_data_size()
 	 */
 	union {
-		/** Access data as uint8_t. */
-		uint8_t data[0];
-		/** Access data as uint16_t */
-		uint16_t data16[0];
-		/** Access data as uint32_t */
-		uint32_t data32[0];
+		uint8_t data[CSP_BUFFER_SIZE];
+		uint16_t data16[CSP_BUFFER_SIZE / 2];
+		uint32_t data32[CSP_BUFFER_SIZE / 4];
 	};
 
 } csp_packet_t;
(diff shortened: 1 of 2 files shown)


fjmolinas:
Hmm I'm not using RDP, so I didn't catch that it is still using the union

johandc:
It is used here and there, just as a convenience when all you need to transfer is a single uint32 for example. But I think we can ditch. If you add the fix for the single usage in RDP, then I will go through our usages outside the library. But right now is the time, because 2.0 is not released yet.

johandc:
Another solution would be to just set data16[size/2] for example.

I like the idea that the compiler knows the length of the array because it actually does clever stuff to check for buffer overflows at compile time.

fjmolinas:
> Another solution would be to just set data16[size/2] for example.
> 
> I like the idea that the compiler knows the length of the array because it actually does clever stuff to check for buffer overflows at compile time.

I think that ius the easier solution! Did just that, seems like the CI is happy at least, thanks!

fjmolinas:
Seems like all is green, but I have not tested the RDP side

johandc:
Lets try it out in production and run some testing

fjmolinas:
Thanks for the review and merging!

PR #364, opened 2022-04-20
src/csp_rtable_stdio.c: replace strnlen, strtok_r by ISO_C versions

fjmolinas:
Alternative to https://github.com/libcsp/libcsp/pull/362, simply using non re-rentrant `strtok_r` and regular `strlen`

### DIFF
--- a/src/csp_rtable_stdio.c
+++ b/src/csp_rtable_stdio.c
@@ -1,5 +1,6 @@
 #include <stdio.h>
 #include <inttypes.h>
+#include <string.h>
 
 #include <csp/csp.h>
 #include <csp/csp_debug.h>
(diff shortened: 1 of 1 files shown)


johandc:
Using the non reentrant version of strtok pulls in a lot of unwanted thread synchronization and safety stuff from newlib such as the impure ptr.

See this topic
https://embdev.net/topic/129813

ISO C is not a target of CSP in itself. Where it makes sense, such as safer strnlen we should use that.

fjmolinas:
Should I close both related PRs and patch locally then?

johandc:
Why don't you build with C99? Or GNU99?

fjmolinas:
> Why don't you build with C99? Or GNU99?

`strnlen` is not part of the c99 standard, there is `strlen_s` but only in c11, but building with `gnu99` should do. Thanks!

PR #367, opened 2022-04-28
Updated CSP/CFP Spec

### LINKED ISSUE #365: CFP CAN 2.0 Frame Header Bits
I am a little confused with how the bits in the CAN Frame Header are used.

In the [spec](https://github.com/libcsp/libcsp/blob/f429e14f71d9f1cb5187eb6db29b7fcfa8cc373b/doc/CSP%202.0%20Specification%20-%20Ark1.pdf), the CFP CAN 2.0 Frame Header goes from bit 0 to bit 29 - which in total makes it 30 bits long. The fields are used something like this:

```
[29-28](2) Prio
[27-22](6) Dest
[21-16](6) Src
[15-14](2) Src Count
[13-11](3) Frame Count
[10] (1) Short
[09] (1) Begin
[08] (1) End
[07-00] (8) Data [n* 9] (why 9?)
Total = 2+6+6+2+3+1+1+1+8 = 30
```

In the [header file](https://github.com/libcsp/libcsp/blob/f429e14f71d9f1cb5187eb6db29b7fcfa8cc373b/include/csp/interfaces/csp_if_can.h) it says
```
   - Source:       5 bits
   - Destination:  5 bits
   - Type:         1 bit
   - Remain:       8 bits
   - Identifier:   10 bits
```

and

```
/**
 * CFP 2.0
 *
 * PRIO: 2
 * DST: 14
 * Sender id: 6
 * Sender counter: 2
 * Fragment counter: 3
 * Begin: 1
 * End: 1
 */
```

Could you help me understand how the bits in the identifier are used?

pfef-to:
Updated the spec document to the best of my understanding.

Closes #365


johandc:
Perfect, thanks!

PR #371, opened 2022-05-31
Typo fixes

bjornbm:
Also in doc/basic.md:

- Line 254: Should it be `"16/31 CAN 4"`?
- Line 263: Missing word, stray space, or stray comma in `interface ,`?

### DIFF
--- a/doc/basic.md
+++ b/doc/basic.md
@@ -150,7 +150,7 @@ from the connection pool:
 >     `csp_connect()`
 >   - server socket for listening
 >     `csp_socket()`
->   - server accepting an incmoing connection
+>   - server accepting an incoming connection
 >     `csp_accept()`
 
 An applications receive queue is located on the connection and is also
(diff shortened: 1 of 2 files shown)


yashi:
Hi @bjornbm

Thank you for the PR.

Here are the comments to your questions:

> * Line 254: Should it be `"16/31 CAN 4"`?

This one is for CSPv1, which only has 5 bits for address space.

From the current doc:

> \<address\>\[/mask\] \<interface name\> \[via\]
>
>   - mask (optional): determines how many MSB bits of address are to be
>     matched. mask = 1 will only match the MSB bit, mask = 2 will match
>     2 MSB bits. Mask values different from 0 and 5, is only supported
>     by the cidr rtable.

For CSPv2, we have 14 bits for the address space. Thus, if you want to route address from 16 to 31, you must write `"16/10 CAN 4"`

We still support CSPv1. So, I'd replace it with the current CSPv2 format, and make a note about CSPv1.

> * Line 263: Missing word, stray space, or stray comma in `interface ,`?

excessive space.  Please remove it.

PR #382, opened 2022-08-16
Fix CMake properties for CAN driver

jessicarod7:
Fixes an issue where a CMake configuration, if `libzmq` is found, fails due to properties not being set properly for the CAN driver directory:

```cmake
...
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.0") 
-- Checking for one of the modules 'libzmq'
-- Checking for one of the modules 'libsocketcan'
CMake Error at src/interfaces/CMakeLists.txt:23 (set_property):
  set_property could not find TARGET driver_can.  Perhaps it has not yet been
  created.


-- Configuring incomplete, errors occurred!
See also "/home/[USER]/develop/libcsp/build/CMakeFiles/CMakeOutput.log".
```

### DIFF
--- a/src/interfaces/CMakeLists.txt
+++ b/src/interfaces/CMakeLists.txt
@@ -20,7 +20,7 @@ if(LIBZMQ_FOUND)
   target_link_libraries(if_zmq PRIVATE ${LIBZMQ_LIBRARIES})
   target_link_libraries(libcsp PRIVATE if_zmq)
   if(BUILD_SHARED_LIBS)
-    set_property(TARGET driver_can PROPERTY POSITION_INDEPENDENT_CODE ON)
+    set_directory_properties(TARGET driver/can PROPERTY POSITION_INDEPENDENT_CODE ON)
   endif()
 endif()
 


jessicarod7:
Closing PR, actual issue was related to `libsocketcan` installation on Fedora. The library is not provided in DNF, so it has to be compiled and installed from source, and then `pkgconf` environment vars need to be updated with `PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/lib`.

yashi:
@cam-rod, thank you for reporting this.  This is actually a bug. Would you reopen?

yashi:
@cam-rod, please test the PR if you have time.  Thanks.

jessicarod7:
I'm still unable to get `csp_server_client` to run from any of the CMake builds (although Meson does work, so I think that's a separate issue entirely), but it successfully builds, so yeah the fix in #383 works. Thanks!

PR #386, opened 2022-09-26
Taylorr/buffer get improvement

### LINKED ISSUE #385: csp_buffer_get does not set the length field of the packet buffer it returns.
It's strange to have to set the length of the packet manually when the function to return a packet has a size parameter

Taylorr82:
I meant to make this one :)

This fixes https://github.com/libcsp/libcsp/issues/385 by setting the length field of the gotten buffer before return.

This changes the definition of csp_buffer_get to explicitly return a csp_packet_t *. The docs say it returns a csp_packet_t * and everywhere in the code treats it as a csp_packet_t * yet it was declared with void * return.

### DIFF
--- a/include/csp/csp_buffer.h
+++ b/include/csp/csp_buffer.h
@@ -15,7 +15,7 @@
    @param[in] data_size minimum data size of requested buffer.
    @return Buffer (pointer to #csp_packet_t) or NULL if no buffers available or size too big.
 */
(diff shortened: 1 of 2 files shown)


johandc:
Thank you very much for the contribution. The change has been merged into my working branch, and another pull request opened for that merge.

PR #390, opened 2022-11-09
Fixed reception MTU by incorporating header offset.

### LINKED ISSUE #389: Received CAN messages with size close to the MTU get dropped on receive
In csp_if_can.c CSP V2 CAN message reception starts writing at packet->frame_begin, which is offset by -6 bytes from packet->data. The overflow check uses packet->frame_length, dlc and iface->mtu to determine overflow. This fails to compensate for the -6 bytes offset. This means that the actual MTU is 6 bytes less than the iface->mtu. Packets get sent properly at the MTU length, but dropped on reception. Only packets that are 6 bytes under the MTU get sent properly.

Proposed fix is subtracting 6 from the calculated packet size, as CSP_ID2_HEADER_SIZE from csp_id.c is not visible in csp_if_can.c. There is also plenty of other code that has implicit assumptions on the header size. I will attach a pull request. If this needs to be solved more systematically this define needs to be moved to the header file.

ghvisser:
This fixes #389.

### DIFF
--- a/src/interfaces/csp_if_can.c
+++ b/src/interfaces/csp_if_can.c
@@ -335,7 +335,7 @@ int csp_can2_rx(csp_iface_t * iface, uint32_t id, const uint8_t * data, uint8_t 
 	}
 
 	/* Check for overflow */
-	if (packet->frame_length + dlc > iface->mtu) {
+	if (packet->frame_length + dlc - 6 > iface->mtu) {
 		csp_dbg_can_errno = CSP_DBG_CAN_ERR_RX_OVF;
 		iface->frame++;
 		csp_can_pbuf_free(ifdata, packet, 1, task_woken);


johandc:
I think this is correct behavior. The MTU defines the maximum size of the frame on the interface level. So that includes the network layer (CSP) header.

Take Ethernet for example, it's MTU is typically 1500 bytes. Not including the ethernet header (total 1518 bytes). But it does include the IP, IPX or whatever protocol is transferred over the ethernet frame. 

So there problem must sending side its producing too long frames. (The real CAN limitation comes from the maximum frame length in the CFP field.)

ghvisser:
I can make an IP packet many bytes longer than the MTU of Ethernet and it still works, because fragmentation is handled there. For a CSP packet the limiting factor seems to be the MTU (-6 bytes for header). So, the analogy with IP does not really work. The MTU for Ethernet would be more analogous to the maximum CAN frame size (8 bytes), not the CSP packet size.

That spawns two questions:

1. If the current behavior is correct, then the maximum packet size is calculated differently on the RX and TX sides, should then the RX calculation be modified? For context, I can send an MTU length message on wire fine with the CSP stack, it's just the receive side that drops it.
2. What is the intended way for a user to find the maximum packet size?

What also complicates matters, is that the actual CSP 2.0 header is mixed in with the CFP header in the [specification](https://github.com/libcsp/libcsp/blob/develop/doc/CSP%202.0%20Specification%20-%20Ark1.pdf). At least on wire.

EDIT:
MTU is also determined like this:
iface->mtu = csp_buffer_data_size() - 8;

So, as used in the protocol it is currently determined by the receive and transmit buffer sizes, not link-layer limitations. My view on the issue here is that it is a miscalculated overflow condition, as the buffer is not yet filled when the overflow triggers (6 bytes early, because of header).

johandc:
Please see https://github.com/libcsp/libcsp/pull/391 for a fix

PR #391, opened 2022-11-12
Nomtu, crc32 on connless, CSP 1 L2 broadcast + buffer leak fix, UDP interface single socket

johandc:
Hello Instead of creating 4 pull requests and merging in order, i have joined our current development tree into a single PR.

- Adressing #390 i have decided to remove the iface->mtu field, and go with a simpler approach. Please read the commit message for that change.
- CRC32 is now required on incoming connection-less packets. Usually it is the one who calls csp_connect or csp_socket, who sets the options. But for connection-less (callback pattern) messages the options was set to 0x0, which meant any it was accepting garbage data as valid csp packets occasionally. If there is no higher layer check (csp_ping, poke and peek for example has none) it could mean writing data to memory at arbitrary memory fields. So this was a security concern that is mitigated by requiring CRC32 to be valid on all connection-less incoming packets. As a consequence, a csp_ping, peek poke 
client functions have added csp_o_crc32.
- CSP 1 supports CAN broadcast on address 31 now
- One buffer leak and one null pointer check have been fixed.
- The UDP interface have been modified slightly, so instead of using two different sockets (one for each direction) it uses a single socket. This makes it compatible with the satlab implementation of csp_if_udp (https://github.com/satlab/libcsp/blob/satlab-master/src/interfaces/csp_if_sludp.c) which eventually could get jointed into master.


johandc:
Also include the changes from Taylor's fix of https://github.com/libcsp/libcsp/issues/385.

csp buffer get now explicitly returns a csp packet, and the size field has been renamed and the documentation changed to show that the size field is no longer used.

src/interfaces/csp_if_kiss.c
```diff
@@ -77,8 +77,8 @@ void csp_kiss_rx(csp_iface_t * iface, const uint8_t * buf, size_t len, void * px
 		/* Input */
 		uint8_t inputbyte = *buf++;
 
-		/* If packet was too long */
-		if (ifdata->rx_length >= ifdata->max_rx_length) {
+		/* If packet was too long, truncate and restart */
+		if (&ifdata->rx_packet->frame_begin[ifdata->rx_length] >= &ifdata->rx_packet->data[sizeof(ifdata->rx_packet->data)]) {
```
yashi: Why do you have to replace `rx_length` with the pointer arithmetic?

Is this fixing a bug not considering the header size in the original?  If so, we should fix it in another commit or at least add a comment in the commit log.

yashi (commented):
just a quick review.  I'll review more in the coming week.

src/interfaces/csp_if_kiss.c
johandc: Understood one is looking more nice in this context.
But bear in mind that somewhere the max_rx_length parameter needs to be set.  So we are deleting a variable, and a place for the programmer to make an error.
The pointer arithmetic checks that the actual write address does not exceed the length of the actual packet allocation. It looks meesy, and maybe we could make it look nicer by adding a middle compile time defined const.

const int max_rx_length = sizeof(packet->data) + csp header length.

However, the last part of the header length is a variable. So ill take that back again.

If you can find a prettier method to check the overflow, without adding a variable to the data structure, or god forbid, another MTU field. You are very welcome to do so. 

I actually just considered adding a single sentinel byte to the end of the packet_t structure, which was easy to just take the address of for frame_end comparisons. What do you think about that? Furthermore we could set it to a known good sentinel value 0x5A for example, and check for buffer overflows on buffer_free too.

src/csp_yaml.c
```diff
@@ -208,8 +208,10 @@ void csp_yaml_init(char * filename, unsigned int * dfl_addr) {
 
 	csp_print("  Reading config from %s\n", filename);
 	FILE * file = fopen(filename, "rb");
-	if (file == NULL)
+	if (file == NULL) {
+		printf("  ERROR: failed to find CSP config file\n");
```
yashi, line 210: it may not be an issue but should we use `csp_print()` instead?

src/interfaces/csp_if_udp.c
```diff
@@ -18,35 +20,18 @@ static int csp_if_udp_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packe
 
 	csp_if_udp_conf_t * ifconf = iface->driver_data;
 
-	int sockfd;
-	if ((sockfd = socket(AF_INET, SOCK_DGRAM, 0)) < 0) {
-		return CSP_ERR_BUSY;
+	if (ifconf->sockfd == 0) {
+		csp_buffer_free(packet);
+		return CSP_ERR_NONE;
```
yashi: Why are you freeing the packet when `sockfd` is 0?  `socket()` returns a negative for errors, and 0 and positive for valid file descriptors.

src/interfaces/csp_if_udp.c
```diff
@@ -18,35 +20,18 @@ static int csp_if_udp_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packe
...
-	close(sockfd);
-
 	return CSP_ERR_NONE;
 }
 
-int csp_if_udp_rx_get_socket(int lport) {
-
-	int sockfd = socket(AF_INET, SOCK_DGRAM, 0);
-
-	struct sockaddr_in server_addr = {0};
```
yashi, line 41: No need to initialize as `0`.  We are initializing all three members, `sin_family`, `sin_port`, and `sin_addr`, right after this.

src/interfaces/csp_if_udp.c
```diff
@@ -80,19 +65,29 @@ void * csp_if_udp_rx_loop(void * param) {
...
+		ifconf->sockfd = socket(AF_INET, SOCK_DGRAM, PF_PACKET);
+
+		struct sockaddr_in server_addr = {0};
+		server_addr.sin_family = AF_INET;
+		server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
+		server_addr.sin_port = htons(ifconf->lport);
+
+		bind(ifconf->sockfd, (struct sockaddr *)&server_addr, sizeof(server_addr));
+		
+		if (ifconf->sockfd < 0) {
```
yashi, line 83: too late to test it.  If we test, test it after `socket()`.     If we want to, we should test the return value of `bind()` here instead of `ifconf->sockfd`.  And make sure to `ifconf->sockfd = -1` when `bind()` failed`.   Otherwise, you leave the while loop.

src/interfaces/csp_if_udp.c
```diff
@@ -80,19 +65,29 @@ void * csp_if_udp_rx_loop(void * param) {
...
 	csp_iface_t * iface = param;
 	csp_if_udp_conf_t * ifconf = iface->driver_data;
-	int sockfd = -1;
 
-	while (sockfd < 0) {
-		sockfd = csp_if_udp_rx_get_socket(ifconf->lport);
-		if (sockfd < 0) {
+	while (ifconf->sockfd < 0) {
+
+		ifconf->sockfd = socket(AF_INET, SOCK_DGRAM, PF_PACKET);
```
yashi, line 74: Why using `PF_PACKET`?

src/interfaces/csp_if_udp.c
```diff
@@ -80,19 +65,29 @@ void * csp_if_udp_rx_loop(void * param) {
...
+
+		struct sockaddr_in server_addr = {0};
+		server_addr.sin_family = AF_INET;
+		server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
+		server_addr.sin_port = htons(ifconf->lport);
+
+		bind(ifconf->sockfd, (struct sockaddr *)&server_addr, sizeof(server_addr));
+		
+		if (ifconf->sockfd < 0) {
 			csp_print("  UDP server waiting for port %d\n", ifconf->lport);
```
yashi, line 84: The message is a bit bogus.  What do you mean by "waiting for port X"?   Are we waiting for the specified port to be available?  If so, we should be testing `bind()`.

With the current form, we are printing the message when `socket()` failed.

src/interfaces/csp_if_udp.c
```diff
@@ -66,7 +69,7 @@ void * csp_if_udp_rx_loop(void * param) {
 	csp_iface_t * iface = param;
 	csp_if_udp_conf_t * ifconf = iface->driver_data;
 
-	while (ifconf->sockfd < 0) {
+	while (ifconf->sockfd == 0) {
```
yashi, line 72: I think using a valid file descriptor as an error is wrong.

src/csp_route.c
```diff
@@ -175,7 +175,7 @@ int csp_route_work(void) {
 	csp_callback_t callback = csp_port_get_callback(packet->id.dport);
 	if (callback) {
 
-		if (csp_route_security_check(CSP_SO_NONE, input.iface, packet) < 0) {
+		if (csp_route_security_check(CSP_SO_CRC32REQ, input.iface, packet) < 0) {
```
yashi, line 179: It'd be nice if we add a commit message from the comment on this PR.

yashi (changes requested):
Sorry to be late.  I've review a few additional commits.   The biggest issue I've found is that using a valid file descriptor `0` as an error state.  Did I miss your intention?

src/arch/freertos/meson.build
```diff
@@ -3,7 +3,6 @@ csp_sources += files([
 	'csp_semaphore.c',
 	'csp_system.c',
 	'csp_time.c',
-	'csp_clock.c',
```
yashi, line 6: Don't you want to remove `src/arch/freertos/csp_clock.c` as well?

src/interfaces/csp_if_can_pbuf.c
```diff
@@ -45,7 +45,7 @@ void csp_can_pbuf_free(csp_can_interface_data_t * ifdata, csp_packet_t * buffer,
 
 csp_packet_t * csp_can_pbuf_new(csp_can_interface_data_t * ifdata, uint32_t id, int * task_woken) {
 
-	csp_can_pbuf_cleanup(ifdata);
+	csp_can_pbuf_cleanup(ifdata, task_woken);
```
yashi, line 48: Instead of passing `task_woken` to indicate that the call is from ISR or not, why not make `csp_get_ms()` callable from any context including ISR?

Or, does FreeRTOS have an unified API to test that it's in ISR or not?  I see `xPortIsInsideInterrupt` or `uxInterruptNesting` is used to test it.  But my FreeRTOS knowledge is limited.

Or, should we have `_isr()` variant of it?  We have two patterns for functions callable from ISR.  Functions with `_isr` suffix and additional argument `task_woken`.

src/csp_yaml.c
johandc: Well yes. Maybe i just got complacecent because yaml is only used on PCs anyways. But of course

src/interfaces/csp_if_udp.c
johandc: Good catch.

The major change is that the interface must not return ERR_BUSY, because we do not support blocking csp send anyways. So if the outgoing buffer is full, just discard the message, and rely on flow control to fix the issue.

src/interfaces/csp_if_udp.c
johandc: ok

src/interfaces/csp_if_udp.c
johandc: Good question. I dont know. The major change in csp_if_udp came from syncing with the satlab implementation of the same interface. Maybe i got it from there, or maybe we already used PF_PACKET. As far as i can read from the MAN page, PF_PACKET is the type we want, so i dont think its wrong.

src/interfaces/csp_if_udp.c
johandc: Okay

src/interfaces/csp_if_udp.c
johandc: Okay. Yes. We should be waiting on bind. We hardly ever use this interface, so it has not seen much every day testing. So i have not run into this as an issue. But you are correct.

src/arch/freertos/meson.build
johandc: I dont know, what is your feeling about this change?

I had to mess with nasty static library include orders, because GCC has no concept of full scanning all included static libraries unless everything is linked with --link-whole. But this possesses other major problems. So i think i would like to get rid of __weak__ linking alltogether. If nobody talks me out of it??

src/interfaces/csp_if_can_pbuf.c
johandc: Its a bit messy indeed.

FreeRTOS does have functions to tell if we are inside an ISR. Which would be the cleanest method of making functions callabel from both contexts. However, FreeRTOS own API requires us to keep this TaskWoken parameter in userspace, instead of having an internal variable it checks when exiting the ISR. 

So the reason we follow this pattern is not to inform our own functions which API function to use. Its actually to do FreeRTOS dirty work, and moving the task_woken parameter with us all the way through. 

CSP was written on FreeRTOS first, so the implementation of their API has influenced the CSP api, so that we have _isr functions and stuff. So this is not related to this pull request itself. But we might want to create a separate issue to cleanup this leaky API some day.

src/interfaces/csp_if_udp.c
yashi: correct me if I'm mistaken, but `socket(2)` takes protocol number as the third argument.  You don't specify `PF_PACKET` as the third.

If you want to use cooked UDP, use
```c
socket(AF_INET, SOCK_DGRAM, protonum);
```

If you want to use packet interface with raw packets, do
```c
socket(AF_PACKET, SOCK_RAW, protonum);
```

Or, if you want to use packet interface with cooked packet,
```c
socket(AF_PACKET, SOCK_DGRAM, protonum);
```

You can use `PF_` prefix instead of `AF_`, ie) `PF_INET` instead of `AF_INET`, or `PF_PACKET` instead of `AF_PACKET`.

src/arch/freertos/meson.build
yashi: Hm... I don't really have a problem with weak refs. Maybe that's because I rarely replace weak ref symbols on my system.  Or Zephyr build system is doing what I want to do.  Not sure.

Regarding `csp_clock_{get,set}_time()`, I'm totally fine removing the API as well.  We are not using it in the CSP stack at all.  These are just handy wrappers for POSIX `clock_gettime()`.

Issue #322, opened 2021-12-21
python3 binding is broken


yashi:
With the commit 407e89661d7, `csp_socket_t` is no longer `csp_conn_t`.  This caused various parts of the bind to be broken.

johandc:
Oh yes

stephan-ack:
I'm guessing this is also related to this error below that I am seeing? Afaik, "address" has been moved away from he csp_conf_s struct in CSP2.0?

```
../src/bindings/python/pycsp.c:105:47: error: ‘csp_conf_t’ {aka ‘struct csp_conf_s’} has no member named ‘address’
  105 |  if (!PyArg_ParseTuple(args, "bsss", &csp_conf.address, &csp_conf.hostname, &csp_conf.model, &csp_conf.revision)) {
```

johandc:
Yes, in CSP 2.0 each interface has it's own address. So there is no longer a global CSP node id.

gsantamarina:
Will this be fixed soon? Getting the same error when trying to compile the python3 bindings.

PR #403, opened 2023-06-23
Fix python bindings

sigmundklaa:
Fixes the Python bindings that were broken in CSP 2.0, discussed in #322.

Summary:
* Added a small example script as the current example scripts are quite big and I figured a minimal example could be good to have.
* An optional argument `version` was added to `pycsp_init` to allow the user to specify CSP version.
* The address that is passed to `pycsp_init` was removed because it is no longer used. This can be readded for backwards compatibility, but from my understanding the bindings have not been working at all so this should not break any functioning codebase. Please correct me if that is wrong, and I will readd the argument and just not use it.
* `pycsp_socket` caused a segmentation fault due to a pointer to a stack allocated socket being used. This was changed to instead use pythons memory allocator for the allocation of the socket.


yashi:
Hi @sigmundklaa thanks for the pr. I'll take a look in a week.

yashi (commented):
~Would you please write in the commit log why you need to "remove address from pycsp init"?~

Sorry.  you wrote it in the first comment.

> The address that is passed to pycsp_init was removed because it is no longer used. This can be readded for backwards compatibility, but from my understanding the bindings have not been working at all so this should not break any functioning codebase. Please correct me if that is wrong, and I will readd the argument and just not use it.

wscript
```diff
@@ -73,7 +73,7 @@ def configure(ctx):
 
     # Setup CFLAGS
     if (len(ctx.stack_path) <= 1) and (len(ctx.env.CFLAGS) == 0):
-        ctx.env.prepend_value('CFLAGS', ["-std=gnu11", "-g", "-Os", "-Wall", "-Wextra", "-Wshadow", "-Wcast-align",
+        ctx.env.prepend_value('CFLAGS', ["-std=gnu11", "-g", "-O0", "-Wall", "-Wextra", "-Wshadow", "-Wcast-align",
```
yashi: this is notiong to do with "removing address", no?

examples/python_bindings_example_server.py
```diff
@@ -126,4 +126,5 @@ def csp_server():
     libcsp.print_routes()
 
     # start CSP server in a thread
-    threading.Thread(target=csp_server).start()
+    #threading.Thread(target=csp_server).start()
+    csp_server()
```
yashi: nitpicking: no newline

src/bindings/python/pycsp.c
```diff
@@ -73,10 +73,6 @@ static void pycsp_free_csp_conn(PyObject * obj) {
 }
 
 static void pycsp_free_csp_socket(PyObject * obj) {
-	csp_conn_t * socket = (csp_conn_t *)get_obj_as_socket(obj, true);
-	if (socket) {
-		csp_close(socket);
-	}
```
yashi: If you don't close the socket, you are leaking file descritors, no?

examples/csp_server_client.py
```diff
@@ -0,0 +1,83 @@
+
```
yashi, line 1: would be nice if you can format according to pep8.

wscript
```diff
@@ -73,7 +73,7 @@ def configure(ctx):
 
     # Setup CFLAGS
     if (len(ctx.stack_path) <= 1) and (len(ctx.env.CFLAGS) == 0):
-        ctx.env.prepend_value('CFLAGS', ["-std=gnu11", "-g", "-O0", "-Wall", "-Wextra", "-Wshadow", "-Wcast-align",
+        ctx.env.prepend_value('CFLAGS', ["-std=gnu11", "-g", "-Os", "-Wall", "-Wextra", "-Wshadow", "-Wcast-align",
```
yashi: This is the change you did in the first commit "remove address from pycsp init".  If you don't need it remove it from the first commit.

examples/python_bindings_example_server.py
```diff
@@ -126,5 +126,4 @@ def csp_server():
     libcsp.print_routes()
 
     # start CSP server in a thread
-    #threading.Thread(target=csp_server).start()
-    csp_server()
\ No newline at end of file
+    threading.Thread(target=csp_server).start()
```
yashi: And this is not related to `-O0` removal.

examples/csp_server_client.py
```diff
@@ -54,10 +54,10 @@ def client_task(addr: int) -> None:
         if packet is None:
             raise Exception('Failed to get CSP buffer')
         
-        data = bytes(f'Hello World {chr(count)}', 'ascii') + b'\x00'
+        data = bytearray(f'Hello World {chr(count)}', 'ascii') + b'\x00'
```
yashi, line 61: Why do you need `bytearray` instead of `bytes`?

src/bindings/python/pycsp.c
```diff
@@ -129,10 +129,10 @@ static PyObject * pycsp_socket(PyObject * self, PyObject * args) {
 		return NULL;  // TypeError is thrown
 	}
 
-	csp_socket_t * socket = malloc(sizeof(*socket));
+	csp_socket_t * socket = PyMem_RawMalloc(sizeof(*socket));
```
yashi, line 133: Is this the one you mentioned in the first comment?

> pycsp_socket caused a segmentation fault due to a pointer to a stack allocated socket being used. This was changed to instead use pythons memory allocator for the allocation of the socket.

CMakeLists.txt
```diff
@@ -70,12 +70,14 @@ target_compile_options(csp PRIVATE ${CSP_C_ARGS})
 add_subdirectory(src)
 add_subdirectory(examples)
 
-if(${enable-python3-bindings})
-  find_package(Python COMPONENTS Development.Module)
+if(${CSP_ENABLE_PYTHON3_BINDINGS})
+  find_package(Python COMPONENTS Development)
   if(Python_Development.Module_FOUND)
-    Python_add_library(libcsp_py3 src/bindings/python/pycsp.c WITH_SOABI)
+    Python_add_library(libcsp_py3 STATIC src/bindings/python/pycsp.c)
```
yashi: These lines should be something like:
```cmake
if(${CSP_ENABLE_PYTHON3_BINDINGS})
  find_package(Python3 COMPONENTS Development.Module)
  if(Python3_Development.Module_FOUND)
    Python3_add_library(libcsp_py3 MODULE src/bindings/python/pycsp.c)
```

Because 
1. We don't support python2 any more
2. We are targetting module and module only

yashi:
Overall, thank you for your time on fixing the python binding.  We really appreciate it.  We haven't have good commit messages in libcsp but it'd be really helpful if you can put "why"s in each commit message.
Let me know if you need any help.  Thanks.

src/bindings/python/pycsp.c
sigmundklaa: Good catch, the port will remain open and associated with the freed socket. `csp_close` is not correct though as that works on `csp_conn_t`, not `csp_sock_t`, so I will make a new function to perform cleanup on a socket.

src/bindings/python/pycsp.c
sigmundklaa: No, it was previously stack-allocated and I changed it to malloc. After looking at the [Python docs](https://docs.python.org/3/c-api/memory.html) I later decided to change it to the Python allocator instead as I figured that was more fitting.

examples/csp_server_client.py
sigmundklaa: Yes, will do

examples/csp_server_client.py
sigmundklaa: `packet_set_data` does not accept read-only bytes object:
Screenshot 2023-07-10 at 09 16 59

examples/python_bindings_example_server.py
sigmundklaa: Should be fixed with pep8

wscript
sigmundklaa: Yes this was accidentally commited, my bad. What do you mean by "remove it from the first commit" though? Is there another way of doing it other than reverting and making a new commit?

CMakeLists.txt
sigmundklaa: Updated

wscript
yashi: It depends on your git client (aka git porcelain).  A porcelain like [Magit](https://magit.vc/) let you move a hunk from a commit and merge the hunk to the other, which effectively fix this kind of mistake easily.

examples/csp_server_client.py
yashi: Yeah, that's because we have `"Ow*"`.  But to me, It seems the function just does `memcpy()` from the passed argument to `packet->data`.  I don't know why it can't be read-only.

sigmundklaa:
I will look at resolving the rest of the comments next week hopefully, will also try to do better commit names. Agree the commit messages are quite bad so im tempted to make a brand new pull request with better commits, what do you think @yashi ?

yashi:
@sigmundklaa Sure, that's fine by me.  TIA

PR #405, opened 2023-07-03
Fix calling non-ISR functions from CAN ISR

Abestanis:
[`csp_can1_rx`](https://github.com/libcsp/libcsp/blob/12326d109e32672e83a95e728b9579a66ee47802/src/interfaces/csp_if_can.c#L62) calls [`csp_can_pbuf_new`](https://github.com/libcsp/libcsp/blob/12326d109e32672e83a95e728b9579a66ee47802/src/interfaces/csp_if_can_pbuf.c#L48), which calls [`csp_can_pbuf_free`](https://github.com/libcsp/libcsp/blob/12326d109e32672e83a95e728b9579a66ee47802/src/interfaces/csp_if_can_pbuf.c#L13). All of these functions can be called from an interrupt service routine, which is why `csp_can1_rx` and `csp_can_pbuf_new` accept a `task_woken` parameter that is non-`NULL` if called in an ISR. But that parameter is not passed to `csp_can_pbuf_free`, which then calls functions that are not safe to call from an ISR, eventually causing an assertion in FreeRTOS queue code.

The fix is just to propagate the `task_woken` and call the ISR safe functions if it is not `NULL`.

### DIFF
--- a/src/interfaces/csp_if_can_pbuf.c
+++ b/src/interfaces/csp_if_can_pbuf.c
@@ -45,7 +45,7 @@ void csp_can_pbuf_free(csp_can_interface_data_t * ifdata, csp_packet_t * buffer,
 
 csp_packet_t * csp_can_pbuf_new(csp_can_interface_data_t * ifdata, uint32_t id, int * task_woken) {
 
-	csp_can_pbuf_cleanup(ifdata);
+	csp_can_pbuf_cleanup(ifdata, task_woken);
 
 	uint32_t now = (task_woken) ? csp_get_ms_isr() : csp_get_ms();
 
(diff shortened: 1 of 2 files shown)


yashi:
I agree that we need to pass `task_woken` to any function using `csp_get_ms_isr()` or any other ISR related functions.  And I approve this PR.

But, the root cause of this is getting ridicurous.  We should have abstract version of `csp_get_ms()` or we should drop  `portTICK_TYPE_IS_ATOMIC == 0` of FreeRTOS.  Does any one using FreeRTOS with non atomic tick?

I can't even find one arch on the official repo.

```
git grep --recurse-submodules -P '#define portTICK_TYPE_IS_ATOMIC.+0'
FreeRTOS/Source/include/FreeRTOS.h:    #define portTICK_TYPE_IS_ATOMIC    0
```
```
git grep --recurse-submodules -P '#define portTICK_TYPE_IS_ATOMIC.+1' | wc -l
81
```

Abestanis:
Our tick type is atomic, so we never encountered a problem with calling the `csp_get_ms` from the ISR, but `csp_buffer_free` was crashing consistently because it calles some queue function from FreeRtos which then asserts that it is not in an interrupt handler context.

So even if you drop the support for non-atomic tick types and declare `csp_get_ms` as ISR safe, I don't think it's possible to get around passing the `task_woken`, especially because that information about whether or not a task was woken needs to be propagated up to the interrupt handler so it can do a context switch and make sure the correct task is resumed after exiting from the ISR.

yashi (approved):
@Abestanis Thanks.  You are right that it's not only `csp_get_ms()`.

Abestanis:
Awesome, thank you! 🚀

johandc:
Im sorry for the inconvenience.

We have the same fix in spaceinventor/libcsp but i havent merged it back to upstream. Your fix is good.

PR #407, opened 2023-07-07
Update list of known satellites using CSP

Qerbz:
NUTS got stopped and Orbit NTNU filled its place. Selfiesat is in orbit, FramSat-1 and 1.5 are in late stages of development


yashi:
I feel like we should keep satellites even after their missions have ended. What do you think? And would it be better if we mentioned the primary source or at least some public information or a quote from people involved in the mission?

Last but not least, I would greatly appreciate it if you could copy and paste your comment into the commit message.

Thank you.

PR #410, opened 2023-07-12
include: Add extern "C"

### LINKED ISSUE #409: When building a C++ program, need to add extern "C" block around libcsp includes.
Compiler: g++ 11.3.0

Expected behavior: C++ programs compile and link against libcsp just by including the appropriate header files and linking the object files.

Actual behavior: Get the linker error ``undefined reference to `csp_init()'``

We can work around this just fine for now by adding `extern "C"` blocks around the libcsp includes.  However, if libcsp pulled in any libraries that had their own C++-related macros, this workaround might not work.

---

Reproducible example: In a new project directory:

Clone libcsp into `lib/csp` (as done here: https://github.com/spaceinventor/csh)

meson.build file:
```
project('TestProgram',
	'cpp',
	subproject_dir: 'lib'
)

csp_dep = dependency('csp', fallback: ['csp', 'csp_dep'], required: true)

main = executable(
	'main',
	[ 'src/main.cpp' ],
	dependencies: [csp_dep]
)
```

src/main.cpp file.  Program compiles and links after `extern "C"` lines uncommented:
```
// extern "C" {
#include <csp/csp.h>
// }

int main() {
  csp_init();

  return 0;
}
```

yashi:
Add extern "C" to most of header files.  This allows us to link libcsp against C++ app.

This commit also fixes a few white space issues.  No functionality should have been changed.

This fixes #409

### DIFF
--- a/include/csp/arch/csp_queue.h
+++ b/include/csp/arch/csp_queue.h
@@ -1,5 +1,3 @@
-
-
 #pragma once
 
 #include <inttypes.h>
@@ -9,6 +7,10 @@
 #if (CSP_FREERTOS)
 #include <FreeRTOS.h>
 #include <queue.h>
+#endif
+
+#ifdef __cplusplus
+extern "C" {
 #endif
 
 #define CSP_QUEUE_OK 0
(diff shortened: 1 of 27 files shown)


johandc:
Incredible how this pollution still has to take place.
I would have hoped it could be solved with a simple #pragma lang-c or something these days.

PR #411, opened 2023-07-16
Fix python bindings

### LINKED ISSUE #322: python3 binding is broken

sigmundklaa:
This is based on #403 , but with better commit names and also the following:
* Changed argument parsing in pycsp to only work with primitive types and instead convert it to the fixed-width types defined in `stdint.h` after argument parsing (See https://github.com/libcsp/libcsp/commit/1f6a004fc2863dd48ee3a244c0173f60bc3f23e6 for details.)
* `pycsp_init` now accepts all the fields of `csp_conf_t` as arguments
* Adds a `csp_socket_close` for properly closing a socket.

Resolves #322


examples/csp_server_client.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+import time
+import threading
+import libcsp_py3 as csp
+from typing import Any, Callable
+
+
+def printer(node: str, color: str) -> Callable:
+    def f(inp: str) -> None:
+        print(f'{color}[{node.upper()}]: {inp}\033[0m')
```
yashi: If you want to use f-string, we need  Python >= 3.6.  And I agree to use >= 3.6 since Ubuntu 20.04 LTS has 3.8.  The check is in `wscript`, `meson.build`.

Would be great to add the check to `CMakeLists.txt`.

wscript
```diff
@@ -158,6 +158,9 @@ def configure(ctx):
     if ctx.options.enable_python3_bindings:
         ctx.check_python_version((3,5))
         ctx.check_python_headers(features='pyext')
+        # type-limits can trigger a 'false' error when compiling Python bindings
+        # due to explicit bounds-checking of parsed Python arguments
+        ctx.env.append_unique('CFLAGS', ['-Wno-type-limits'])
```
yashi: This is just hiding the warning about comparing unsigned value against `< 0`.  Most of the call site is unsigned and easy to fix.

For `conn_queue_len = conn_queue_len_parsed`, LHS is `size_t` and RHS is `ssize_t`.  `csp_listen()` takes `size_t`.  The parser unfortunately doesn't have `size_t`.   If you want to `Py_ssize_t`, you must check for negative value.  But, I'd suggest using `k` (unsinged long) for that.

src/bindings/python/pycsp.c
```diff
@@ -9,11 +9,49 @@
...
+/**
+ * Macro that assigns the value of @p src to @p dst. If @p src is not
+ * within the given range @p min_value - @p max_value, an OverflowError is set
+ * and the expression evaluates to false. On success, the expression evalutes to
+ * true.
+ *
+ * This macro is useful for parsing arguments in a Python function. Python
+ * only supports the native types like unsigned int, signed long etc. whereas CSP
+ * uses the fixed-width integer types defined in <stdint.h>. This can cause
+ * problems if we pass a uint16_t* to PyArg_ParseTuple that expects unsigned
```
yashi: Let's check this in build time rather than runtime.

Meson:
```
cc.sizeof('uint16_t', prefix : '#include <stdint.h>')
```

CMake
```
include (CheckTypeSize)
check_type_size(uint16_t SIZEOF_UINT16)
```

I'm sure there is a way to check size of a type in Waf as well.  (I'm not an expart of Waf)

src/csp_port.c
```diff
@@ -129,3 +129,32 @@ int csp_bind_callback(csp_callback_t callback, uint8_t port) {
...
+
+	for (size_t i = 0; i < CSP_PORT_MAX_BIND + 2; i++) {
+		csp_port_t * port = &ports[i];
+
+		if (port->state == PORT_OPEN && port->socket == sock) {
+			port->state = PORT_CLOSED;
+			port->socket = NULL;
+			break;
+		}
+	}
```
yashi: You first need to set `port->socket` to `NULL` because that's how the other ends check to see the port is active.  otherwise, you have a race condition between consumer (`csp_socket_close` in this case) and producer (the other end)

src/csp_port.c
```diff
@@ -129,3 +129,32 @@ int csp_bind_callback(csp_callback_t callback, uint8_t port) {
...
+
+	return CSP_ERR_NONE;
+}
```
yashi: nitpick: no newline at the end.

yashi (changes requested):

src/bindings/python/pycsp.c
sigmundklaa: From what I can tell this only checks that `sizeof(uint16_t)==SIZEOF_UINT16`, no? If `uint16_t` is defined it is already guaranteed to be 16 bits, but it is not guaranteed to be the same as  `unsigned short`.

I guess you could add a compile time assertion that `sizeof(unsigned short)==sizeof(uint16_t)`, though that would make the Python bindings unusable on a system with "unusual" sizes of the primitive types, such as `sizeof(unsigned int) == sizeof(uint16_t)`. It is probably unlikely that such a system will run Python anyways though, so maybe thats the best option.

src/csp_port.c
sigmundklaa: I will add that, but the file has been formatted with clang-format using the `.clang-format` file in the repository, so if this is a formatting rule it really should be added to `.clang-format`.

src/bindings/python/pycsp.c
yashi: > From what I can tell this only checks that `sizeof(uint16_t)==SIZEOF_UINT16`, no? If `

ah, right.  I ment to do `sizeof(unsigned short)==sizeof(uint16_t)` or `sizeof(unsigned short) == 2`.

> that would make the Python bindings unusable on a system with "unusual" sizes of the primitive types, such as `sizeof(unsigned int) == sizeof(uint16_t)`. It is probably unlikely that such a system will run Python anyways though, so maybe thats the best option.

https://github.com/python/cpython/blob/main/configure.ac#L3005

I doubt CPython support `sizeof(unsigned int) == 2`.  We can start small with a limited set of supported architectures and expand later.

sigmundklaa: Ok. Will change.

src/csp_port.c
sigmundklaa: I'm not sure I understand, where is `port->socket` checked to see if the port is active? From what I can tell all other functions use exclusively the state to determine if the port is active?

Or do you mean that the port must be closed first, and then only after that the queue can be flushed?

src/csp_port.c
yashi: That's true! Let me handle it in #412.

Note that it's added in [Clang 16](https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#insertnewlineateof).

src/csp_port.c
yashi: `csp_port_get_socket()`
https://github.com/libcsp/libcsp/blob/develop/src/csp_port.c#L56-L75

yashi: OK. it's not socket but it's `state` you need set first.

wscript
sigmundklaa: Undone. Also uses of 'n' (Py_ssize_t) has been replaced with 'k' (unsigned long)

examples/csp_server_client.py
sigmundklaa: Changed to use `.format` instead

src/csp_port.c
sigmundklaa: I don't know why the newline is not appearing here. I have a newline in my local repository, with a clean working tree and the branch is up to date with origin.

src/bindings/python/pycsp.c
sigmundklaa: I have removed this. I have not yet added compile-time checks though, as I can't figure out how to do it in waf. For the time being CPython are already doing these checks, but they should probably be added to CSP aswell in case CPython changes. One option could be static assertions in `pycsp.c`, but with C99 this would mean quite bad error messages.

src/bindings/python/pycsp.c
```diff
@@ -833,6 +833,7 @@ static PyObject * pycsp_cmp_clock_get(PyObject * self, PyObject * args) {
 						 be32toh(msg.clock.tv_nsec));
 }
 
+#if CSP_HAVE_LIBZMQ
```
yashi, line 835: without libzmq3-dev on my Debian system, `#ifdef`ing in `pycsp.c` doesn't seem to be enough.
```
24/50] Compiling C object examples/zmqproxy.p/zmqproxy.c.o
FAILED: examples/zmqproxy.p/zmqproxy.c.o 
sccache cc -Iexamples/zmqproxy.p -Iexamples -I../examples -I. -I.. -I../include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=gnu11 -Os -g -pthread -Wshadow -Wcast-align -Wwrite-strings -Wno-unused-parameter -MD -MQ examples/zmqproxy.p/zmqproxy.c.o -MF examples/zmqproxy.p/zmqproxy.c.o.d -o examples/zmqproxy.p/zmqproxy.c.o -c ../examples/zmqproxy.c
../examples/zmqproxy.c:3:10: fatal error: zmq.h: No such file or directory
    3 | #include <zmq.h>
      |          ^~~~~~~
compilation terminated.
```

yashi (approved):

src/bindings/python/pycsp.c
sigmundklaa: What build system are you using?

yashi: All three gives you some error.

**waf** (configure time error)
```
Checking for 'libzmq'                    : not found 
The configuration failed
(complete log in /home/yashi/src/libcsp/build/config.log)
Traceback (most recent call last):
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 71, in <module>
    main(args.build_system, rest)
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 57, in main
    build_with_waf()
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 52, in build_with_waf
    subprocess.check_call(['./waf', 'distclean', 'configure', 'build'] + options)
  File "/usr/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['./waf', 'distclean', 'configure', 'build', '--with-os=posix', '--enable-rdp', '--enable-promisc', '--enable-crc32', '--enable-hmac', '--enable-dedup', '--enable-python3-bindings', '--enable-can-socketcan', '--with-driver-usart=linux', '--enable-if-zmqhub', '--enable-examples']' returned non-zero exit status 1.
```

**cmake**
```
[43/48] Building C object examples/CMakeFiles/zmqproxy.dir/zmqproxy.c.o
FAILED: examples/CMakeFiles/zmqproxy.dir/zmqproxy.c.o 
/usr/lib/ccache/cc  -I/home/yashi/src/libcsp/include -I/home/yashi/src/libcsp/build -Os -DNDEBUG -MD -MT examples/CMakeFiles/zmqproxy.dir/zmqproxy.c.o -MF examples/CMakeFiles/zmqproxy.dir/zmqproxy.c.o.d -o examples/CMakeFiles/zmqproxy.dir/zmqproxy.c.o -c /home/yashi/src/libcsp/examples/zmqproxy.c
/home/yashi/src/libcsp/examples/zmqproxy.c:3:10: fatal error: zmq.h: No such file or directory
    3 | #include <zmq.h>
      |          ^~~~~~~
compilation terminated.
[44/48] Building C object CMakeFiles/csp.dir/src/csp_service_handler.c.o
ninja: build stopped: subcommand failed.
Traceback (most recent call last):
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 71, in <module>
    main(args.build_system, rest)
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 59, in main
    build_with_cmake()
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 32, in build_with_cmake
    subprocess.check_call(cmake_compile + targets)
  File "/usr/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ninja', '-C', 'build', 'examples/csp_server_client', 'examples/csp_arch', 'examples/zmqproxy']' returned non-zero exit status 1.
```

**meson**
```
Run-time dependency libzmq found: NO (tried pkgconfig and cmake)
Has header "sys/socket.h" : YES 
Has header "arpa/inet.h" : YES 
Configuring csp_autoconfig.h using configuration
Build targets in project: 4

Found ninja-1.11.1 at /usr/bin/ninja
ninja: Entering directory `build'
[41/50] Compiling C object examples/zmqproxy.p/zmqproxy.c.o
FAILED: examples/zmqproxy.p/zmqproxy.c.o 
sccache cc -Iexamples/zmqproxy.p -Iexamples -I../examples -I. -I.. -I../include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=gnu11 -Os -g -pthread -Wshadow -Wcast-align -Wwrite-strings -Wno-unused-parameter -MD -MQ examples/zmqproxy.p/zmqproxy.c.o -MF examples/zmqproxy.p/zmqproxy.c.o.d -o examples/zmqproxy.p/zmqproxy.c.o -c ../examples/zmqproxy.c
../examples/zmqproxy.c:3:10: fatal error: zmq.h: No such file or directory
    3 | #include <zmq.h>
      |          ^~~~~~~
compilation terminated.
[45/50] Compiling C object libcsp.so.p/src_csp_service_handler.c.o
ninja: build stopped: subcommand failed.
Traceback (most recent call last):
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 71, in <module>
    main(args.build_system, rest)
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 61, in main
    build_with_meson()
  File "/home/yashi/src/libcsp/./examples/buildall.py", line 20, in build_with_meson
    subprocess.check_call(meson_compile + targets)
  File "/usr/lib/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ninja', '-C', 'build', 'examples/csp_server_client', 'examples/csp_arch', 'examples/zmqproxy']' returned non-zero exit status 1.
```

yashi: hmmm looking at it closer, it's not even `pycsp.c`'s problem.

sigmundklaa: Yeah, its the buildall script that includes ZMQ without an option for disabling it. Could add an option to the buildall script to enable/disable zmq.

sigmundklaa: But if you want to test it now, building regularly with CMake should work with the `-DCSP_ENABLE_PYTHON3_BINDINGS=ON` option

yashi: Never mind, I'll fix it with #413.

BTW, how did you test your #ifdef's ?

yashi:
Would you mind if I reorder your commits?  Because, for obvious reason, at the commit "examples: Enable Python3 bindings in waf buildall", a build fails.  This might be a problem when you bisect.  I'm not really a guy like "every commit must be buildable", but this one is easy enough to prevent a future pitfall.

src/bindings/python/pycsp.c
sigmundklaa: I had a seperate project that used CSP that had undefined symbols before adding the checks and therefore wouldn't compile (I did not use ZMQ in that project). Adding the checks fixed it.

sigmundklaa:
> Would you mind if I reorder your commits? Because, for obvious reason, at the commit "examples: Enable Python3 bindings in waf buildall", a build fails. This might be a problem when you bisect. I'm not really a guy like "every commit must be buildable", but this one is easy enough to prevent a future pitfall.

I don't mind, go for it.

yashi:
@sigmundklaa Thank you for your contribution!

PR #418, opened 2023-07-28
Add reference to documentation in README

SiboVG:
As a first-time user, it took me a short time to find documentation of the library. I think it'd be good to reference the documentation in the README.


yashi:
@SiboVG Thanks!

PR #420, opened 2023-07-29
End of file newlines

yashi:
End non-empty files with a newline.  This is a requirement from C standards.  This PR also has a commit adding a newline to non-source code as well.


Buzzzz:
Looks fabulous. Bonus points for removing other white-space at EOL. :-)

Issue #423, opened 2023-08-01
MacOSX support dropped but not documented


dmopalmer:
MacOSX and Windows support have been dropped as of 2.0, according to the `CHANGELOG`.
```
- removed: Windows Support (moved to contrib, maintainer needed)
- removed: Mac OSX Support (moved to contrib, maintainer needed)
```
However the `README.md` still indicates that they are supported.

Is there a build procedure that can compile the targets in contrib (with the caveat that the code is unmaintained, and maybe not supporting all three of `waf`, `meson` and `cmake`)?

Is it possible to checkout the 1.6 version (`git checkout  libcsp-1`) and compile it for the Mac?  I did this but the python library wasn't generated.

Should the `README.md` be changed to indicate (footnoted?) that these targets will not be available until a maintainer is found?

From my particular project, an ad-hoc specialized python implementation of CSP is the selfish-least-resistance path, but that incurs technical debt, doesn't test interoperability, and doesn't support the ecosystem.

yashi:
Hi,

The code for OSX is moved to contrib at a5256340b6d8c1fcfd344e8ebcafea32d7971a8e. Everything left in the `develop` and `libcsp-2-0` branches is something we forgot to remove.

Theoretically, what you need to do is to move code under contrib to the original location and update the API to match the current one. Unfortunately, we don't have any instructions to do that.

We are happy to take PRs to bring Mac support back to the current branch, but I'll merge a cleanup PR which removes Mac OS bits from the develop if no one is willing to step up.

dmopalmer:
I am working on this.  The major consequences are:

-  `<csp/csp_endian.h>` is back but it is just an include of the standard `<endian.h>`, except: 
    - On Mac: include `<machine/endian.h>` and `#define` the `htobe16()` etc. functions to alias macos equivalents.
- `# include <stdlib.h>` to define `malloc()` instead of the non-standard `<malloc.h>`
    - This should probably be done with or without the Mac, unless it affects another target   
- `__attribute__((section(".noinit")))` will need to be `#defined` with a slight change (or drop?) on the Mac
- `pthread_queue_t` needs to be made to work on the Mac.

These should have no as-compiled effect on the other targets, but a slight uglification of the code. Also, I do not promise to be a permanent maintainer.  

Is this enough to keep Mac on the future roadmap of libcsp, or do you want to strip it out as unmaintained?

yashi:
[...]

PR #432, opened 2023-08-04
Add macros for attributes

yashi:
It is needed to abstract `.noinit` section attributes for non-elf plaform.  This came up in #423 .

While doing so, I've added other macros for attributes we are using.


johandc (approved):
I like this. Nice and easy.
One thing i would like to change in the future is to move csp_macro.h to src/ folder, to not expose it to the "public" interface. But i could say that for several of the csp/include/ headers. So this is approved and getting merged now.

PR #434, opened 2023-08-05
Initialize CSP address

yashi:
In CSP v2.0, CSP address has been moved to per interfaace. But, we still need the `csp_conf` to set the CSP address on the interface because we don't have APIs to set an address for some interfaces.

With this PR, 
- socketcan interface takes the address from `csp_conf`.
- `csp_server_client.c` sets the address given with  the command line option `-a` to `csp_conf` so that it can work with CANbus as expected

Since we haven't released v2.0 yet, we can, instead of this PR, fix interface APIs.

For reference, the address field on `csp_conf` is removed in 4dadf8ff917402.

### DIFF
--- a/examples/csp_server_client.c
+++ b/examples/csp_server_client.c
@@ -191,6 +191,9 @@ int main(int argc, char * argv[]) {
 
     csp_print("Initialising CSP");
 
+    /* Set local CSP Address thorough the csp_conf */
+    csp_conf.address = address;
+
     /* Init CSP */
     csp_init();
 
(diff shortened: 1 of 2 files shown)


yashi:
This should be re-tested after #436

johandc:
This is solved by passing down the address now

yashi:
@johandc, do you mean "this _should_ be solved by passing down the address now"? Just like 0d4dac28a5bf75390bdd1ef26e1c109094e0a032? The commit is only for `develop` branch.  And this PR is about v2.0.

johandc:
Ahh i didnt see it was a backport. In that case fine. But perhaps we should sync up some of the 2.0 stuff anyways. I think its a bad idea to release a 2.0 with inconsistent addressing between interfaces

PR #435, opened 2023-08-05
Remove MacOS leftover from v2.0

yashi:
This is releated with #424 and #423. 
@dmopalmer seems to be working on MacOS port.  But I'm sure he is working on `develop` branch.  We can remove MacOS from v2.0.


yashi:
This should be re-tested after #436 is merged.

PR #446, opened 2023-09-05
Add install target in CMake

### LINKED ISSUE #444: Add install target in CMake
as noted in https://github.com/libcsp/libcsp/issues/251#issuecomment-1684316429, CMake doesn't have install target.  We thought nobody install libcsp but a few [use](https://github.com/libcsp/libcsp/issues/251#issuecomment-1682274165) [cases](https://github.com/spacecubics/scsat1-rpi/pull/7) arised.

It would be nice to have them back.

KantaTamura:
Added a target to enable CMake to install both libraries and header files.

Fixes #444

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,3 +82,11 @@ if(${CSP_ENABLE_PYTHON3_BINDINGS})
 endif()
 
 configure_file(csp_autoconfig.h.in include/csp/autoconfig.h)
+
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Zephyr")
+  install(TARGETS csp LIBRARY COMPONENT runtime)
+  install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/;${CMAKE_CURRENT_SOURCE_DIR}/include/;
+    TYPE INCLUDE
+    FILES_MATCHING PATTERN "*.h*"
+  )
+endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -82,3 +82,12 @@ if(${CSP_ENABLE_PYTHON3_BINDINGS})
 endif()
 
 configure_file(csp_autoconfig.h.in include/csp/autoconfig.h)
+
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Zephyr")
+  install(TARGETS csp LIBRARY COMPONENT runtime)
+  install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include/;${CMAKE_CURRENT_SOURCE_DIR}/include/;
+    TYPE INCLUDE
+    COMPONENT development
```
yashi: Please remove `COMPONET` form the header files installation.  Without it, we can install 
- `.so`: wtih `--component runtime`
- `.so` and header files: _without_ `--component` (full installation)

> [During a full installation all components are installed unless marked with EXCLUDE_FROM_ALL.](https://cmake.org/cmake/help/latest/command/install.html#introduction)

The doc also need to be updated accordingly.

yashi (changes requested):
Thanks for the PR.  it LGTM except that we don't need `COMPONENT` for header files.  Please fix it.

PR #449, opened 2023-09-10
include: Move private csp_macro.h under src/

### LINKED ISSUE #448: Move csp_macro.h to src/
One thing i would like to change in the future is to move csp_macro.h to src/ folder, to not expose it to the "public" interface. But i could say that for several of the csp/include/ headers. So this is approved and getting merged now.

_Originally posted by @johandc in https://github.com/libcsp/libcsp/pull/432#pullrequestreview-1618606240_

yashi:
`csp_macro.h` is not a public's API.  Move it under `src/`.

This fixes #448.


yashi:
@johandc, what do you think about `csp/csp_cmp.h`?  It uses packed attributes but if it pulls `mcros.h` in, `macros.h` must be public.

1. Does `if_stats` in `struct csp_cmp_message` have to be packed? I dont see a reason in f4486ca90697677417790e5510928d0855463043
2. If so, are you ok using `__atrribute__((__packed__))` for this case?

yashi:
ah, `csp_cmp_handler()` and others use the size of the struct as the data length on wire by `CMP_SIZE()` macro.

yashi:
So, for this PR, I'm using `__attribute__` in the public headers.  This should work until we add any other compiler, which doesn't support `__attribute__`.

johandc:
This is good. We should consider moving stuff from csp_cmp.h into csp_cmp.c instead. And also move some of the services to indivudal files to break it up a little.

PR #456, opened 2023-09-27
Rename CSP_HAVE_RTABLE to CSP_USE_RTABLE

yashi:
Each `CSP_HAVE_*` is used to indicate that we have a dependency. RTABLE, static routing table, is instead a feature we can choose at compile time.  For that, we use the prefix `CSP_USE_`.


johandc:
You are right :)

PR #464, opened 2023-10-03
cmake: Define CSP_HAVE_LIBZMQ and CSP_HAVE_LIBSOCKETCAN

### LINKED ISSUE #463: Can't use ZMQ interface on csp_server_client with cmake
On Ubuntu 22.04, building with `buildall.py --build-system cmake`, `CMakeLists.txt` doesn't set `CSP_HAVE_LIBZMQ` so `csp_server_client` isn't being built with a ZMQ interface, which means it can't be used with the `zmqproxy` that **is** being built. If I update `CMakeLists.txt` to set `CSP_HAVE_LIBZMQ=1`, I'm getting an error about an undefined reference to `csp_zmqhub_init`... but it's not clear to me why `csp_zmqhub_init` isn't available when I can see that we're making it into the `if(LIBZMQ_FOUND)` block of `src/interfaces/CMakeLists.txt`. Appreciate you helping me understand. 🙏 

```
[46/49] Linking C executable examples/csp_server_client
FAILED: examples/csp_server_client 
: && /usr/bin/cc -Os -DNDEBUG  examples/CMakeFiles/csp_server_client.dir/csp_server_client.c.o examples/CMakeFiles/csp_server_client.dir/csp_server_client_posix.c.o -o examples/csp_server_client  -Wl,-rpath,/src/build  libcsp.so && :
/usr/bin/ld: examples/CMakeFiles/csp_server_client.dir/csp_server_client.c.o: in function `main':
csp_server_client.c:(.text.startup+0x146): undefined reference to `csp_zmqhub_init'
collect2: error: ld returned 1 exit status
```

yashi:
CMake tries to find libzmq and libsocketcan and compile `interfaces/csp_if_zmqhub.c` and `drivers/can/can_socketcan.c` but it doesn't define `CSP_HAVE_LIBZMQ` and `CSP_HAVE_LIBSOCKETCAN` respectively even when `pkg-config` detects them.

When these defines are not defined, `csp_if_zmqhub.c` effectively becomes an empty file, and `can_socketcan.c` doesn't use any socketcan APIs.  Thus, the final libcsp.so isn't linked against the found libraries.

Define `CSP_HAVE_LIBZMQ` and `CSP_HAVE_LIBSOCKETCAN` in `csp_autoconfig.h` so that compile time condition works across the source tree.

This fixes #463.


danieldavidson (approved):
Thank you for the thorough explanation.

PR #465, opened 2023-10-03
doc: Build it in the dir user specified

yashi:
Do not use a fixed directory, "build-docs", for build. Instead, use the directory user specified.


yashi:
@IlievIliya92 WDYT?

IlievIliya92:
> @IlievIliya92 WDYT?

I think this is a better way for organizing the  build. Thanks!

PR #467, opened 2023-10-10
Fix example code support for ZMQHUB interface

danieldavidson:
Issues:
The server thread is only started if `server_address == 255` and the client thread is only started if `server_address != 255`. This makes it impossible to run the client and specify that it should connect to a server with address 255 (for example `-z localhost -a 1 -r 255`). While `-r 255` is the example was intended to configure the client to connect to the server, it instead started the server thread.

The address of the zmqhub interface was always set to 0. With the `zmqproxy` running, and the client and server started in separate processes, the server didn't recognize messages destined to it, and the client didn't recognize messages from itself or to the server.

Solutions:
Remove the checks for `server_address == 255` and `server_address != 255`, and replace them with new parameters the directly specify if the server thread or the client thread should be started.

Pass the local CSP address (set with `-a`) to `csp_zmqhub_init`. `csp_zmqhub_init_w_name_endpoints_rxfilter` sets the address of the interface.

### DIFF
--- a/doc/example.md
+++ b/doc/example.md
@@ -6,12 +6,12 @@ simple server/client setup, where the client sends a request to the
 server and receives a reply. The code can be compiled to an executable
 using `./examples/buildall.py`.
 
-The example supports the drivers and interfaces in CSP:
+The example supports these drivers and interfaces in CSP:
 
   - ZMQHUB: `-z <host name|ip>`
 
     Requires no extra hardware, as it uses standard network. The
-    zmqproxy will need to be started.
+    `zmqproxy` will need to be started.
 
   - CAN: `-c <can device>`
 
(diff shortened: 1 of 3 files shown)


examples/csp_server_client.c
```diff
@@ -168,7 +172,7 @@ int main(int argc, char * argv[]) {
     const char * rtable = NULL;
 #endif
     int opt;
-    while ((opt = getopt(argc, argv, "a:r:c:k:z:tR:h")) != -1) {
+    while ((opt = getopt(argc, argv, "a:r:c:k:z:tR:sR:lR:h")) != -1) {
```
yashi, line 175: Why do you specify 'R'?

examples/csp_server_client.c
```diff
@@ -150,7 +152,9 @@ static void print_usage(void)
 			  " -R <rtable>      set routing table\n"
 #endif
 			  " -t               enable test mode\n"
-			  " -h               print help\n");
+			  " -h               print help\n"
+			  " -s               enable server mode\n"
+			  " -l               enable client mode\n");
```
yashi, line 157: Let's sync with the order of the `optstring`.

examples/csp_server_client.c
```diff
@@ -283,13 +293,13 @@ int main(int argc, char * argv[]) {
...
     /* Start server thread */
-    if ((server_address == 255) ||  /* no server address specified, I must be server */
+    if ((server_mode) ||  /* server mode specified, I must be server */
         (default_iface == NULL)) {  /* no interfaces specified -> run server & client via loopback */
         server_start();
     }
 
     /* Start client thread */
-    if ((server_address != 255) ||  /* server address specified, I must be client */
+    if ((client_mode) ||  /* client mode specified, I must be client */
```
yashi, line 302: Do you accept both `-s` and `-l`? because you have `client_mode` and `server_mode`, we have

|| client_mode | server_mode |
|----|----|----|
| loopback | false | false |
| client mode | true | false |
| server mode | false | true |
| undefined | true | true |

You must error out when undefined.

Or, you can add a new `--mode` which takes `loopback`, `server`, or `client`.  Just an idea.

doc/example.md
```diff
@@ -34,35 +34,92 @@ The example supports the drivers and interfaces in CSP:
 
 ## Running the example
 
-If the example is started without any interfaces, it will use the
+If the example is started without any interfaces or additional configuration, it will use the
 loopback interface for communication between client and server:
 
-    ubuntu-18:~/libcsp$ ./build/csp_server_client
-    1586816581.410181 Initialising CSP
+    root@b21b36a935b5:/libcsp# ./build/example/csp_server_client
```
yashi, line 40: Do you need to run with the root privilege?  I personally don't like running anything under root.  I just tried both loopback and zmq under a user and it seems work just fine.

yashi, line 40: Also can we drop the host name? I assume `b21b36a935b5` is a Docker container ID but it's not obvious for users who want to try libcsp.

doc/example.md
```diff
@@ -34,35 +34,92 @@ The example supports the drivers and interfaces in CSP:
 
 ## Running the example
 
-If the example is started without any interfaces, it will use the
+If the example is started without any interfaces or additional configuration, it will use the
```
yashi, line 37: what additonal configuration do you have in your mind?  `./build/examples/csp_server_client -l` has additional configraution of `-l` and `-l` only, but it will ping to address 0.

yashi (changes requested):
Hi, thank you for your PR!  I love the idea! 

I've read your changes and I feel like this PR should be split to 4 commits.

- Fix the current doc 
- Add mode flags (this one should add `-s` and `-l` and update doc accordingly)
- Fix zmq interface by setting given addr to drv->iface.addr
- Fix examples/csp_server_client.c for ZMQ

The first one is obvious.  Your commits has a few doc fixes unrelated to zmq nor modes.  This should be one commit.

And the second commit should be adding modes.  This can be adding additional option like `-s` and `-l`, or any other updated implementation.  And this change should work for all other interfaces.

zmq interface code needs a fix.  This is to set given addr to `iface.addr`.

and finally you need to fix `csp_server_client.c` for ZMQ, which has always pass `0` as the address regardless of `-a`.

Speaking of `-a` because we didn't have mode, we only had one address to set.  But if we add `-s` and `-l`, can we make `-s` to take the server address and `-l` to take the client address and server addrees it needs to connect.

Oh, if you are interested in, would you write GitHub action to run ZMQ in test mode? So that when ever we broke it again we can be notified.

Also don't forget to write "why" you did it the way you did it in the commit log.

Finally, this is not asking or requesting you anything.  but I always wanted to split `csp_server_client.c` to `csp_server.c` and `csp_client.c`.  so that we can just run two separate command independently without worrying about these issues.  If we need to test loopback, we can have `csp_loopback.c` to just that.

Thanks again for your PR and your patience reading this comment. :smiley:

danieldavidson:
Thanks for the feedback! All great points. I'll have those PRs up shortly.

yashi:
They look great! Thanks!

But it seems you forgot to copy & paste the commit messages to the actual commit.  They are only on github AFAICS.

yashi:
Done merging #468, #469, #470. Thank you!  I'm closing this but if you still need something from this PR please let me know.

PR #468, opened 2023-10-11
Update example doc to reflect expected output.

danieldavidson:
The example documentation doesn't reflect the location of `csp_server_client` after it's built, or the expected output.

Update the example documentation with:
* the correct location of `csp_server_client` after it's been built with `buildall.py`.
* the correct output of `csp_server_client`.

These changes should help make the project more accessible to first-time users.


yashi:
Thanks!

PR #469, opened 2023-10-11
Fix: zmq interface by setting given addr to drv->iface.addr

danieldavidson:
When we setup the zmqhub interface, we accept an address that **should** be assigned to the zmq interface, but that address is never set in the interface. This will cause the zmq interface to ignore messages destined for the host, because the destination address to the message won't match the destination address of the interface.

In `csp_zmqhub_init_w_name_endpoints_rxfilter`, when we're configuring the zmq interface in the zmq driver, also set the `iface.addr` with the value of the `addr` parameter.


yashi:
Thanks!

PR #470, opened 2023-10-11
Fix: Set address when initializing zmq interface in example code

danieldavidson:
In the `csp_server_client` example code, when we initialize the zmqhub interface, the interface address is hardcoded to 0, rather than being set with the address specified by the `-a` parameter. This will cause the zmq interface to ignore messages that should be received by the host, based on the configured local CSP address.

Update the parameters in the call to `csp_zmqhub_init` to provide the value of `address` rather than 0.


yashi:
Thanks!

PR #473, opened 2023-10-15
Example: Create server, client, and loopback modes

danieldavidson:
It's unclear when and how parameters `-a` and `-r` should be used. The logic of how the client and server threads are started prevents modifying the client and server's addresses.

Refactor the parameters and logic so that there are well defined server, client, and loopback modes. A server just needs a server addresses specified so that it knows how to listen for messages; a client needs a client address so that it knows how to listen and a server address so that it knows where to connect to; when neither client or server addresses are specified both server and client run on the loopback interface.

This change to the parameters should make it clearer to the user how to run the client and server in different configurations, and the logic for starting client and server threads matches the parameter usage.

### DIFF
--- a/examples/csp_server_client.c
+++ b/examples/csp_server_client.c
@@ -18,7 +18,8 @@ int client_start(void);
 #define MY_SERVER_PORT		10
 
 /* Commandline options */
-static uint8_t server_address = 255;
+static uint8_t server_address = 0;
+static uint8_t client_address = 0;
 
 /* test mode, used for verifying that host & client can exchange packets over the loopback interface */
 static bool test_mode = false;
@@ -79,7 +80,7 @@ void server(void) {
 /* Client task sending requests to server task */
 void client(void) {
 
-	csp_print("Client task started");
+	csp_print("Client task started\n");
 
 	unsigned int count = 'A';
 
@@ -137,8 +138,6 @@ void client(void) {
 static void print_usage(void)
 {
 	csp_print("Usage:\n"
-			  " -a <address>     local CSP address\n"
-			  " -r <address>     run client against server address\n"
 #if (CSP_HAVE_LIBSOCKETCAN)
 			  " -c <can-device>  add CAN device\n"
 #endif
@@ -150,13 +149,14 @@ static void print_usage(void)
 			  " -R <rtable>      set routing table\n"
 #endif
 			  " -t               enable test mode\n"
+			  " -s               enable server mode and set the server address; if used with -l, sets the address of the server that the client should connect to\n"
+			  " -l               enable client mode and set the client address\n"
 			  " -h               print help\n");
 }
 
 /* main - initialization of CSP and start of server/client tasks */
 int main(int argc, char * argv[]) {
 
-    uint8_t address = 0;
 #if (CSP_HAVE_LIBSOCKETCAN)
     const char * can_device = NULL;
 #endif
@@ -168,14 +168,8 @@ int main(int argc, char * argv[]) {
     const char * rtable = NULL;
 #endif
     int opt;
-    while ((opt = getopt(argc, argv, "a:r:c:k:z:tR:h")) != -1) {
+    while ((opt = getopt(argc, argv, "c:k:z:R:ts:l:h")) != -1) {
         switch (opt) {
-            case 'a':
-                address = atoi(optarg);
-                break;
-            case 'r':
-                server_address = atoi(optarg);
-                break;
 #if (CSP_HAVE_LIBSOCKETCAN)
             case 'c':
                 can_device = optarg;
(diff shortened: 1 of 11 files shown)


yashi:
Hi @danieldavidson, Thanks for the PR.  I'll review it in a few days. Meanwhile, would you please fold your commit message at 70 or so?  Long lines are quite hard to read with tools like `gitk`.

examples/csp_server_client.c
```diff
@@ -208,7 +208,27 @@ int main(int argc, char * argv[]) {
         }
     }
 
-    csp_print("Initialising CSP");
+    /* Determine local address */
+    uint8_t address = 0;
+    char mode = 'x';
```
yashi: Let's make `mode` an enum.

```c
enum MODE {
	MODE_LOOPBACK,
	MODE_SERVER,
	MODE_CLIENT,
};
```

Or, if we use "mode" for "test mode", we should come up with better word.

yashi:
This is my brain dump.

### To test with a server and a client

Let's assume we have two distinct commands, `csp-server` and `csp-client`.  We also assume that we are using `getopt_long(3)` instead of `getopt(3)`.  What do you do? Here's what I'd do:

```
$ csp-server --my-address 1
```

```
$ csp-client --my-address 2 --connect-to 1
```

If you want to run both on the loopback interface:

```
$ csp-server --my-address 0
```

```
$ csp-client --my-address 0 --connect-to 0
```

### Addresses

Loopback test is easy because both server and client are under its control.  We can assign any address to the looback interface, and run it against it.  So we can think loopback as a special case.

OTOH, the reason we need to specify where the client should access is that the client will not inherently know the server's location. So, client must know the address it connects to.

server can randomly select an address on an interface, but that'd be tedious to run a client.  So, server also needs it's own address.

The client can have a random local address. However, if the randomly chosen address happens to be the same as the server's, it won't work.

In summary, 
- client needs
  - a local address which can be anything but not server's.
  - a server address to connect to. 
- server
  - a local address

### Modes

The current `csp-server-client` was created in #147. Before that, we had a few examples for each interfaces.  From the PR, I assume we want standalone server-client code.

To distinguish server mode and client mode on a single binary, we have to tell which mode we want it to run:

```
$ csp-server-client --server-mode
$ csp-server-client --client-mode
$ csp-server-client --client-mode --server-mode    # if we want to run both?
```
or

```
$ csp-server-client --mode server
$ csp-server-client --mode client
$ csp-server-client --mode loopback
```

or we can create a symlink so that exec binary can check `argv[0]` to see how it's executed
```
$ ls -l csp_*
lrwxrwxrwx 1 yashi yashi    17 Oct 18 18:10 csp_client -> csp_server_client
lrwxrwxrwx 1 yashi yashi    17 Oct 18 18:10 csp_server -> csp_server_client
```

I'm not sure this one works for non Linux.

### Addresses again

Both client and server needs local address.  If this statement is true, the local address can be a required command line parameter.  However, we also need to run in loopback mode, which we don't need to specify the local address with.  Thus, the local address must be an optional argument unless we use 1) `argv[0]` trick, or 2) per mode `getopt()`.

An address in CSPv2 is not on a node (server or client), but it's on an interface.  Then, it might be logical to specify an address for given interface.

```
$ csp-server-client --kiss /dev/ttyS0 --kiss-address 4
$ csp-server-client --zmq localhost --zmq-address 4
$ csp-server-client --ethernet eth0 --ethernet-address  4
```

or, if we introduce new way to specify deivce with address `<device>:<addr>`

```
$ csp-server-client --kiss /dev/ttyS0:4
$ csp-server-client --zmq localhost:4
$ csp-server-client --ethernet eht0:4    # this can be invalid because Linux allows to have `:` in device name
```

### Conclusion

We need

- Mode we want the executable to run in
- Local address (for bot server and client)
- Server address to connect to (for client)

Open Question:
- Do we support mulitple interfaces with `csp-server-client`.

yashi:
I forgot to mention this: Examples should be simple enough and focussed to one topic for users to grasp how to use the API. I believe this is the most important aspect of it

danieldavidson:
I like the idea of separate client and server binaries; I think it cleans up the code and simplifies the logic. But if I understand the loopback interface properly, we'll never get a message between the client and server if they're running in separate processes. Are you okay with a third binary like `csp_server_client_loopback` that only exercises the loopback test? While `csp_server` and `csp_client` are used for configurable interfaces, running in separate processes?

yashi:
> But if I understand the loopback interface properly, we'll never get a message between the client and server if they're running in separate processes.

Yes, you are right. My bad. Because libcsp is userland network protocol, you can't really have two processes on a system sharing the same network.

> Are you okay with a third binary like csp_server_client_loopback

I think the reason to merge the code as `csp_server_client` was maintainability.  But for the first time users, separate binaries are simpler, IMHO.  @perwulff @johandc Do you guys have any comment on this?

PR #475, opened 2023-10-18
.github/workflows: Disable documentation generation for forks

yashi:
I believe that people do not want to generate and publish the libcsp documentation on forked repositories. Therefore, we will only run the generation if the current repository owner is 'libcsp'.

@IlievIliya92 WDYT?


IlievIliya92:
> I believe that people do not want to generate and publish the libcsp documentation on forked repositories. Therefore, we will only run the generation if the current repository owner is 'libcsp'.
> 

Yes, that is reasonable. Thanks for spotting that, really didn't think about this scenario.

PR #476, opened 2023-10-19
interfaces: Add csp_can_remove_interface()

yashi:
In conjunction with csp_can_add_interface(), this commit introduces csp_can_remove_interface(), which, as the name suggests, removes the specified interface from the list.

This one is needed to implement Zephyr CAN interface, which tries to be idempotence.

### DIFF
--- a/include/csp/interfaces/csp_if_can.h
+++ b/include/csp/interfaces/csp_if_can.h
@@ -184,6 +184,15 @@ typedef struct {
 int csp_can_add_interface(csp_iface_t * iface);
 
 /**
+ * Remove interface.
+ *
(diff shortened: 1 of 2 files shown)


johandc:
Why do we need a csp_can_remove_iface when we already have csp_iflist_remove?

I know we have a helper csp_can_add.... so for consistency it would look better. But these helpers are of varying quality across the interfaces as also discussed elsewhere, some of them use malloc and there are multiple versions with different arguments and such.

yashi:
It's just for consistancy.  I'm sure we need to decide how we layer drivers and interfaces APIs.  But until then, I wanted it.

PR #478, opened 2023-11-08
Fix example folder typo

SiboVG:
There was a small typo in the docs; instead of the `build/examples` directory, the `build/example` was written.


yashi:
Thanks!

PR #487, opened 2023-11-13
Add missing macro includes

Abestanis:
This was required for me to build for FreeRTOS. I think CI is not checking building for embedded FreeRTOS, just FreeRTOS on Linux, which is why this didn't show up.

### DIFF
--- a/src/arch/freertos/csp_clock.c
+++ b/src/arch/freertos/csp_clock.c
@@ -1,6 +1,7 @@
 
 #include <csp/csp_types.h>
 #include <csp/csp_hooks.h>
+#include "csp_macro.h"
 
 __weak void csp_clock_get_time(csp_timestamp_t * time) {
 	time->tv_sec = 0;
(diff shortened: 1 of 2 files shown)


yashi (approved):
Thanks!

yashi:
would you please update your git commit log just like comment in github? If you can please do so for other PRs.  Thanks.

Abestanis:
Just to clarify, you want me to put the description from the Pull request here on GitHub into the git commit description? I can definitely do that.

yashi:
Yes.

Having git commit log is extremely valuable.  In fact, if you write commit log and then create PR from it, GitHub automatically generates description from the commit log.

Oh, don't forget to fold the line at 70 or so.

Thanks.

Abestanis:
> Having git commit log is extremely valuable.

I totally agree, added it. I'm going to add it to the other PRs as well.

PR #488, opened 2023-11-13
Fix possible endless loop when routing

Abestanis:
There are two conditions in this loop where we can continue, but in these cases route is never changed so we were just looping endlessly.


Abestanis:
Probably best to review the diff with the `Hide whitespace` option.

yashi:
@edvardxyz Would you please take a look?

yashi (approved):
LGTM.

edvardxyz:
Nice catch and fix on that bug. Sorry for the slip-up on my end.

PR #497, opened 2023-12-27
Avoiding potential buffer overflow in csp_can1_rx

pr0me:
Hello,

I propose to add an additional condition to the overflow check in `csp_can1_rx`:

**Reasoning:**
For incoming packets of type `CFP_MORE`, the old condition only checks
```C
if ((packet->rx_count + dlc - offset) > packet->length)
```
assuming that `packet->length` holds a sane value.  

For `CFP_BEGIN` packets, this is checked in line 103: `if (packet->length > sizeof(packet->data))`  

But for invalid values in `packet->length` for `CFP_MORE` packets, e.g., due to corrupted / malicious packages or random bit flips during transmission, the subsequent memcpy would potentially run out of bounds:
```C
memcpy(&packet->data[packet->rx_count], data + offset, dlc - offset);
```

Hence the added check.

Please let me know if there are any questions or whether there was a misconception on my end and the additional check is not necessary but I felt like this made sense.

Best,
Lukas

### DIFF
--- a/src/interfaces/csp_if_can.c
+++ b/src/interfaces/csp_if_can.c
@@ -131,7 +131,7 @@ int csp_can1_rx(csp_iface_t * iface, uint32_t id, const uint8_t * data, uint8_t 
 			packet->remain--;
 
 			/* Check for overflow */
-			if ((packet->rx_count + dlc - offset) > packet->length) {
+			if ((packet->length > sizeof(packet->data)) || (packet->rx_count + dlc - offset) > packet->length) {
 				csp_dbg_can_errno = CSP_DBG_CAN_ERR_RX_OVF;
 				iface->frame++;
 				csp_can_pbuf_free(ifdata, packet, 1, task_woken);


yashi:
Hi @pr0me,

Thank you for your PR.  Sorry to be late to respond.

Would you mind clarifying "why" you want to check `packet->length`?  As you already know the length only comes with a packet with `CFP_BEGIN` flag set (`0`).  So when we copy it, we make sure we have enough space in our data buffer by checking it aginst the lenght of data buffer (`packet->data`).   This is the only time we get the length and update `packet->length`.  We don't get length with `CSP_MORE`.  But we keep using `packet->length`, which we made sure less than or equal to `sizeof(packet->data)`, unless we have some type of data corruption.

Here are a few reasons you've listed:

1. due to corrupted
2. malicious packages
3. random bit flips

For (1), I don't think your change fix it.  Since we are not sure where you have corrupted in memroy.  What if the address of the packet is corrupted?
For (2), I need more info.  What exactly are "malicious packages" in your mind?  Carefully crafted packets from some bad nodes?  In that case we have already checked when we copied.
For (3), I'm sure we have SEU problems all the time.  But what if we get SEU right _after_ we checked? SEU cases can't be fixed by your chagne neither IMO.

pr0me:
Hi @yashi,

Thanks for having a look at this.
Some clarifications regarding my intuition:

For (1) / (2): Here I had indeed a misconception. I was thinking that by sending only a `CFP_MORE` packet one could circumvent the bounds check and the specified `packet->length` is directly used from the packet. But I now see that this is not possible, as raw packet and internal representation `packet` (which we retrieve with `csp_can_pbuf_find`) are separated and no value from the raw packet would be interpreted as its length.  
Only question would be what happens on uninitialized `packet` structs (always allocated with 0s?) but I am not sure whether this edge case even is achievable, I don't want to take up too much of your time and would need to read deeper into the CFP standard myself.

For (3) / (1): certainly no perfect fix, but by checking again a couple of instructions before the actual `memcpy` we can minimize the risk (kinda time-of-check-time-of-use).

Ultimately, I thought it's a quick better-safe-than-sorry-check without much change needed
but I can now see why you feel that this is unnecessary.
Feel free to close (or merge) :)


Best,
Lukas

yashi:
@pr0me, Ok I'm closing it.

I don't think sprinkling bound checks here and there isn't a good way to prevent SEU on CPU registers or memory. If you worry about SEU, you better get an OBC with SEU prevention / protection built-in.

But, that shouldn't block adding protections to the protocol, though.

Anyway, thank you for checking the code.  Let us know if you find anything else.

PR #502, opened 2024-01-15
off-by-one in buffer overflow check in `csp_can2_rx`

pr0me:
Hi,

Inside `csp_can2_rx`, the `if` guard checking that the "frame input + dlc must not exceed the end of the packet data field" calculates the end of the `data` array as `&packet->data[sizeof(packet->data)]` which is already one byte beyond the allocated struct (the last field of an array is `[sizeof(curr_type) - 1]`).

Consequently, the calculation should be adjusted or `>=` should be used instead of `>`.
Otherwise, the buffer can still be overflown.

|  offset  |  addr of last byte  |   symbol  |
|----------|---------------------|-----------|
|    0     |         295         |  &packet  |
|   40     |         295         | packet->data | 
|  296     |         296         | &packet->data[sizeof(packet->data)] |
|  297     |         ...         | > &packet->data[sizeof(packet->data)] |
|  296     |         ...         | > &packet->data[sizeof(packet->data) - 1] |
|  296     |         ...         | >= &packet->data[sizeof(packet->data)] |

### DIFF
--- a/src/interfaces/csp_if_can.c
+++ b/src/interfaces/csp_if_can.c
@@ -343,7 +343,7 @@ int csp_can2_rx(csp_iface_t * iface, uint32_t id, const uint8_t * data, uint8_t 
 	}
 
 	/* Check for overflow. The frame input + dlc must not exceed the end of the packet data field */
-	if (&packet->frame_begin[packet->frame_length] + dlc > &packet->data[sizeof(packet->data)]) {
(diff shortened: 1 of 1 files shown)


yashi:
I've spent a few hours reviewing the code, and it appears that we have more issues than just the one off-by-one. Let me see what I can do.

pr0me:
closing this, will open a single PR incorporating multiple bug fixes to reduce noise

pr0me:
closing this, will provide a single PR fixing multiple bugs to reduce noise

PR #503, opened 2024-01-15
unchecked MTU size and change to `csp_buffer_get` API can lead to buffer overflow in SFP

pr0me:
In `csp_sfp_send_own_memcpy`, an MTU of more than `(CSP_BUFFER_SIZE - sizeof(sfp_header_t))` could lead to buffer overflows in subsequent calls to `csp_sfp_header_add` and the memcpy in l. 77.

This is partially due to the unchanged usage of `csp_buffer_get`:
```C
csp_packet_t * packet = csp_buffer_get(mtu + sizeof(*sfp_header));
[...]
sfp_header = csp_sfp_header_add(packet); // no check, because buffer was allocated with extra size.
```

The argument to `csp_buffer_get` is actually unused and all `packet->data` buffers have a fixed size of `CSP_BUFFER_SIZE` (defaulting to 256).

As above comment (`"// no check, [...]"`) and not up to date documentation (mtu.md: `"so the recommended buffer allocation size is MAX MTU + 16 bytes. In this case the max MTU would be 256."`) add to the opaqueness of this issue, I decided to suggest a patch including clarifications in the documentation.

### DIFF
--- a/doc/mtu.md
+++ b/doc/mtu.md
@@ -6,16 +6,16 @@ There are two things limiting the MTU of CSP.
 2.  The link layer protocol.
 
 So let’s assume that you have made a protocol called KISS with a MTU
-of 256. The 256 is the total amount of data that you can put into the
+of 128. The 128 is the total amount of data that you can put into the
 CSP-packet. However, you need to take the overhead of the link layer
 into account. Typically this could consist of a length field and/or a
 start/stop flag. So the actual frame size on the link layer would for
-example be 256 bytes of data + 2 bytes sync flag + 2 bytes length field.
+example be 128 bytes of data + 2 bytes sync flag + 2 bytes length field.
 
-This requires a buffer allocation of at lest 256 + 2 + 2. However, the
+This requires a buffer allocation of at least 128 + 2 + 2. However, the
 CSP packet itself has some reserved bytes in the beginning of the packet
 (which you can see in csp.h) - so the recommended buffer allocation size
-is MAX MTU + 16 bytes. In this case the max MTU would be 256.
+is MAX MTU + 16 bytes < CSP_BUFFER_SIZE. In this case the max MTU would be 240.
 
 If you try to pass data which is longer than the MTU, the chance is that
 you will also make a buffer overflow in the CSP buffer pool. However,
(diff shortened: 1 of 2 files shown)


yashi:
Hi @pr0me,

Sorry for late reply.  Took me way too long to get back to this. Anyways,

`csp_packet_t` on `develop` branch is basically:

```c
typedef struct csp_packet_s {

	struct {
		uint16_t rx_count;          /*< Received bytes */
		uint16_t remain;            /*< Remaining packets */
		uint32_t cfpid;             /*< Connection CFP identification number */
		uint32_t last_used;         /*< Timestamp in ms for last use of buffer */
		uint8_t * frame_begin;
		uint16_t frame_length;
	};

	uint16_t length;			/*< Data length */
	csp_id_t id;				/*< CSP id (unpacked version CPU readable) */
	struct csp_packet_s * next; /*< Used for lists / queues of packets */

	uint8_t header[CSP_PACKET_PADDING_BYTES];
	uint8_t data[CSP_BUFFER_SIZE];

} csp_packet_t;
```

`CSP_BUFFER_SIZE` is used sololy for `data`, and the buffer pool is:


```c
typedef struct csp_skbf_s {
	unsigned int refcount;
	void * skbf_addr;
	csp_packet_t skbf_data;
} csp_skbf_t;

static csp_skbf_t csp_buffer_pool[CSP_BUFFER_COUNT]  __noinit;
```

Compare to v1.6:

```c
typedef struct csp_skbf_s {
	unsigned int refcount;
	void * skbf_addr;
	char skbf_data[]; // -> csp_packet_t
} csp_skbf_t;

static char * csp_buffer_pool;

int csp_buffer_init(void) {

	// calculate total size and ensure correct alignment (int *) for buffers
	const unsigned int skbfsize = CSP_BUFFER_ALIGN * ((sizeof(csp_skbf_t) + csp_buffer_size() + (CSP_BUFFER_ALIGN - 1)) / CSP_BUFFER_ALIGN);

	csp_buffer_pool = csp_malloc(csp_conf.buffers * skbfsize);
		:
}
```

So doc about the buffer pool in `mtu.md` is out of date, IMHO.

Also, why did you change the size from 256 to 128?

johandc:
Fixed in 4ab4934

PR #505, opened 2024-01-24
drivers/can: Remove CAN_FILTER_DATA

sasataku:
In Zephyr, with PR below, `CAN_FILTER_DATA` and `CAN_FILTER_RTR` have been removed.

  - https://github.com/zephyrproject-rtos/zephyr/pull/67127

Originally, `CAN_FILTER_DATA` was specified explicitly to filter CAN Data Frames, but in reality, in the Zephyr driver, it is possible to filter CAN Data Frames without specifying `CAN_FILTER_DATA`. Therefore, in this commit, `CAN_FILTER_DATA` is removed.


yashi (approved):
LGTM.

PR #506, opened 2024-01-29
Add reproducible builds

### LINKED ISSUE #499: Storing the date and time in the code make a build non reproducible
Hello,

I observed that that, when buildign the library, the compile date and time are stored in the generated binary.  Due to this, I cannot assure that a given version of a software version using your library is reproducible since this is the only difference observable in my binary.    
For information, this happen in the `do_cmp_ident()` function of the `src/csp_service_handler.c` file.

Do you plan to do something to allow full reproducibility of the project?    
Ideas could be to add a build flag to give a defined date, use the GIT sha instead (and an hardcoded number if built from outside a GIT repository), add a real version number tracking.

For now I just patched it with an hardcoded value but I am probably not the only person interested in having a reproducible build.

Thanks!

yashi:
This PR adds support for [reproducible builds](https://reproducible-builds.org/) to libcsp.

This fixes #499.


doc/INSTALL.md
```diff
@@ -85,3 +85,52 @@ use the following command:
...
+./waf configure --enable-reproducible-builds
+```
+
+### Meson
+
+```shell
+meson setup builddir . -Denable_reproducible_builds=true
+```
+
+### Camek
```
ElouanPetereau:
```suggestion
### Cmake
```

Also it is pretty hard to see the difference between the level 2 and 3 titles, what do you think about using a level 4 instead?

doc/INSTALL.md
```diff
@@ -85,3 +85,52 @@ use the following command:
...
+
+```shell
+cmake -S . -B builddir -DCSP_REPRODUCIBLE_BUILDS=1
+```
+
+Note: By default, CMake embeds the build directory in the binaries,
+resulting in non-deterministic builds. To address this, use
+`CSP_REPRODUCIBLE_BUILDS=ON` or `CMAKE_SKIP_RPATH=ON` as follows:
```
ElouanPetereau:
```suggestion
resulting in non-deterministic builds. To address this, use
`CMAKE_BUILD_RPATH_USE_ORIGIN=ON` or `CMAKE_SKIP_RPATH=ON` as follows:
```

I think you meant this no?

doc/INSTALL.md
```diff
@@ -85,3 +85,52 @@ use the following command:
...
+### Meson
+
+```shell
+meson setup builddir . -Denable_reproducible_builds=true
+```
+
+### Camek
+
+```shell
+cmake -S . -B builddir -DCSP_REPRODUCIBLE_BUILDS=1
```
ElouanPetereau: I don't think this is required since it will not work. The bellow text is enought imo.

If you prefer to keep it, I think it shoudl at least be coherent with the rest of the documentation. I am not sure whether  you want to recommand `ninja` or not but if you want to have the same as the `Using CMake` header, I think it should be:
```suggestion
cmake -Gninja -B builddir -DCSP_REPRODUCIBLE_BUILDS=ON
```

doc/INSTALL.md
```diff
@@ -85,3 +85,52 @@ use the following command:
...
+```
+
+Note: By default, CMake embeds the build directory in the binaries,
+resulting in non-deterministic builds. To address this, use
+`CSP_REPRODUCIBLE_BUILDS=ON` or `CMAKE_SKIP_RPATH=ON` as follows:
+
+```shell
+cmake -S . -B builddir -DCSP_REPRODUCIBLE_BUILDS=ON -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON
```
ElouanPetereau:
```suggestion
cmake -Gninja -B builddir -DCSP_REPRODUCIBLE_BUILDS=ON -DCMAKE_BUILD_RPATH_USE_ORIGIN=ON
```

 I think it should be coherent with the rest of the documentation and since the `-S .` is the default when not precised it can be removed.

doc/INSTALL.md
yashi: I really don't care what users use.  As long as it works for them, I'm happy.

But I agree that the doc should be consistent. So I'll use `-GNinja` for all.  Thanks for your suggestion.

doc/INSTALL.md
yashi: Done fixing it to "CMake", which seems to be the one the official uses.

Regarding to the heading size.  Data should represent what it is.  And, UI should be addressed by CSS from the theme we are using.

yashi:
@ElouanPetereau Thank you for the comments! I really appreciate it.

I think I addressed everything @ElouanPetereau pointed out.  I'll merge it if no other issue comes up.

PR #508, opened 2024-01-29
.github: Add dependabot

yashi:
Dependabot seems to check GitHub Actions.  Activate it.


yashi:
@johandc It seems this repo doesn't enable dependabot version updates.  Would you mind checking it?

PR #509, opened 2024-01-30
csp_rdp: Define _DEFAULT_SOURCE for rand_r()

yashi:
rand_r() is a function from POSIX 2001 (and deprecated in 2008).  If you compile with `-std=c11` or without GNU extension (`__STRICT_ANSI__`), `rand_r()` is not declared.

With `_DEFAULT_SOURCE`, we can have POSIX functions as well as functions from BSD and SVID.

### DIFF
--- a/src/csp_rdp.c
+++ b/src/csp_rdp.c
@@ -1,3 +1,5 @@
+#define _DEFAULT_SOURCE
+
 /*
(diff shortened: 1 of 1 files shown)


yashi:
This depends on an underlying OS, HAL, or environment.  So it might be better to set in the build system, which should know what to compile against.

PR #510, opened 2024-01-31
drivers, interfaces: Fixes for multiple security-relevant issues

pr0me:
This PR introduces patches to multiple potentially security-relevant bugs
after a code review.

Issues:
- csp_if_can: a buffer overflow due to an off-by-one in an overflow check in the CAN interface
- can_socketcan: potential buffer overflows in the CAN interface due to unchecked size field in the raw incoming CAN frame
- usart_linux: using a pointer returned by malloc without checking for NULL
- usart_linux: memory leaks due to unfreed pointers and unclosed file descriptors in possible paths in the USART driver, fixed by early failing
- csp_if_udp: destroying (freeing) resources allocated within a pthread `attribute` after using them as defined by POSIX

This PR was developed in collaboration with @yashi

### DIFF
--- a/src/drivers/usart/usart_linux.c
+++ b/src/drivers/usart/usart_linux.c
@@ -65,6 +65,10 @@ int csp_usart_write(csp_usart_fd_t fd, const void * data, size_t data_length) {
 }
 
 int csp_usart_open(const csp_usart_conf_t * conf, csp_usart_callback_t rx_callback, void * user_data, csp_usart_fd_t * return_fd) {
+	if (rx_callback == NULL && return_fd == NULL) {
+		csp_print("%s: No rx_callback function pointer or return_fd pointer provided\n", __FUNCTION__);
+		return CSP_ERR_INVAL;
+	}
 
 	int brate = 0;
 	switch (conf->baudrate) {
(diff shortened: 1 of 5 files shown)


yashi:
- I have addressed Clang complaing about double parenthathese on `if` condition.
- Added some commit log to a few commits.  (more to come).

@pr0me, would you mind taking a look at the changes in `"drivers: usart: Return early when no callback or fd ptrs are provided; This also prevents memory leaks later on"`? I believe the original code allowed users not to provide a callback. I am confident that the usage is valid: if a user can obtain a file descriptor from the function, they can write their own function to read from the `fd`.

So, the only invalid case is when a user doesn't provide both `rx_callback` and `return_fd`.

WDYT?

pr0me:
Agreed, the original code allowed that @yashi 
As it was intended, I made changes to reflect this usage and still fail early if neither is provided while fixing the memory leaks.

Let me know what you think

src/drivers/usart/usart_linux.c
```diff
@@ -193,10 +197,14 @@ int csp_usart_open(const csp_usart_conf_t * conf, csp_usart_callback_t rx_callba
...
 		}
+	} else {
+		free(ctx);
 	}
 
 	if (return_fd) {
 		*return_fd = fd;
+	} else {
+		close(fd);
 	}
```
yashi: Because we check `rx_callback` and `return_fd` at the very beginning of the function, we can now do something like:

```c
if (rx_callback) {
    ctx = calloc();
    ctx->rx_callback = rx_callback;
    ctx->fd = fd;
}

if (return_fd) {
    return_fd = fd;
}

return CS_ERROR_NONE;
```

Once we know we have either or both of `rx_callback` and `return_fd`, we are sure that `fd` won't be leaked. (Except for, of course, error cases)

IOW, we don't want to close the file when `rx_callback` expect to be open.

yashi (changes requested):

src/drivers/usart/usart_linux.c
pr0me: oh that's right, I've overlooked that rx_callback is using the open fd.
I've updated the commit accordingly

yashi:
@pr0me OK. I've updated all commit messages, and added one change I've found while reviewing them.  Please check them again.  If you are OK with it, I can merge this PR.

Hope the commit messages make sense.

pr0me:
Changes and commit messages are looking great, especially the off-by-one error one, thanks for the effort @yashi!

Let's ship it

PR #516, opened 2024-02-19
doc: Add Git Commit Guidelines

yashi:
Introduce a simple guidelines for Git commits to ensure clear and effective communication in the development process. The guidelines cover:

- Ensuring code adheres to established coding standards.
- Avoiding the mixing of multiple logical changes in one commit.
- Prohibiting the introduction of new trailing white spaces.

PR #517, opened 2024-02-20
Remove duplicate definition of csp_hex_dump()

dimitrovand:
csp_hex_dump() is already defined for all possible values of CSP_ENABLE_CSP_PRINT in csp_hex_dump.c

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,7 +5,6 @@ target_sources(csp PRIVATE
   csp_crc32.c
   csp_debug.c
   csp_dedup.c
-  csp_hex_dump.c
   csp_id.c
   csp_iflist.c
   csp_init.c
@@ -24,6 +23,10 @@ target_sources(csp PRIVATE
 
 if (CSP_HAVE_STDIO)
   target_sources(csp PRIVATE csp_rtable_stdio.c)
+endif()
+
+if (CSP_ENABLE_CSP_PRINT)
+  target_sources(csp PRIVATE csp_hex_dump.c)
 endif()
 
 add_subdirectory(arch)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,7 +10,6 @@ target_sources(csp PRIVATE
   csp_init.c
   csp_io.c
   csp_port.c
-  csp_promisc.c
   csp_qfifo.c
   csp_route.c
   csp_rtable_cidr.c
@@ -27,6 +26,10 @@ endif()
 
 if (CSP_ENABLE_CSP_PRINT)
   target_sources(csp PRIVATE csp_hex_dump.c)
+endif()
+
+if (CSP_USE_PROMISC)
+  target_sources(csp PRIVATE csp_promisc.c)
 endif()
 
 add_subdirectory(arch)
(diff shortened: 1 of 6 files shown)


yashi:
Hi @dimitrovand,

Thanks for the PR.  Please write your comment as a git commit log.  Not here in GitHub.  GitHub will take your git commit log and puts in the description area if your PR only has one commit.  Please fix your commit and push it again.  Thanks.

yashi:
would you please fold the body so that length is less than 72 or so?

Please see the example at https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines

yashi:
Thanks for comment fixes.

Now onto the techinical part.  I'd say that `csp_hex_dump.c` should not be compiled when `CSP_ENABLE_CSP_PRINT == 0`.  However, we want to keep inline version of both funcitons just to keep the symbols for applications.

WDYT?

dimitrovand:
No worries.

On first glance, I can't see any major/critical drawbacks in using any of these two solutions. One is adding to the build process, while the other - source code itself. I did the change led by the fact `csp_hex_dump.c` was already compiled unconditionally by all supported build systems, as it is done for other options (such as `RDP` and `PROMISC`). 
On a second pass over the source code, there seems to be no unified way of handling such scenarios. I guess this is purely a design decision, but as a library user - one would rather not like to deal with a complex build process (especially when having to recreate the existing one for a different build system), in my opinion.

yashi:
@dimitrovand In your comment, you mentioned that 'one would rather not like to deal with a complex build process.' Could you please provide more details or examples of what you consider a complex build process in this context?

Just to clarify, the conditional compilation we're discussing involves adding or removing the file.  And we already do that for other files.

Thanks.

dimitrovand:
Hello @yashi ,
By complex build process, I mean complex CMakeLists. You're right that the current implementation of it is not complex, and it's not the 1st time. I can easily redo it as suggested, but would you like this done also in all places where another approach has been used?

yashi:
> I can easily redo it as suggested, but would you like this done also in all places where another approach has been used?

Would you please share which are in your mind?  If they are the same type of problem, I'd love them to be fixed.  But want to make sure we are on the same page.  I don't want to waste you precious time otherwise.

dimitrovand:
They are not the exact same problem, but `csp_promisc.c` `csp_rdp.c` `csp_rdp_queue.c` are also being compiled even when their corresponding enable/disable option is set to 0 (same as `csp_hex_dump.c`).

yashi:
@dimitrovand Thanks. They can also be compiled conditionally.

yashi:
@dimitrovand Please fill in commit message.  Please see #516 for more info.

yashi (changes requested):
Thanks!

It's a nitpick but please do spell check on commit message.  Something like this should do:

  > When CSP_ENABLE_CSP_PRINT == 0, we can skip the compilation of csp_hex_dump.c due to the fact that there is already an inline implementation of csp_hex_dump() in csp.h. Due to the same reason, we can also remove the empty version of csp_hex_dump() implementation from csp_hex_dump.c.

I'm doing build test now but everything else LGTM.

yashi (approved):
LGTM.  Thanks!

PR #525, opened 2024-02-26
Convert global variables to static

dimitrovand:
'bif_a' and 'bif_b' are currently set as
global variables. They have been modified
by the commit e39c97cfa6adf77249f0494414cc3bb5e803e701. This commit is changing them to 'static'
since users can easily access them by using
the 'extern' keyword.

If we want access to them, in future (or now),
it might be a better choice to add a dedicated
function for gaining access to those variables.

### DIFF
--- a/src/csp_bridge.c
+++ b/src/csp_bridge.c
@@ -6,8 +6,8 @@
 #include "csp_dedup.h"
 #include <csp/arch/csp_time.h>
 
-csp_iface_t * bif_a;
-csp_iface_t * bif_b;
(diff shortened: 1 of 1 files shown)


yashi:
Nice catch.  I think these two are mistakes.

How about the following line for your git commit summary? I think it conveys more information without sacrificing the number of characters

```
csp_bridge: Make bif_a and bif_b static
```

yashi (approved):
LGTM.  Thanks!

PR #528, opened 2024-02-26
Rename csp_crc32_verify() to csp_crc32_verify_and_strip()

yashi:
This PR is basically renaming `csp_crc32_verify()` to `csp_crc32_verify_and_strip()`

### DIFF
--- a/doc/api/csp_crc32_h.rst
+++ b/doc/api/csp_crc32_h.rst
@@ -7,5 +7,6 @@ Interface Functions
 -------------------
 
 .. autocfunction:: csp_crc32.h::csp_crc32_append
+.. autocfunction:: csp_crc32.h::csp_crc32_verify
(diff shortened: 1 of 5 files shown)


yashi:
We also have `csp_hmac_verify()`.

dimitrovand:
Hello @yashi, if this PR is still relevant, I see your point for renaming the function, but since its part of the public API some existing code might already depend on this specific behaviour. It might be better to keep `csp_crc32_verify` as it is and introduce a new function, such as `csp_crc32_verify_no_strip`, what do you think?
Otherwise, I would only suggest comparing the return value of `csp_crc32_verify`, called from `csp_crc32_verify_and_strip`, to `CSP_ERR_NONE` instead of `0`.

yashi:
> It might be better to keep `csp_crc32_verify` as it is and introduce a new function, such as `csp_crc32_verify_no_strip`

That's a good idea.

PR #532, opened 2024-02-28
csp_io: remove unused variable

dimitrovand:
This commit removes the unused variable
'csp_promisc_queue' from 'csp_io.c'.
The use of the 'extern' keyword suggests
the variable is declared somewhere as global,
but its not. The only variable with such name
is located in 'csp_promisc.c' and it is 'static'.

### DIFF
--- a/src/csp_io.c
+++ b/src/csp_io.c
@@ -21,11 +21,6 @@
 #include "csp_qfifo.h"
 #include "csp_rdp.h"
 
-
-#if (CSP_USE_PROMISC)
-extern csp_queue_handle_t csp_promisc_queue;
-#endif
-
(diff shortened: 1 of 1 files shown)


yashi:
Thanks!

It seems that when we moved the promiscuous routine to its own file, csp_promisc.c, the signature of csp_promisc_add() was changed to remove the second argument, csp_promisc_queue. The commit moved to using the new csp_promisc_add() by introducing csp_promisc.h but forgot to remove the variable.

ref) 
https://github.com/libcsp/libcsp/commit/30cbebd5e9b75dad121b045e4d0123e928b1e0c2#diff-ad1b800b866f3e27dd116e67f387ea0185349d5760dd8758cced890516b8ccc5

@dimitrovand would you mind updating the commit log?

yashi (approved):

Issue #526, opened 2024-02-26
csp_bridge: potential issue in csp_bridge_work()


dimitrovand:
In `csp_bridge_work(void)` the following piece of code decides which interface is to be used:
```
	csp_iface_t * destif;
	if (input.iface == bif_a) {
		destif = bif_b;
	} else {
		destif = bif_a;
	}
```
If `csp_bridge_set_interfaces()` hasn't been called by the user, or simply called with invalid parameters, before invoking `csp_bridge_work(void)` this might lead to `csp_send_direct_iface()` being called with an invalid `destif` parameter. `csp_send_direct_iface()` doesn't perform any checks on the validity of its inputs parameters and this could lead to a crash when `iface->nexthop` is invoked from within the function.

pr0me:
suggested a patch in #542

yashi:
@dimitrovand, reopen this if you still have any problem.  Thanks.

PR #542, opened 2024-03-15
Improve Error Handling in csp_bridge_work

pr0me:
This PR makes two improvements to error handling inside the `csp_bridge_work`
function:
1. Add a NULL-pointer guard before calling `csp_send_direct_iface`
    - `csp_send_direct_iface` assumes the `iface` argument to be non-null and will otherwise produce undefined behavior when accessing `iface->nexthop`
    - other call sites of `csp_send_direct_iface` check that the supplied pointer is non-null before calling it, we stick to this 
2.  Return CSP error codes
    - `csp_bridge_work` has multiple locations where it could return early due to various errors but there is no way of telling from the outside / the call site
    - we add appropriate error codes on early-returns

This PR addresses #526.

### DIFF
--- a/src/csp_bridge.c
+++ b/src/csp_bridge.c
@@ -23,6 +23,12 @@ __weak void csp_input_hook(csp_iface_t * iface, csp_packet_t * packet) {
 
 void csp_bridge_work(void) {
 
+	if ((bif_a == NULL) || (bif_b == NULL)) {
+		csp_print("Bridge interfaces are not setup yet. "
+				  "Make sure to call csp_bridge_set_interfaces()\n");
+		return;
+	}
+
 	/* Get next packet to route */
 	csp_qfifo_t input;
 	if (csp_qfifo_read(&input) != CSP_ERR_NONE) {
--- a/src/csp_bridge.c
+++ b/src/csp_bridge.c
@@ -32,15 +32,18 @@ void csp_bridge_work(void) {
 	/* Get next packet to route */
 	csp_qfifo_t input;
 	if (csp_qfifo_read(&input) != CSP_ERR_NONE) {
+		csp_print("Failed to receive packet from router input queue\n");
 		return;
 	}
 
 	csp_packet_t * packet = input.packet;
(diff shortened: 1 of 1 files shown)


yashi:
Hi @pr0me.  Thanks for the PR.

Would you please keep the function's return type as void for now? I personally believe many functions should return a value indicating success or failure, but @johandc prefers to keep functions simple and has changed many to void in version 2.0. Meanwhile, we agree to include debug print statements with `csp_print()` to inform users. That approach would work for tablesats, especially with debug print enabled.

If we decide to change `csp_bridge_work()` to return a value, we would also want `csp_send*()` to return a value as well.

Furthermore, if we consider NULL to be an error, we should return as soon as possible. This approach is more aligned with a concept known as [Design by Contract](https://en.wikipedia.org/wiki/Design_by_contract).

```diff
1 file changed, 6 insertions(+)
src/csp_bridge.c | 6 ++++++

modified   src/csp_bridge.c
@@ -23,6 +23,12 @@ __weak void csp_input_hook(csp_iface_t * iface, csp_packet_t * packet) {
 
 void csp_bridge_work(void) {
 
+	if ((bif_a == NULL) || (bif_b == NULL)) {
+		csp_print("Bridge interfaces are not setup yet. "
+				  "Make sure to call csp_bridge_set_interfaces()\n");
+		return;
+	}
+
 	/* Get next packet to route */
 	csp_qfifo_t input;
 	if (csp_qfifo_read(&input) != CSP_ERR_NONE) {

```

Also, please check the commit messages.  The subject line must be followed by a empy line, then message body.

pr0me:
Thanks for the feedback @yashi 
I replaced the return error codes with error messages
and put the null check in the beginning as suggested.

I have to say that I don't really see how simple int return codes increase function complexity.
At the same time, having multiple different early returns without any way for a caller to 
check on them is not really a best practice. But I guess that's a discussion for elsewhere.

yashi (approved):
LGTM.  Thanks!

PR #543, opened 2024-03-20
cmake: move cmake_minimum_required()

dimitrovand:
Moving cmake_minimum_required() at the top of
CMakeLists.txt is required dew to warnings coming
from CMake when building CSP under different
build environment (such as CYGWIN).

From cmake_minimum_required() docs:
'Call the cmake_minimum_required() command at the
beginning of the top-level CMakeLists.txt file even before calling the project() command.'


yashi (approved):
Thank you for fixing this. I've been too lazy to do this. appreciate it.

PR #546, opened 2024-03-22
csp_crc32_verify: Exclude CRC size from packet length calculation

### LINKED ISSUE #514: bug in csp_crc32_verify
For what looks like historical reasons, this function tries to run the CRC32 twice;

Once  "with headers" and again "without headers"  (hmm)

(I notice a comment  that CSP_21 will default to "with headers" (makes sense))

However, in the first try, with headers, it does this, taking frame start & whole frame length

	crc = csp_crc32_memory(packet->frame_begin, packet->frame_length);

But of course that's not right, it should be:

	crc = csp_crc32_memory(packet->frame_begin, packet->frame_length-sizeof(crc));

Later on it uses memcmp to compare the last 4 bytes with the CRC, but you included it in computing the CRC.

(sometimes, eg with CCSDS CRC16,  a CRC including CRC itself comes out with zero, but not with this one (I think CRC32-C))

Shall I push the fix, it's trivial?

moonlight83340:
When verifying the CRC in csp_crc32_verify(), it is incorrect to calculate the CRC for the entire packet, including the CRC itself. 
In csp_crc32_append(), we generate a CRC value for the entire packet and then append this value at the end.
This commit corrects the issue by subtracting the size of the CRC from the frame_length, thus fixing the end address calculation.
This fix aligns with the proposal made by clythersHackers in https://github.com/libcsp/libcsp/issues/514.
Fixes #514

### DIFF
--- a/src/csp_crc32.c
+++ b/src/csp_crc32.c
@@ -119,7 +119,7 @@ int csp_crc32_verify(csp_packet_t * packet) {
 
 	/* Calculate CRC32, convert to network byte order */
 	csp_id_prepend(packet);
-	crc = csp_crc32_memory(packet->frame_begin, packet->frame_length);
+	crc = csp_crc32_memory(packet->frame_begin, packet->frame_length - sizeof(crc));
 	crc = htobe32(crc);
 
 	/* Compare calculated checksum with packet header */


moonlight83340:
I have tested my modifications using the server/client example.

```bash
sudo ip link add dev vcan0 type vcan
sudo ip link set up vcan0
ip link show vcan0
./examples/csp_server -c vcan0 -a 2 &
./examples/csp_client -c vcan0 -a 3 -C 2
```

The issue was with the CRC header part, so I defined CSP_21 to append the CRC on the header and the data. 
Then, I used some debug prints to check if the code was working properly with and without the header.

For the debug print, I modified `csp_crc32_verify()` function :

```
int csp_crc32_verify(csp_packet_t * packet) {
	uint32_t crc;

	if (packet->length < sizeof(crc)) {
		return CSP_ERR_CRC32;
	}

	/* Calculate CRC32, convert to network byte order */
	csp_id_prepend(packet);
	crc = csp_crc32_memory(packet->frame_begin, packet->frame_length - sizeof(crc));
	crc = htobe32(crc);
	
	/* Compare calculated checksum with packet header */
	if (memcmp(&packet->data[packet->length] - sizeof(crc), &crc, sizeof(crc)) != 0) {
		printf("CRC Header Failed");
		
	}else{
		printf("CRC Header OK");
	}

	/* Strip CRC32 */
	packet->length -= sizeof(crc);
	return CSP_ERR_NONE;
}
```

I also modified the `csp_crc32_append()` function, to be sure I was on CSP_21 path.
```

int csp_crc32_append(csp_packet_t * packet) {

	uint32_t crc;

	if ((packet->length + sizeof(crc)) > sizeof(packet->data)) {
		return CSP_ERR_NOMEM;
	}

	/* Calculate CRC32, convert to network byte order */
#if CSP_21 // In CSP 2.1 we change to include header per default
		csp_id_prepend(packet);
		crc = csp_crc32_memory(packet->frame_begin, packet->frame_length);
		printf("CSP_21");
#else
		crc = csp_crc32_memory(packet->data, packet->length);
#endif

	/* Convert to network byte order */
	crc = htobe32(crc);

	/* Copy checksum to packet */
	memcpy(&packet->data[packet->length], &crc, sizeof(crc));
	packet->length += sizeof(crc);

	return CSP_ERR_NONE;
}
```

PR #548, opened 2024-04-04
driver: Forward pxTaskWoken in kiss_driver_rx function.

olmanqj:
Is there a reason the `pxTaskWoken` pointer was set as NULL within `kiss_driver_rx`?

I need this simple thing to use the kiss driver with interruption support with my micro-controller.


yashi:
I assume nobody ever needed?

`kiss_driver_rx()` was introduced in https://github.com/libcsp/libcsp/commit/03903bb69c1fb89e5a6c3fd4cc75eba31e85dda8#diff-81b464066faa0ac17557ba9595ffeb177fcf2b01568c3d327c662c75aba542bbR45-R50

But this was, I assume, derived from `usart_linux.c` https://github.com/libcsp/libcsp/commit/d5116b84947a6eedd51b57fd9a770796ec575868#diff-4e8cdf3731047ce422dcfb3b35837bb63665e68dba2dedf3034ba4d149325f80R222-R224

So Yes I think we can pass `pxTaskWoken` to `kiss_driver_rx()`.

@olmanqj Please check our git commit guideline at https://github.com/libcsp/libcsp/pull/516/files and update yours?

Thanks.

olmanqj:
Sorry for the dumb question: Shoudl I update the merge request message or create anew commit with the updated message?

moonlight83340:
> Sorry for the dumb question: Shoudl I update the merge request message or create anew commit with the updated message?

I think you should first clean up the git history, because you have a merge commit on your PR.
If you want to update your current fork from the upstream you should use rebase instead of merge.
`git rebase upstream/develop`

If you want to modify only the git message of your last commit you can use `git amend`.

I check your branch, so in your case you should do :
`git reset HEAD^ --hard` it will remove the merge commit you have done.
and then a 
`git rebase upstream/develop` it will rebase from upstream so your branch will be up to date.

Finally, you'll need to modify your commit message using git commit --amend."

Issue #547, opened 2024-03-27
Example on how to use I2C


curat97:
Hello, I am opening this issue to ask if there is any example on how to use the i2c interface. I am using the libcsp library for a project involving two raspberry pi zero. I was able to correctly send and receive data using the UDP and CAN interface, however I am having troubles configuring the i2c interface. 
I am using the function _csp_i2c_add_interface_ to add the i2c interface and the function _csp_sendto_ to send the data. When I execute my C program, I get a segmentation fault error in the function _csp_i2c_tx_. I attach here the code of the C program. I premise that I am not an expert in C programming, so I could make trivial mistakes, but I hope this issue could be useful to others trying to use the i2c interface. 

```c
#include <unistd.h>
#include <stdio.h>
#include <string.h>
#include <csp/csp.h>
#include <csp/csp_debug.h>
#include <stdlib.h>
#include <pthread.h>
#include <csp/interfaces/csp_if_i2c.h>
#include <csp/csp_crc32.h>

static unsigned int rpi_sent = 0;
uint32_t opts = CSP_O_CRC32;
uint8_t prio = CSP_PRIO_NORM;
uint16_t rpi_csp_Addr = 2000;
uint8_t rpi_csp_port = 12;	
uint16_t cubesat_I2C_csp_Addr = 4005;
uint8_t cubesat_I2C_csp_port = 20;

static int csp_pthread_create(void * (*routine)(void *)) {

	pthread_attr_t attributes;
	pthread_t handle;
	int ret;

	if (pthread_attr_init(&attributes) != 0) {
		return CSP_ERR_NOMEM;
	}
	/* no need to join with thread to free its resources */ 
	pthread_attr_setdetachstate(&attributes, PTHREAD_CREATE_DETACHED);

	ret = pthread_create(&handle, &attributes, routine, NULL);
	pthread_attr_destroy(&attributes);

	if (ret != 0) {
		return ret;
	}

	return CSP_ERR_NONE;
}

// Router task

static void * task_router(void * param) {
	while (1) {
		csp_route_work();
	}
	return NULL;
}

int router_start(void) {
	return csp_pthread_create(task_router);
}

// client task 

void client(void) {
    int j;
	for (j=0; j<5; j=j+1){

		csp_packet_t * packet = csp_buffer_get(100);		
		char exStr[] = "Hello world \r\n";
		memcpy(packet->data, exStr ,strlen(exStr)+1);
		packet->length = (strlen((char *) packet->data) +1);
[...]

PR #553, opened 2024-04-19
CHANGELOG: Remove I2C testing from TODO

yashi:
@curat97 has reported that the I2C interface works with libcsp 2.0 in #547. Therefore, it is being removed from the TODO list.


yashi:
as noted at https://github.com/libcsp/libcsp/issues/547#issuecomment-2066069870, it's not fully working yet. Closing this for now.

PR #569, opened 2024-05-02
python: Fix argument type for can init

moonlight83340:
Fix argument type on `pycsp_can_socketcan_init()` . 
Use `int16_t` instead of an `int`.

### DIFF
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -854,8 +854,8 @@ static PyObject * pycsp_can_socketcan_init(PyObject * self, PyObject * args) {
 	char * ifc;
 	int bitrate = 1000000;
 	int promisc = 0;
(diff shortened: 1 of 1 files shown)


src/bindings/python/pycsp.c
```diff
@@ -854,8 +854,8 @@ static PyObject * pycsp_can_socketcan_init(PyObject * self, PyObject * args) {
 	char * ifc;
 	int bitrate = 1000000;
 	int promisc = 0;
-	int addr = 0;
-	if (!PyArg_ParseTuple(args, "s|iii", &ifc, &addr, &bitrate, &promisc)) {
+	int16_t addr = 0;
```
yashi: Why `int16_t` instead of `uint16_t`?

yashi (approved):

PR #582, opened 2024-05-31
fix address not being set in example on libcsp-2-0

novakjak:
When running the example a client was never able
to connect to the server because the address was
never being set. I.e. pings would always return -1.

This issue is fixed in `develop` but not on the stable branch.

### DIFF
--- a/examples/csp_server_client.c
+++ b/examples/csp_server_client.c
@@ -31,7 +31,7 @@ void server(void) {
 
 	/* Create socket with no specific socket options, e.g. accepts CRC32, HMAC, etc. if enabled during compilation */
 	csp_socket_t sock = {0};
-    
+
 	/* Bind socket to all ports, e.g. all incoming connections will be handled here */
 	csp_bind(&sock, CSP_ANY);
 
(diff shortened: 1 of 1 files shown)


examples/csp_server_client.c
```diff
@@ -240,6 +240,7 @@ int main(int argc, char * argv[]) {
         }
     } else if (default_iface) {
         csp_rtable_set(0, 0, default_iface, CSP_NO_VIA_ADDRESS);
+        default_iface->addr = address;
```
yashi: If you set `->addr` here, it's not set when you specify rtable, no?

novakjak: Yes, you are right. I was so desperate to fix the issue that I didn't double check if my changes were making sense. Fix it now so it should be more like the example in `develop`.

yashi:
please squash the commits.  See https://github.com/libcsp/libcsp/pull/590/files

yashi:
I'm pretty sure that you need the same fix for socketcan and ZeroMQ. Have you tested them?

yashi (approved):
LGTM. Thanks.

Issue #491, opened 2023-11-16
RDP and CRC32 dropping/duplicating packets


alegnani:
As the description says, when using both RDP and CRC32, packets get dropped.
I tried using both RDP and CRC32 separately and they work fine by themselves.

The issue is encountered when sending packets, with a monotonically increasing counter as a message, over a connection that randomly corrupts bytes.
Could it be that the ACK gets send when receiving the packet before it gets dropped due to a wrong CRC?

yashi:
Care to share how to reproduce?

alegnani:
Sorry forgot to actually upload the code.
`csp_server.c` and `csp_client.c` are replacements for the files in the example folder and can be built with `python buildall.py`.
To run them I used: `./build/examples/csp_server -z localhost -a 2` and `./build/examples/csp_client -z localhost -a 1 -C 2`.
The proxy that introduces corruption can be run with `python lossy_zmq_proxy.py`.

I am observing that packets arrive duplicated at the server.
Additionally I have observed that it sometimes skips packets (receiving nr.4 and then nr.6), thus leading me to believe that the ACK is sent before checking the CRC, which then drops the packet.

[code.zip](https://github.com/libcsp/libcsp/files/13411773/code.zip)

yashi:
I can't even make it work with `CSP_O_NONE`.  Can you?

alegnani:
I uploaded the wrong file. It should be `CSP_O_RDP | CSP_O_CRC32`.

These are the tests I performed:
-  `CSP_O_RDP`, packet_loss = 0.1 and no corruption:
   - Sending increasing counter works
- `CSP_O_CRC32`, no packet_loss and corruption = 0.02:
   - Sending the same message over and over does not arrive corrupted.
   - If the CRC32 is wrong it gets dropped
- `CSP_O_CRC32 | CSP_O_RDP`, packet_loss = 0.1 and corruption = 0.02:
   - Sending increasing counter does not work
   - Sometimes a duplicated message arrives
   - Sometime a message gets skipped
   - Same thing happens with packet_loss = 0.0

yashi:
Is this due to ZMQ?

> A successful invocation of zmq_send() does not indicate that the message has been transmitted to the network, only that it has been queued on the socket and ØMQ has assumed responsibility for the message.

> When a ZMQ_PUB socket enters the mute state due to having reached the high water mark for a subscriber, then any messages that would be sent to the subscriber in question shall instead be dropped until the mute state ends.

[...]

PR #594, opened 2024-06-24
rdp: Plug packet buffer leak in csp_rdp_queue_flush()

moonlight83340:
A conn closing could delete packet from others conn, now we enqueue them again.

I was checking another [issue](https://github.com/libcsp/libcsp/issues/491) during my debug, I found out that when another connection was closing it would dequeue and free the packet from other connection. That shouldn't be the case because it could cause a packet loss on RDP.

### DIFF
--- a/src/csp_rdp_queue.c
+++ b/src/csp_rdp_queue.c
@@ -1,6 +1,7 @@
 #include <csp/arch/csp_queue.h>
 #include <csp/csp_types.h>
 #include <csp/csp.h>
+#include <csp/csp_debug.h>
 
 static csp_queue_handle_t tx_queue;
 static csp_static_queue_t tx_queue_static; /* Static storage for rx queue */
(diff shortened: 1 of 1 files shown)


yashi:
Good catch. Thanks.

Would you mind if I ask you to work on the commit log a bit?

moonlight83340:
> Good catch. Thanks.
> 
> Would you mind if I ask you to work on the commit log a bit?

No problem !

moonlight83340:
I add some comments and csp_debug for fail error.
I was thinking on rewriting the function...

```
static void __csp_rdp_queue_flush(csp_queue_handle_t * queue, csp_conn_t * conn) {
    csp_packet_t * packet;
    int queue_size = csp_queue_size(queue);
    while (queue_size--) {
		if (csp_queue_dequeue(queue, &packet, 0) == CSP_QUEUE_ERROR) {
			/* csp_queue_dequeue will fail if the queue is empty, so it's better to break here */
			break;
		}

		if (packet == NULL) {
			continue;
		}
		if ((conn == NULL) || (conn == packet->conn)) {
			csp_buffer_free(packet);
		} else {
			if (csp_queue_enqueue(queue, &packet, 0) != CSP_QUEUE_OK) {
                csp_print("csp_rdp_queue_flush : queue fail to enqueue, data lost !\n");
				/* Free the packet if re-enqueue fails */
				csp_buffer_free(packet);
			}
		}
	}
}

void csp_rdp_queue_flush(csp_conn_t * conn) {
	/* Flush conn data from TX queue */
	__csp_rdp_queue_flush(tx_queue, conn);
	/* Flush conn data from RX queue */
	__csp_rdp_queue_flush(rx_queue, conn);
}
```
I hope it make sense.

yashi:
this could be a root cause of #491.

yashi (approved):
LGTM. Thanks.

Issue #512, opened 2024-02-06
KISS for Zephyr


ignacio-gr:
I've got an application that is currently running zephyr on a MCU and needs CSP KISS over USART to work.
Any clue on how to implement it?

yashi:
We are working on it for our project https://github.com/spacecubics/scsat1-rpi.
Stay tuned.

If you can't wait for it, what you have to do is to create a new layer, `libcsp/src/drivers/usart/usart_zephyr.c`, which binds Zephyr UART driver and the libcsp kiss interface layer, `libcsp/src/interfaces/csp_if_kiss.c`

PR #599, opened 2024-07-02
add zephyr uart driver

KantaTamura:
This PR adds a Zephyr UART driver to libcsp.

Currently, the implementation uses the Polling API for tx and rx.

This resolves issue #512

### DIFF
--- a/contrib/zephyr/Kconfig
+++ b/contrib/zephyr/Kconfig
@@ -35,6 +35,13 @@ config CSP_USE_RTABLE
 	help
 	  This option enables to use the CSP static routing table
 
+config CSP_ATOMIC_MUTEX_TIMEOUT
+	int "Waiting period to lock the mutex (msec) in atomic"
+	depends on ARMV6_M_ARMV8_M_BASELINE
+	default 10
+	help
+	  Waiting period to lock the mutex in atomic.
+
 if CAN
 
 config CSP_HAVE_CAN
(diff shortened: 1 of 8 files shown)


contrib/zephyr/Kconfig
```diff
@@ -35,6 +35,42 @@ config CSP_USE_RTABLE
 	help
 	  This option enables to use the CSP static routing table
 
+config CSP_MUTEX_TIMEOUT
```
yashi, line 45: This Kconfig is used for both atomic operation and `csp_usart_lock()`.  These two should use two different Kconfigs.

yashi (changes requested):

contrib/zephyr/Kconfig
KantaTamura: Ok. Add `CONFIG_CSP_ATOMIC_MUTEX_TIMEOUT` for atomic operation.

yashi: And, this should depends on `CONFIG_ARMV6_M_ARMV8_M_BASELINE`, right?

KantaTamura: That's certainly true. 
Fixed `CONFIG_CSP_ATOMIC_MUTEX_TIMEOUT` to depend on `CONFIG_ARMV6_M_ARMV8_M_BASELINE`.

PR #600, opened 2024-07-03
zephyr: Remove LIBCSP_HAVE_STDIO

yashi:
LIBCSP_HAVE_STDIO is only used to control the inclusion of Zephyr logging macro. The actual CSP_HAVE_STDIO is default ON in the top level CMakeLists.txt.

Thus, the only function of this config is to make a build error for all Zephyr logging when not set.

Remove it.


yashi:
this was merged to #602

PR #602, opened 2024-07-03
hotfix: Fix Zephyr build failure

### LINKED ISSUE #601: Build error with Zephyr
Currently we have a bug in GitHub Actions for Zephyr build test, which causes all Zephyr related checks to fail. I'm preparing a hot fix for it.

yashi:
This PR has two fixes:

- Test against PR revision instead of `develop` revision
- Remove `LIBCSP_HAVE_STDIO`

The first was a bug in one of our actions, which takes `develop revision` for testing rather than the pull-requested revision. This makes tests against Zephyr all green even if PR has some bugs in it.  It was the case for #599.

The second is the actual bug we had in #599.

This fixes #601


yashi:
@KantaTamura Please review.

KantaTamura:
LGTM!

PR #616, opened 2024-07-13
github: build-test: Add Ubuntu 24.04

yashi:
Ubuntu 24.04 is the latest LTS release from Ubuntu. It's currently beta for GitHub Actions but it seems to work for libcsp.

Add ubuntu-24.04 for yet another supported platform.


yashi:
~We have an issue with Waf and Python 3.12 or later.~
~See https://gitlab.com/ita1024/waf/-/issues/2442~

Issue #618, opened 2024-07-14
Prepare v2.1


yashi:
We've tagged v2.0 last April but the branch `libcsp-2-0` has been branched since December 2021, which is more than 2.5 years ago. This means that `develop` branch has already accumulated 2.5 years of work. So I feel like I should tag v2.1.

I'd like to fix or include following changes before v2.1:

- [x] #633
- [x] #491 
- [x] #369
- [x] #597 
- [x] #583
- [x] #564
- [x] #503 
- [x] #864 
- [x] #898

If possible, I'd also want to include

- #586 (waiting for how to test)

If you have any issues or PRs that you would like to be included in v2.1, please let us know.

Thanks.

yashi:
I just run [ABI Compliance Checker](https://lvc.github.io/abi-compliance-checker/) on v2.0 and v2.1.  Here is the sumary.  Clearly, we are far from Semantic Versioning.


**Problem Summary**

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">


<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-right" />
</colgroup>
<tbody>
<tr>
<td class="org-left">&#xa0;</td>
<td class="org-left">Severity</td>
<td class="org-right">Count</td>
</tr>

<tr>
<td class="org-left">Added Symbols</td>
<td class="org-left">-</td>
<td class="org-right">16</td>
</tr>

<tr>
<td class="org-left">Removed Symbols</td>
<td class="org-left">High</td>
<td class="org-right">25</td>
</tr>

<tr>
<td class="org-left">Problems with Data Types</td>
<td class="org-left">High</td>
<td class="org-right">0</td>
</tr>

<tr>
<td class="org-left">&#xa0;</td>
<td class="org-left">Medium</td>
<td class="org-right">5</td>
</tr>

<tr>
<td class="org-left">&#xa0;</td>
<td class="org-left">Low</td>
<td class="org-right">16</td>
</tr>

<tr>
<td class="org-left">Problems with Symbols</td>
<td class="org-left">High</td>
<td class="org-right">12</td>
</tr>

<tr>
<td class="org-left">&#xa0;</td>
<td class="org-left">Medium</td>
<td class="org-right">11</td>
</tr>

<tr>
<td class="org-left">&#xa0;</td>
<td class="org-left">Low</td>
<td class="org-right">12</td>
</tr>

<tr>
<td class="org-left">Problems with Constants</td>
[...]

PR #631, opened 2024-08-08
github: workflow: Add ABI Compliance Check

moonlight83340:
I try to make a github workflow for ABI compliance check.
It's currently a WIP. I'm not used to make github workflow file so I wanted to share my first version 😅 
I saw it would be nice to do a github action from this issue https://github.com/libcsp/libcsp/issues/618

### DIFF
--- a/.github/workflows/abi-checker.yml
+++ b/.github/workflows/abi-checker.yml
@@ -1,0 +1,61 @@
+name: ABI Compliance Check
+on: [pull_request]
+jobs:
+  abi-check:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Setup packages on Linux
+        run: |
+          sudo apt-get update
+          sudo apt-get install libzmq3-dev libsocketcan-dev
+
+      - name: Setup build system packages on Linux
+        run: |
+          sudo apt-get install ninja-build cmake
+
+      - name: Setup ABI Compliance Checker
+        run: |
+          sudo apt-get install abi-compliance-checker abi-dumper
+
+      - name: Checkout merged-base
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+
+      - name: Build and Dump ABI for `merged-base` version
+        run: |
+          cmake -GNinja -B build -DCMAKE_BUILD_TYPE=Debug -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C build
+          abi-dumper build/libcsp.so -lver "merged-base" -o ../tmp/libcsp-merged-base.dump
+
+      - name: Checkout Current PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+
+      - name: Build and Dump ABI for Current PR
+        run: |
+          cmake -GNinja -B build -DCMAKE_BUILD_TYPE=Debug -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C build
+          abi-dumper build/libcsp.so -lver "PR" -o ../tmp/pr.dump
+
+      - name: Compare ABI
+        run: |
+          abi-compliance-checker -l libcsp -old ../tmp/libcsp-merged-base.dump -new ../tmp/pr.dump
+
+      - name: Check ABI Compatibility Result
+        if: failure()
+        run: echo "::warning::ABI compatibility issues detected. Please review the report."
+
+      - name: Upload ABI Report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: abi-compliance-report-${{ github.run_id }}
+          path: compat_reports
+          if-no-files-found: warn


.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,66 @@
...
+      matrix:
+        os:
+          - ubuntu-24.04
+        arch:
+          - posix
+        buildsystem:
+          - cmake
+        compiler:
+          - CC: gcc
+            CXX: g++
```
yashi: not needed. unless you want to build it with clang.

yashi (commented):
Thanks! Looks good to me.

.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,52 @@
...
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Setup packages on Linux
+        run: |
+          sudo apt-get update
+          sudo apt-get install libzmq3-dev libsocketcan-dev socat
```
yashi: Why do you need socat

.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Build and Dump ABI for Current Branch/PR
+        run: |
+          cmake -GNinja -B build -DCMAKE_BUILD_TYPE=Debug -DCSP_HAVE_LIBZMQ=1 -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C build
+          abi-dumper build/libcsp.so -lver 2.1 -o /home/runner/work/tmp/develop.dump
+
+      - name: Compare ABI
+        run: |
+          abi-compliance-checker -l libcsp -old /home/runner/work/tmp/libcsp-2-0.dump -new /home/runner/work/tmp/develop.dump
+        continue-on-error: true
```
yashi: What we need to know is that a PR introduce ABI change or not. So you have to compare against the merge base instead of the stable branch.

yashi: This means that we don't really need to run this on push, but need to check on all PRs.

yashi (changes requested):

.github/workflows/abi-checker.yml
moonlight83340: weird thing I get an error on checkout here but not on my own repository...
https://github.com/moonlight83340/libcsp/actions/runs/10313278326/job/28549781344?pr=1

.github/workflows/abi-checker.yml
moonlight83340: No need indeed.

.github/workflows/abi-checker.yml
yashi: where? I only see greens

image

moonlight83340: Nevermind I find out !

.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Build and Dump ABI for Current PR
+        run: |
+          cmake -GNinja -B build -DCMAKE_BUILD_TYPE=Debug -DCSP_HAVE_LIBZMQ=1 -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C build
+          abi-dumper build/libcsp.so -lver "PR" -o ./../tmp/pr.dump
+
+      - name: Compare ABI
+        run: |
+          abi-compliance-checker -l libcsp -old ./../tmp/libcsp-merged-base.dump -new ./../tmp/pr.dump
+        continue-on-error: true
```
yashi: Why do you need this? If the following upload need to work independent from `abi-comliance-chcker`'s exit code, use `if: always()`.  You can also use `if-no-files-found`. Check the doc: https://github.com/actions/upload-artifact

Also please make it indicate that incompatible is found.

yashi (changes requested):

.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,61 @@
...
+          sudo apt-get update
+          sudo apt-get install libzmq3-dev libsocketcan-dev
+
+      - name: Setup build system packages on Linux
+        run: |
+          sudo apt-get install ninja-build cmake
+
+      - name: Setup ABI Compliance Checker
+        run: |
+          sudo apt-get install -y abi-compliance-checker abi-dumper
```
yashi: Consistancy.  Remove `-y` to follow the same pattern from the above and the doc.

.github/workflows/abi-checker.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Checkout Current PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+
+      - name: Build and Dump ABI for Current PR
+        run: |
+          cmake -GNinja -B build -DCMAKE_BUILD_TYPE=Debug -DCSP_HAVE_LIBZMQ=1 -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C build
```
yashi: Why `DCSP_HAVE_LIBZMQ=1`? This is not user option

yashi (changes requested):

yashi:
Let's see how it goes.

PR #633, opened 2024-08-09
:boom: Activate CRC32 header protection

### LINKED ISSUE #46: CRC doesn't protect headers
Related to #45, CSP does not include the header when calculating the checksum. We cannot easily fix it, because CSP would then break compatability with itself.

### LINKED ISSUE #617: crc: Include header part for CRC calculation
As noted in [`csp_crc32.c`](https://github.com/libcsp/libcsp/blob/f21c1aa923303c02b9f788c7b6286bc8219c37cb/src/csp_crc32.c#L86-L100), we will include the header part of the CSP packet in the CRC calculation.

I am considering adding a compile-time option to fall back to the current behavior, but this is not yet decided.

This issue is just a heads-up notification.

yashi:
This enables header part inclusion of CRC32 calculation. This is a breaking change. Please test.

This addresses and actually fixes #46.
And this #617.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 libcsp 2.1, xx-yy-zzzz
 ----------------------
+- new: CRC32 calculation includes header when enabled. This means wire protocol incompatibility with v1.6 or earlier.
 
 libcsp 2.0, 19-04-2024
 ----------------------
--- a/examples/buildall.py
+++ b/examples/buildall.py
@@ -42,7 +42,6 @@ def build_with_waf():
         '--with-os=posix',
         '--enable-rdp',
         '--enable-promisc',
-        '--enable-crc32',
         '--enable-hmac',
         '--enable-dedup',
         '--enable-python3-bindings',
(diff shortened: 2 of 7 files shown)


yashi:
With so many v1.x instances out there, I believe we should maintain the v1.0 wire protocol in libcsp when it is configured at runtime to communicate with v1.x.

Any thoughts?

johandc:
We cannot implement this yet, because there was a bug up untill recently, that would prohibit the support. I belive we should keep the default to sending without header CRC for 2.1 as well, but implement it as optional runtime parameter to do or not. Then in 2.2 we change the default, but keep the optional runtime parameter for backwards capability.

johandc:
Actually one could argue, that adding header crc would be version 3 of the header... But maybe that is stretching it too far?

yashi:
> Actually one could argue, that adding header crc would be version 3 of the header... But maybe that is stretching it too far?

I think that's reasonable. I'm not sure how

yashi:
Let's not activate this for v2.1. I'm dropping it. We'll discuss how to proceed in another issue.

PR #638, opened 2024-08-14
csp_route: Fix return value comparison

dimitrovand:
The return values of `csp_route_security_check()`
and `csp_conn_enqueue_packet()` are compared
directly to 0. This commit updates the
comparison to use `CSP_ERR_NONE`, which both
functions return to indicate success.


yashi (approved):
LGTM. Thanks.

PR #639, opened 2024-08-14
csp_port: Fix return value of csp_bind_callback()

dimitrovand:
The `csp_bind_callback()` function returns debug
error codes. This commit updates the return values to use the official error codes from `csp_error.h`.

Also included some typo corrections.


yashi:
Please move the typo fixes for "Binding" to another commit.  Thanks.

PR #640, opened 2024-08-15
csp_route: Fix return value

moonlight83340:
This commit updates the return value to use CSP_ERR_NONE instead of 0 directly, to indicate success.


yashi:
Please rebase on top of the current tip.

PR #642, opened 2024-08-16
github: workflow: Add build and test python binding

moonlight83340:
Add build and test python binding.
This commit aim to help to found if there are any build error in python binding.

### DIFF
--- a/.github/workflows/build-test-python.yml
+++ b/.github/workflows/build-test-python.yml
@@ -1,0 +1,49 @@
+name: Build and test python
+on: [push, pull_request]
+jobs:
+  build-test-python:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04
+          - ubuntu-22.04
+          - ubuntu-20.04
+        arch:
+          - posix
+        buildsystem:
+          - meson
+          - cmake
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Setup packages on Linux
+        run: |
+          sudo apt-get update
+          sudo apt-get install libzmq3-dev libsocketcan-dev
+
+      - name: Setup build system packages on Linux
+        run: |
+          sudo apt-get install ninja-build ${{ matrix.buildsystem }}
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build libcsp examples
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-python.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build libcsp examples
+        run: python3 examples/buildall.py ${{ matrix.arch }} --build-system=${{ matrix.buildsystem }}
+
+      - name: Build libcsp with python binding with cmake
+        if: ${{ matrix.buildsystem == 'cmake' }}
+        run: |
+          cmake -GNinja -B buildBinding -DCSP_ENABLE_PYTHON3_BINDINGS=1 -DCSP_USE_RTABLE=1 -DCMAKE_POSITION_INDEPENDENT_CODE=ON && ninja -C buildBinding
```
yashi: Do we need `DCMAKE_POSITION_INDEPENDENT_CODE=`?

Also, can we just use `builddir` for `-B`?

yashi (changes requested):

.github/workflows/build-test-python.yml
moonlight83340: No we don't need that option, yes we can use `builddir `👍

yashi (approved):

.github/workflows/build-test-python.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Build libcsp with python binding with meson
+        if: ${{ matrix.buildsystem == 'meson' }}
+        run: |
+          meson setup builddir -Denable_python3_bindings=true -Duse_rtable=true && ninja -C builddir
+
+      - name: Run ZMQ Python binding Test
+        run: |
+          build/examples/zmqproxy &
+          LD_LIBRARY_PATH=builddir PYTHONPATH=builddir python3 examples/python_bindings_example_server.py &
```
yashi: Do we need `LD_LIBRARY_PATH`?

yashi (changes requested):

.github/workflows/build-test-python.yml
moonlight83340: No indeed. Removed !

yashi (approved):

PR #645, opened 2024-08-17
Cast all `printf("%p")` values to void pointers

parker-research:
Reason: gcc gives the following build warning:

```
format '%p' expects argument of type 'void *', but argument 3 has type 'csp_socket_t *' {aka 'struct csp_conn_s *'} [-Wformat=]
```

### DIFF
--- a/src/arch/freertos/csp_semaphore.c
+++ b/src/arch/freertos/csp_semaphore.c
@@ -55,7 +55,7 @@ int csp_bin_sem_remove(csp_bin_sem_handle_t * sem) {
 }
 
 int csp_bin_sem_wait(csp_bin_sem_handle_t * sem, uint32_t timeout) {
-	csp_log_lock("Wait: %p", sem);
+	csp_log_lock("Wait: %p", (void *)sem);
 	if (timeout != CSP_MAX_TIMEOUT) {
 		timeout = timeout / portTICK_RATE_MS;
 	}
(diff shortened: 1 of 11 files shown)


yashi:
Hi @parker-research,

Thank you for the PR. Please update the commit message according to [the guideline](https://libcsp.github.io/libcsp/git-commit.html#commit-message-guidelines).

Also, it might be a good idea to reference the C standard, or the draft thereof, N2310.

**7.21.6.1 The `fprintf` function**
> If any argument is not the correct type for the corresponding conversion specification, the behavior is undefined.

I have installed gitlint to `libcsp-1`. So please rebase on top of it.

Thanks,

parker-research:
All three of my recent PRs (#645, #646, #647) had the purpose of working toward building without errors with `-Wpedantic`. 

Should I combine them  into a single PR/commit, @yashi? Any advice?

yashi:
No need to combine them in one PR.  Separate PRs are fine. I need commit messsages in all though.

parker-research:
Sorry, I don't understand what's wrong with the commit message. As far as I can tell, it conforms to your guideline, so the issue might be that the guideline is missing a detail.

For contributions from occassional contributors, have you considered doing a GitHub "Squash and Merge" instead? It lets you pick/rewrite the commit message at merge time.

yashi:
If you go https://github.com/libcsp/libcsp/actions/runs/10434343582/job/28898021792 by clicking "Details", you see

> 8: B1 Line exceeds max length (116>80): "> If any argument is not the correct type for the corresponding conversion specification, the behavior is undefined."

So, your commit message has a long line, 116 chars, which should be less than 80.

parker-research:
Thanks! Fixed!

yashi:
@parker-research It seems we have a conflicts. Would you please fix it?

parker-research:
Fixed! Thanks

PR #663, opened 2024-08-18
type safety: Make data types explicit in bit modifications

6arms1leg:
Compiling for most targets, left bit shifts of "small" data types might cause integer promotion.  On a 32 bit system, for example, left-shifting the priority promotes it to a signed 32 bit integer.  The priority is a 2 bit value that is left-shifted to bit position 30, thereby overwriting the sign bit.  This causes some analyzers (e.g. GCC sanitizer) to issue an error.
The clean solution is to cast the variables to the proper lvalue data type of the assignment *before* left-shifting.

### DIFF
--- a/src/csp_id.c
+++ b/src/csp_id.c
@@ -40,12 +40,12 @@
 static void csp_id1_prepend(csp_packet_t * packet) {
 
 	/* Pack into 32-bit using host endian */
-	uint32_t id1 = ((packet->id.pri << CSP_ID1_PRIO_OFFSET) |
-					(packet->id.dst << CSP_ID1_DST_OFFSET) |
-					(packet->id.src << CSP_ID1_SRC_OFFSET) |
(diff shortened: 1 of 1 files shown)


yashi:
Thank you for the PR!

For curious readers, here is how you enable GCC Undefined Behavior Sanitizer (ubsan)

```c
#include <stdint.h>

int main()
{
	uint8_t a = 3;

	(void)(a << 30);

	return 0;
}
```
```
❯ gcc -fsanitize=undefined -Wall -Wextra -g a.c
❯ ./a.out
a.c:7:11: runtime error: left shift of 3 by 30 places cannot be represented in type 'int'
```

Or, if you want to enable for libcsp,

**CMake**
```
❯ cmake -B builddir -DCMAKE_C_FLAGS=-fsanitize=undefined 
```

**Meson**
```
❯ meson setup builddir -Db_sanitize=undefined
```

PR #670, opened 2024-08-22
github: workflow: Add CAN interface testing

moonlight83340:
Added steps to set up a virtual CAN (vcan0) interface in the CI environment. 
Included tests for the CAN interface using the `csp_server` and `csp_client` examples.
Ensured the CAN interface is correctly configured with `modprobe vcan`and `ip link add dev vcan0 type vcan`.
Implemented both server and client tests to validate CAN communication.

This change enables automated testing of the CAN interface within the CI pipeline, improving test coverage for CAN-related functionality.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -24,7 +24,8 @@ jobs:
         if: ${{ runner.os == 'Linux' }}
         run: |
           sudo apt-get update
-          sudo apt-get install libzmq3-dev libsocketcan-dev socat
+          sudo apt-get install libzmq3-dev libsocketcan-dev socat iproute2
+          sudo apt-get install linux-modules-extra-$(uname -r)
 
       - name: Setup build system packages on Linux
         if: ${{ runner.os == 'Linux' && matrix.buildsystem != 'waf' }}
@@ -81,3 +82,24 @@ jobs:
           ./build/examples/csp_client -z localhost -a 2 -C 1 -T 10 &
           ./build/examples/csp_server -z localhost -a 1 -t
           pkill zmqproxy
+
+      - name: Setup vcan0
+        run: |
+          sudo modprobe vcan
+          sudo ip link add dev vcan0 type vcan
+          sudo ip link set up vcan0
+          echo "Waiting for vcan0 to be up..."
+          while ! ip -br link show vcan0 | grep -q "UP"; do
+            sleep 0.1
+          done
+          echo "vcan0 is up"
+
+      - name: Run CAN Server Test
+        run: |
+          ./build/examples/csp_server -c vcan0 -a 1 -T 10 &
+          ./build/examples/csp_client -c vcan0 -a 2 -C 1 -t
+
+      - name: Run CAN Client Test
+        run: |
+          ./build/examples/csp_client -c vcan0 -a 2 -C 1 -T 10 &
+          ./build/examples/csp_server -c vcan0 -a 1 -t


.github/workflows/build-test.yml
```diff
@@ -81,3 +82,20 @@ jobs:
           ./build/examples/csp_client -z localhost -a 2 -C 1 -T 10 &
           ./build/examples/csp_server -z localhost -a 1 -t
           pkill zmqproxy
+
+      - name: Set vcan0
+        run: |
+          sudo modprobe vcan
+          sudo ip link add dev vcan0 type vcan
+          sudo ip link set up vcan0
+          ip -br link show vcan0
```
yashi: This does not wait until the interface `vcan0` is up.

yashi (changes requested):

.github/workflows/build-test.yml
moonlight83340: I think it should be okay but if we want to be really sure we can do something like that : 
```
  while ! ip -br link show vcan0 | grep -q "UP"; do
    echo "Waiting for vcan0 to be up..."
    sleep 1
  done
  echo "vcan0 is up"
```

So we don't wait if we don't need to. But in case vcan0 isn't up we will wait.

yashi: It is **_NOT_** OK. If it were OK, you wouldn't have added `sleep 1` in the first place. How can you say it's OK? And I'm sure I want to make sure it doesn't fail from time to time. I'm the one who needs to re-run the failed tests and fix or maintain them.

yashi: It is **_NOT_** OK. Making sure `vcan0` is up isn't about being "really sure"; it's a fundamental requirement for the system to function correctly. The use of `sleep 1` suggests that we already anticipate this being a common issue, so it's not about adding extra caution—it's about handling the normal, expected behavior of the system. We need to have a reliable way to ensure `vcan0` is up before proceeding because this is a basic necessity, not an optional step. As the one responsible for re-running failed tests and maintaining this, it's crucial to have this reliability built in.

moonlight83340: Sorry, I think we missunderstand each other. I answer right away. But the true was I was unsure if we have to wait some time for vcan0 to be up, I put sleep 1 in case of but it's clearly unreliable.
I have add a loop to wait and I will wait 0.1 s because, I don't think we need much time but in case we needed it, the loop will do the job. I hope the solution is correct.
I can surely try another solution.

.github/workflows/build-test.yml
```diff
@@ -24,7 +24,8 @@ jobs:
         if: ${{ runner.os == 'Linux' }}
         run: |
           sudo apt-get update
-          sudo apt-get install libzmq3-dev libsocketcan-dev socat
+          sudo apt-get install libzmq3-dev libsocketcan-dev socat iproute2 can-utils
```
yashi: What do you use `can-utils` for?

moonlight83340: Not used, thanks.

.github/workflows/build-test.yml
```diff
@@ -81,3 +82,24 @@ jobs:
           ./build/examples/csp_client -z localhost -a 2 -C 1 -T 10 &
           ./build/examples/csp_server -z localhost -a 1 -t
           pkill zmqproxy
+
+      - name: Set vcan0
```
yashi: ditto

yashi (changes requested):

yashi (approved):

PR #671, opened 2024-08-23
examples: python: Add CAN interface testing to python example

moonlight83340:
This PR aim to add the support of CAN interface testing in the python example.

- Add the command line option for can and CAN init on python server example.
- Modify the CAN interface init in client to show up to the user how to configure CAN interface in python example.
- Add CAN interface test in the python github workflow

### DIFF
--- a/.github/workflows/build-test-python.yml
+++ b/.github/workflows/build-test-python.yml
@@ -59,14 +59,14 @@ jobs:
       - name: Run ZMQ Python binding Test
         run: |
           build/examples/zmqproxy &
-          PYTHONPATH=builddir python3 examples/python_bindings_example_server.py -z localhost -a 3 &
-          PYTHONPATH=builddir python3 examples/python_bindings_example_client.py -z localhost -s 3 -a 2
+          PYTHONPATH=builddir python3 examples/python_bindings_example_server.py -z localhost -a 3 -R "0/0 ZMQHUB" &
+          PYTHONPATH=builddir python3 examples/python_bindings_example_client.py -z localhost -s 3 -a 2 -R "0/0 ZMQHUB" 
           pkill zmqproxy
 
       - name: Run KISS Python binding Test
         run: |
           socat -d -d -d pty,raw,echo=0,link=/tmp/pty1 pty,raw,echo=0,link=/tmp/pty2 &
           sleep 1
-          PYTHONPATH=builddir python3 examples/python_bindings_example_server.py -k /tmp/pty2 -a 1 &
-          PYTHONPATH=builddir python3 examples/python_bindings_example_client.py -k /tmp/pty1 -a 2 -s 1
+          PYTHONPATH=builddir python3 examples/python_bindings_example_server.py -k /tmp/pty2 -a 1 -R "0/0 KISS" &
+          PYTHONPATH=builddir python3 examples/python_bindings_example_client.py -k /tmp/pty1 -a 2 -s 1 -R "0/0 KISS"
           pkill socat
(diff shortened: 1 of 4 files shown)


examples/python_bindings_example_client.py
```diff
@@ -44,8 +44,9 @@ def get_options():
 
     if options.can:
         # add CAN interface
-        libcsp.can_socketcan_init(options.can)                  
-   
+        libcsp.can_socketcan_init(options.can, options.address, 1000000, 1)
```
yashi: You didn't say why you chose to use 1 Mbps and promiscuous mode

yashi (changes requested):

examples/python_bindings_example_client.py
moonlight83340: I wanted to test as much possible arguments.
I choose 1Mbs because it's the default value 😅 . ( And I'm quite not sure what I should test).

yashi: If it's the default, leave it alone.

.github/workflows/build-test-python.yml
```diff
@@ -70,3 +71,15 @@ jobs:
...
+
+      - name: Run CAN Python binding Test
+        run: |
+          sudo modprobe vcan
+          sudo ip link add dev vcan0 type vcan
+          sudo ip link set up vcan0
+          while ! ip -br link show vcan0 | grep -q "UP"; do
+            sleep 0.1
+          done
+          echo "vcan0 is up"
```
yashi, line 84: If somehow this failed, you are left alone without any message. Print a message _before_ waiting.

yashi: also, can you split `vcan` setup part from the test?

examples/python_bindings_example_client.py
```diff
@@ -44,8 +44,9 @@ def get_options():
...
+        libcsp.rtable_load("0/0 CAN")
```
yashi: do we need this?

examples/python_bindings_example_server.py
```diff
@@ -124,6 +125,11 @@ def csp_server():
         libcsp.kiss_init(options.kiss, options.address)
         libcsp.rtable_load("0/0 KISS")
 
+    if options.can:
+        # add CAN interface
+        libcsp.can_socketcan_init(options.can, options.address, 1000000, 1)
+        libcsp.rtable_load("0/0 CAN")
```
yashi: This mean that we always use rtable?

yashi (changes requested):

.github/workflows/build-test-python.yml
```diff
@@ -70,3 +71,19 @@ jobs:
           PYTHONPATH=builddir python3 examples/python_bindings_example_server.py -k /tmp/pty2 -a 1 &
           PYTHONPATH=builddir python3 examples/python_bindings_example_client.py -k /tmp/pty1 -a 2 -s 1
           pkill socat
+
+      - name: Set vcan0
```
yashi: Setup instead of Set

yashi (changes requested):

examples/python_bindings_example_server.py
moonlight83340: In fact, at first, I thought it was mandatory to use rtable because of how the Python example was written. 
However, after reading your comment, I checked and found that there's currently a bug in the Python binding; the code doesn't work if rtable isn't used (we have to use rtable_load).
I will create a PR after this one to fix that problem.

yashi (commented):
If you add two interfaces, what happens?

Please test by adding two of the following interfaces—KISS, CAN, and/or ZMQ—from Python, and let me know the result.

moonlight83340:
I will close for now. I think https://github.com/libcsp/libcsp/pull/801 would be helpful to continue that python testing.

PR #673, opened 2024-09-01
csp_rdp_queue: Fix undefined return value

dimitrovand:
When `__csp_rdp_queue_flush` encounters an empty
queue (i.e., `csp_queue_size(queue) == 0`),
the function would return an undefined value due
to the uninitialized variable `ret`.

This commit resolves the issue by initializing
`ret` to `CSP_ERR_NONE`. `CSP_ERR_NONE` is a
suitable value since an empty queue is equivalent
to a successfully flushed queue.

Although the return value is not currently used,
this change ensures future robustness if the
return value is utilized later.


yashi:
nice catch.  Thanks!

PR #675, opened 2024-09-02
csp_if_zmqhub: Handle return value of csp_iflist_add

moonlight83340:
Update the code to capture the return value of `csp_iflist_add` in the variable `ret`.
This allows for proper error handling by ensuring that the result of `csp_iflist_add` is returned by the function.

Previously, the return value was ignored, which could lead to missing error information.

The return value is now used to confirm successful interface addition, and if an error occurs, it can be properly propagated.

### DIFF
--- a/src/interfaces/csp_if_zmqhub.c
+++ b/src/interfaces/csp_if_zmqhub.c
@@ -208,13 +208,13 @@ int csp_zmqhub_init_w_name_endpoints_rxfilter(const char * ifname, uint16_t addr
 	assert(ret == 0);
 
 	/* Register interface */
-	csp_iflist_add(&drv->iface);
+	ret = csp_iflist_add(&drv->iface);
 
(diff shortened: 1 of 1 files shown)


src/interfaces/csp_if_zmqhub.c
```diff
@@ -208,13 +208,13 @@ int csp_zmqhub_init_w_name_endpoints_rxfilter(const char * ifname, uint16_t addr
 	assert(ret == 0);
 
 	/* Register interface */
-	csp_iflist_add(&drv->iface);
+	ret = csp_iflist_add(&drv->iface);
 
 	if (return_interface) {
 		*return_interface = &drv->iface;
 	}
```
yashi: Do not update `return_interface` when `csp_iflist_add()` failed.

yashi (changes requested):

PR #676, opened 2024-09-03
eth_linux: Handle return value of csp_iflist_add

moonlight83340:
Update the code to capture the return value of `csp_iflist_add` in the variable `ret`.
This allows for proper error handling by ensuring that the result of `csp_iflist_add` is returned by the function.

Previously, the return value was ignored, which could lead to missing error information.

The return value is now used to confirm successful interface addition, and if an error occurs, it can be properly propagated.

ps : I moved the last `}`, it wasn't indent correctly maybe because of an old merge...

### DIFF
--- a/src/drivers/eth/eth_linux.c
+++ b/src/drivers/eth/eth_linux.c
@@ -87,6 +87,7 @@ int csp_eth_init(const char * device, const char * ifname, int mtu, unsigned int
     /* Ether header 14 byte, seg header 4 byte, CSP header 6 byte */
     if (mtu < 24) {
         csp_print("csp_if_eth_init: mtu < 24\n");
+		free(ctx);
         return CSP_ERR_INVAL;
     }
 
@@ -103,6 +104,7 @@ int csp_eth_init(const char * device, const char * ifname, int mtu, unsigned int
         if (count > 0) {
             csp_print("Use command 'sudo setcap cap_net_raw+ep %s'\n", exe);
         }
+		free(ctx);
         return CSP_ERR_INVAL;
     }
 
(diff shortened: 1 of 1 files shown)


moonlight83340 (commented):
Does it make sense to free `ctx` and close the socket ?
```
    /* Register interface */
    int ret = csp_iflist_add(&ctx->ifdata.iface);

	if (ret == CSP_ERR_NONE) {
		if (return_iface) {
			*return_iface = &ctx->ifdata.iface;
		}
	}else {
		close(ctx->sockfd);
		free(ctx);
	}

```

Because in case of `csp_iflist_add `failed to add the interface, the function will return an error so it make sense to me to close and clean.

moonlight83340:
csp_iflist_add is now a void function, so my first commit is irrelevant, I removed it.

yashi:
Do you mind running [valgrind](https://valgrind.org/) to see the difference?

moonlight83340:
> Do you mind running [valgrind](https://valgrind.org/) to see the difference?

I ran [valgrind](https://valgrind.org/) and with no free, I clearly see memory leak.

```
==126853== 200 bytes in 1 blocks are definitely lost in loss record 8 of 12
==126853==    at 0x484D953: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==126853==    by 0x4869323: csp_eth_init (in /home/gaetan/Desktop/libcsp-PR/libcsp/build/libcsp.so)
==126853==    by 0x109858: add_interface (in /home/gaetan/Desktop/libcsp-PR/libcsp/build/examples/csp_server)
==126853==    by 0x1094CB: main (in /home/gaetan/Desktop/libcsp-PR/libcsp/build/examples/csp_server)
```

yashi (approved):
LGTM

PR #677, opened 2024-09-04
usart_kiss: Handle return values of csp_kiss_add_interface

moonlight83340:
Updated the code to handle the return values of both `csp_kiss_add_interface` and `csp_usart_open` to ensure proper error handling.

The function now checks if `csp_kiss_add_interface` and also `csp_usart_open`.

If both operations are successful, the interface is returned via `return_iface`.

This change ensures that errors in the interface addition or USART opening are correctly propagated and handled.

### DIFF
--- a/src/drivers/usart/usart_kiss.c
+++ b/src/drivers/usart/usart_kiss.c
@@ -59,8 +59,13 @@ int csp_usart_open_and_add_kiss_interface(const csp_usart_conf_t * conf, const c
 		res = csp_usart_open(conf, kiss_driver_rx, ctx, &ctx->fd);
 	}
 
-	if (return_iface) {
-		*return_iface = &ctx->iface;
+	if (res == CSP_ERR_NONE) {
+		if (return_iface) {
+			*return_iface = &ctx->iface;
+		}
+	} else {
+		csp_iflist_remove(&ctx->iface);
+		free(ctx);
 	}
 
 	return res;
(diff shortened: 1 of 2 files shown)


src/drivers/usart/usart_kiss.c
```diff
@@ -59,7 +59,7 @@ int csp_usart_open_and_add_kiss_interface(const csp_usart_conf_t * conf, const c
 		res = csp_usart_open(conf, kiss_driver_rx, ctx, &ctx->fd);
 	}
 
-	if (return_iface) {
+	if (res == CSP_ERR_NONE && return_iface) {
 		*return_iface = &ctx->iface;
 	}
```
yashi, line 69: Hmm... I don't see any code "to ensure proper error handling".  When `csp_usart_open()` failed do you leave added interface as is? how about allocated `ctx`?

yashi (changes requested):

src/drivers/usart/usart_kiss.c
moonlight83340: okay, thanks nice catch. ctx is not free if open failed...
In my previous commit : https://github.com/libcsp/libcsp/commit/913fc1713a95d088d8ac0a5fa858fec8926d9ef2
I think I should also free the interface in case of failing...

yashi: Please double check what need to be freed or closed.

yashi (changes requested):

yashi:
@moonlight83340 What was the state of this PR? Do I need to review or do I need to wait a bit more?

moonlight83340:
> @moonlight83340 What was the state of this PR? Do I need to review or do I need to wait a bit more?

I think it was ready for review at the time but allow me some time to check it this week.

PR #680, opened 2024-09-05
csp_if_eth: Add null check for CSP packet in Ethernet RX

moonlight83340:
Added a check to verify if the `csp_if_eth_pbuf_get` function returns a NULL packet.
If the packet is NULL, the interface's drop counter is incremented, and an error message is logged.

The function now returns `CSP_ERR_NOMEM` in case of a NULL packet, ensuring that the error is handled properly and preventing potential issues from processing a NULL packet.

### DIFF
--- a/src/interfaces/csp_if_eth.c
+++ b/src/interfaces/csp_if_eth.c
@@ -175,6 +175,12 @@ int csp_eth_rx(csp_iface_t * iface, csp_eth_header_t * eth_frame, uint32_t recei
     /* Add packet segment */
     csp_packet_t * packet = csp_if_eth_pbuf_get(&pbuf_list, packet_id, task_woken);
 
+	if (packet == NULL) {
+		iface->drop++;
+		csp_print("eth rx can't get csp packet error\n");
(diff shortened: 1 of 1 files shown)


src/interfaces/csp_if_eth.c
```diff
@@ -175,6 +175,12 @@ int csp_eth_rx(csp_iface_t * iface, csp_eth_header_t * eth_frame, uint32_t recei
     /* Add packet segment */
     csp_packet_t * packet = csp_if_eth_pbuf_get(&pbuf_list, packet_id, task_woken);
 
+	if (packet == NULL) {
+		iface->drop++;
```
yashi, line 179: Would you mind explaining why you increment drop counter? Should we increment it when any other error occurs?

moonlight83340: I was thinking that if libcsp can't get the buffer we should increment the drop counter because we couldn't get the packet.
I think in that case it could be a good idea.

moonlight83340:
Since my PR there was modification that make it irrelevant.
I will close it.

PR #682, opened 2024-09-06
csp_driver_can: Handle error for csp_can_add_interface

moonlight83340:
Added error handling for `csp_can_add_interface`.

If the function returns a value different from `CSP_ERR_NONE`, the function
returns `NULL`.

This ensures that in case of error in the function, further processing is halted
in case of an error.

Fix commentary typo in csp_driver_can_init.


johandc:
I will fix this even better by making the csp_iflist_add a void.

PR #684, opened 2024-09-06
interfaces: csp_if_zmqhub: Free memory after failure

moonlight83340:
Improved error handling in the case where csp_zmqhub_init_w_name_endpoints_rxfilter and csp_zmqhub_init_filter2 fails, ensuring the system properly frees resources in all failure scenarios.

### DIFF
--- a/src/interfaces/csp_if_zmqhub.c
+++ b/src/interfaces/csp_if_zmqhub.c
@@ -211,8 +211,16 @@ int csp_zmqhub_init_w_name_endpoints_rxfilter(const char * ifname, uint16_t addr
 	/* Register interface */
 	csp_iflist_add(&drv->iface);
 
-	if (return_interface) {
-		*return_interface = &drv->iface;
+	if (ret == CSP_ERR_NONE) {
+		if (return_interface) {
+			*return_interface = &drv->iface;
+		}
+	} else {
+		pthread_cancel(drv->rx_thread);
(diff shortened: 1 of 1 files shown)


moonlight83340:
I think we should also take care of the thread but because it's detached. I'm unsure.
I have tried to cancel it and it worked but maybe I would need some time between cancel and closing the socket.
(I saw socket error not existing because the thread is trying to use the socket and they are closed. It's appear few time before program close normaly)

https://github.com/libcsp/libcsp/pull/684/files#diff-46fef57db1cbfa4a4994e3aa5f884d2d457ba83a98a11db10539731f4ca7ea4eR207

src/interfaces/csp_if_zmqhub.c
```diff
@@ -210,8 +210,16 @@ int csp_zmqhub_init_w_name_endpoints_rxfilter(const char * ifname, uint16_t addr
 	/* Register interface */
 	ret = csp_iflist_add(&drv->iface);
 
-	if (ret == CSP_ERR_NONE && return_interface) {
-		*return_interface = &drv->iface;
+	if (ret == CSP_ERR_NONE){
+		if(return_interface) {
+			*return_interface = &drv->iface;
+		}
+	}else {
```
yashi: need a space after `}`.

yashi (changes requested):
The issues pointed out must be addressed before merge.

PR #685, opened 2024-09-08
csp_queue: Define csp_queue_handle_t with a concrete type

yashi:
The commit cc71d1ad4156d63e6fee1ce2a10825fca394268c fixed an issue with the argument order. However, the underlying problem, besides human error, was that GCC did not warn about type mismatches. This occurred because `csp_queue_handle_t` was typedef'd as `void *`, which is a generic pointer type. GCC does not warn when passing any pointer type to a `void *` argument or vice versa.

This commit changes `csp_queue_handle_t` to a concrete type instead of `void *`, allowing for better type checking.


johandc:
Thanks. I made a small fix to the compile on posix and merged.

PR #690, opened 2024-09-26
Add CAN to UDP bridging example

takumiando:
Please tell me a better name than ```csp_bridge_can2udp``` if you came up.


src/csp_bridge.c
```diff
@@ -29,10 +29,9 @@ void csp_bridge_work(void) {
 		return;
 	}
 
-	/* Get next packet to route */
+	/* Get next packet to bridge */
 	csp_qfifo_t input;
 	if (csp_qfifo_read(&input) != CSP_ERR_NONE) {
-		csp_print("Failed to receive packet from router input queue\n");
```
yashi, line 35: Or rather, shouldn't we print an error message in `csp_route_work()`? When testing libcsp on tablesats, especially if you're new to libcsp, it's hard to understand what's happening without error messages. In this case, it just dies and exits from `csp_bridge_work()`. This is especially important since the function doesn't return any value, meaning we can't handle it on the caller side.

examples/csp_bridge_can2udp.c
```diff
@@ -0,0 +1,136 @@
...
+	csp_print("%s: %u(%u) --> %s: %u(%u), priority: %u, flags: 0x%02X, size: %" PRIu16 "\n",
+			  iface->name, packet->id.src, packet->id.sport,
+			  dst_name, packet->id.dst, packet->id.dport,
+			  packet->id.pri, packet->id.flags, packet->length);
+}
+
+static void print_help(void) {
+	csp_print("Usage: csp_bridge_can2udp [options]\n");
+	csp_print(" -c <can>   set CAN interface\n");
+	csp_print(" -H <host>  set UDP address\n"
```
yashi: We tend to use `-a` for "address" in other programs?

examples/csp_bridge_can2udp.c
```diff
@@ -0,0 +1,136 @@
...
+#include <getopt.h>
+
+#include <csp/csp.h>
+#include <csp/csp_debug.h>
+#include <csp/drivers/can_socketcan.h>
+#include <csp/interfaces/csp_if_udp.h>
+
+#define RUN_DURATION_IN_SEC	(3)
+#define DEFAULT_CAN_NAME	"can0"
+#define DEFAULT_UDP_HOST	"127.0.0.1"
```
yashi: "HOST" is rather confusing.  You're using "host", "local", and "remote" in this program. I doubt that all three are different.

examples/csp_bridge_can2udp_posix.c
```diff
@@ -0,0 +1,37 @@
...
+
+	ret = pthread_create(&handle, &attributes, routine, NULL);
+	pthread_attr_destroy(&attributes);
+
+	if (ret != 0) {
+		return ret;
+	}
+
+	return CSP_ERR_NONE;
+}
```
yashi: Not for this PR but we have duplicates.  We should extract out.

examples/csp_bridge_can2udp.c
```diff
@@ -0,0 +1,136 @@
...
+#define DEFAULT_UDP_RPORT	(0)
+
+/* These three functions must be provided in arch specific way */
+int bridge_start(void);
+
+static struct option long_options[] = {
+	{"can", required_argument, 0, 'c'},
+	{"host", required_argument, 0, 'H'},
+	{"lport", required_argument, 0, 'l'},
+	{"rport", required_argument, 0, 'r'},
```
yashi: If these are required, why do we need `DEFAULT_`?

examples/csp_bridge_can2udp.c
```diff
@@ -0,0 +1,136 @@
...
+	{"help", no_argument, 0, 'h'},
+	{0, 0, 0, 0}
+};
+
+void csp_input_hook(csp_iface_t * iface, csp_packet_t * packet) {
+	char dst_name[] = "UDP";
+
+	if (strncmp(iface->name, "UDP", 3)) {
+		strncpy(dst_name, "CAN", sizeof(dst_name));
```
yashi: Why the first `strncmp()` uses `3` but the second one uses `sizeof(dst_name)`?

examples/csp_bridge_can2udp.c
```diff
@@ -0,0 +1,136 @@
...
+	csp_bridge_set_interfaces(can_iface, udp_iface);
+	bridge_start();
+
+	/* Print interfaces list */
+	csp_iflist_print();
+
+	/* Wait for execution to end (ctrl+c) */
+	while(1) {
+		sleep(RUN_DURATION_IN_SEC);
+	}
```
yashi: If we don't do anything on main thread, do we need to create another to handle bridging work?

yashi (changes requested):
Hi,

Thank you for your PR. It looks good but I have to ask a few fixes. Thanks.

examples/csp_bridge_can2udp.c
takumiando: ```required_argument(int has_arg)``` means that "This option requires argument".
It doesn't mean that "This argument is required in the command".
Isn't it correct?

src/csp_bridge.c
takumiando: ```csp_bridge_work()``` prints this message until waiting for any messages.

```shell
./build/examples/csp_bridge_can2udp -c can0 -l 52002 -r 52004
INIT CAN: device: [can0], bitrate: 1000000, promisc: 1
RTNETLINK answers: Operation not permitted
RTNETLINK answers: Operation not permitted
RTNETLINK answers: Operation not permitted
RTNETLINK answers: Operation not permitted
  UDP peer address: 127.0.0.1:52004 (listening on port 52002)
LOOP       addr: 0 netmask: 5 dfl: 0
           tx: 00000 rx: 00000 txe: 00000 rxe: 00000
           drop: 00000 autherr: 00000 frame: 00000
           txb: 0 (0B) rxb: 0 (0B) 

CAN        addr: 0 netmask: 0 dfl: 0
           tx: 00000 rx: 00000 txe: 00000 rxe: 00000
           drop: 00000 autherr: 00000 frame: 00000
           txb: 0 (0B) rxb: 0 (0B) 

UDP        addr: 0 netmask: 0 dfl: 0
           tx: 00000 rx: 00000 txe: 00000 rxe: 00000
           drop: 00000 autherr: 00000 frame: 00000
           txb: 0 (0B) rxb: 0 (0B) 

Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
Failed to receive packet from router input queue
...
```

I think that this looks little strange.
Do you have any good idea?

examples/CMakeLists.txt
```diff
@@ -2,12 +2,14 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   add_executable(csp_server_client EXCLUDE_FROM_ALL csp_server_client.c csp_server_client_posix.c)
   add_executable(csp_server EXCLUDE_FROM_ALL csp_server.c csp_server_posix.c)
   add_executable(csp_client EXCLUDE_FROM_ALL csp_client.c csp_client_posix.c)
+  add_executable(csp_client EXCLUDE_FROM_ALL csp_bridge_can2udp.c csp_bridge_can2udp_posix.c)
```
yashi: This must not be `csp_client` but `csp_bridge_can2_udp`  or whatever you choose.

src/csp_bridge.c
yashi: `csp_qfifo_read()` only returns either `CSP_ERR_NONE` or `CSP_ERR_TIMEDOUT`, so logging for `CSP_ERR_TIMEDOUT` isn't necessary.

Thank you for clarifying this.

yashi (approved):
LGTM.  Thanks.

PR #693, opened 2024-10-01
rdp: Removed EACK

johandc:
The EACKs are part of the RDP RFC. However, for satellite communication the link may be degraded by EACKs because they are rapidly generated and otherwise transmitted out ot sequence. (1 ack per packet) This messes with the rdp_options and we have found this to be both confusing and detrimental to reliable communication. Especially when dealing with half-duplex links.
This commit simply stops generating EACKs and relies instead on the normal packet timeout and retransmission procedure. It has been tested for almost a year in production at Space Inventor.

### DIFF
--- a/doc/basic.md
+++ b/doc/basic.md
@@ -62,7 +62,6 @@ Definition of a buffer element `csp_packet_t`:
    the CSP id to different endian (e.g. I2C), etc.
 */
 typedef struct {
-    uint32_t rdp_quarantine;      // EACK quarantine period
     uint32_t timestamp_tx;        // Time the message was sent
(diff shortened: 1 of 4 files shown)


yashi:
Fixed the linelint issue by #696. Rebase this PR to see it works or not.

yashi:
We have "Extended Acknowledgment" mentioned in `doc/protocolstack.md`. I'd be nice if we explicitly state that we do not support EACK in RDP.

We can also remove `rdp_quarantine` from `csp_packet_s`.

Otherwise, LGTM.

johandc:
I added the comment in dog/protocolstack and removed rdp_quarantine from csp_packet_s.

PR #694, opened 2024-10-01
service: peek/poke v2 with 64 bit memory support

johandc:
RFC:

I would like some feedback for a smarter way of allowing 64 bit access, this includes on 32 bit systems. So memcpy() is not always an option. We have a 32-bit MCU with 64 GB external storage, memory mapping this requires a larger address space.

### DIFF
--- a/include/csp/csp.h
+++ b/include/csp/csp.h
@@ -481,9 +481,11 @@ void csp_rdp_get_opt(unsigned int *window_size, unsigned int *conn_timeout_ms,
 	  unsigned int *ack_timeout, unsigned int *ack_delay_count);
 
 /**
- * Set platform specific memory copy function.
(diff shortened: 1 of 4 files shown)


yashi:
I'm curious how a 32-bit MCU can map 64 GB in linear address space. ARM's Large Physical Address Extensions (LPAE) supports 40-bit address space but only support in 32-bit virtual address space.  x86 does the same. Any other arch?

johandc:
We are not using the processor address space, but out  own virtual memory system
https://github.com/spaceinventor/libparam/tree/master/src/vmem
This supports 64 bit addresses also on 32 bit MCUs.

The question of getting this exposed we have had to make a custom memcpy64 function, which i feel messes up the user API a little. This is the reason i did not directly merge this one.

Issue #603, opened 2024-07-04
CSP RDP Queue is not thread-safe


yashi:
Currently CSP RDP queue manipulation code (`csp_rdp_queue.c`), at least, is not thread safe.  It's impossible to use the code from multiple threads.

dimitrovand:
I can confirm the issue is present and reproducible. It's difficult to think of a clean solution that doesn't involve reintroducing mutexes...

yashi:
If we were to implement this from scratch, I would define a `struct csp_rdp_queue`. However, the current implementation uses implicit queues.

The problematic part is the dequeue -> enqueue section in the flush and get functions. Therefore, the simplest solution would be to add a lock around these sections.

The nobrainer approach would be to just add locks to the public functions. but that's too stupid. all method of csp_queue should be thread safe anyways.

I don't know how bad it'd be if we have per-connection queues.

johandc:
I think we whould consider using a lighter weight system and then go back to per-connection.
The issue with the old per-connection queues, was that they had to be statically pre-allocated.
If we would use linked-lists, with a simple atomic.h lock for pertubations, the memory size is extremely small. And there will be no additional memory required to store queues, and we dont invoke the larger queue api.

The queue api was specifically chosen for its thread safety, but with the way we have to take out, and insert to scan through a list ruins that anyways.

When the compiler supports atomic operations a mutex lock is extremely simple to add on the linked list.

blhiggins:
Can I ask what the symptoms of this bug are for the user? (Wondering if segfaults I've been hitting at `___pthread_mutex_lock()` calls inside `pthread_queue_dequeue()` or `pthread_queue_items()` might be related to this bug or not.)

And isn't threading required to run the router task because csp_read() blocks execution (at least until timeout)? Does this imply RDP is not in a usable state at all right now?

Issue #595, opened 2024-06-24
csp_rdp_queue_rx_add shouldn't be void


moonlight83340:
Hi,

I was investigating the source code and was surprised because the `csp_rdp_queue_rx_add` function is void, and I don't think it should be the case. We should know if the add function fails or not.

Here is the definition of `csp_rdp_queue_rx_add` : https://github.com/libcsp/libcsp/blob/develop/src/csp_rdp_queue.c#L83

`csp_rdp_queue_rx_add` is used in another function, `csp_rdp_rx_queue_add`.

The definition of `csp_rdp_rx_queue_add` is here: https://github.com/libcsp/libcsp/blob/develop/src/csp_rdp.c#L318

This function returns -1 if a duplicate is found and 0 if not. So if `csp_rdp_queue_rx_add` fails for another reason, there is no way to know, and it will send an EACK anyway here, for example: https://github.com/libcsp/libcsp/blob/develop/src/csp_rdp.c#L738

Libcsp should verify every time it tries to add a packet to the queue if it was successful or not.

johandc:
The whole idea if using RDP is to avoid RX buffer exchaustion, therefore the rx queue should be able to be void, because the receiver will ever not ack more than what should be available.

PR #699, opened 2024-10-07
rdp: Replace RDP queues with per-conn linked lists

dimitrovand:
This update transitions from the existing RDP queues to per-connection linked lists for RX and TX packets. The new approach enhances both efficiency and thread safety by providing isolated, direct access to each connection’s packet list.

As a result, 'csp_rdp_queue.c' and 'csp_rdp_queue.h' have been removed, as their functionality is no longer required with the new direct-access approach.

If eventually accepted, it addresses https://github.com/libcsp/libcsp/issues/603 and https://github.com/libcsp/libcsp/issues/595 

I would really appreciate your thoughts and some help with zephyr tests.


src/csp_conn.c
```diff
@@ -201,6 +204,11 @@ csp_conn_t * csp_conn_new(csp_id_t idin, csp_id_t idout, csp_conn_type_t type) {
 
 		/* Ensure connection queue is empty */
 		csp_conn_flush_rx_queue(conn);
+
+#if (CSP_USE_RDP)
+		/* Ensure connection rdp queue is empty */
+		csp_rdp_queue_flush(conn);
```
johandc: We should flush on conn_close() instead of conn_new().
The reason is that there could otherwise be a number of allocated packet buffers assigned on closed connections and therefore starve the system of CSP buffers.

src/csp_conn.h
```diff
@@ -74,6 +74,12 @@ struct csp_conn_s {
 	uint32_t opts;              /* Connection or socket options */
 #if (CSP_USE_RDP)
 	csp_rdp_t rdp; /* RDP state */
+	
+	atomic_int rdp_rx_lock;     /* 1 (locked), 0 (unlocked) */
```
johandc: Perfect! This is exactly how i envisoned the implementation.

src/csp_init.c
```diff
@@ -32,9 +31,6 @@ void csp_init(void) {
 	csp_buffer_init();
 	csp_conn_init();
 	csp_qfifo_init();
-#if (CSP_USE_RDP)
-	csp_rdp_queue_init();
```
johandc, line 41: Really nice to get rid of this one.

src/csp_rdp.c
```diff
@@ -48,6 +48,64 @@ typedef struct __packed {
 
 static int csp_rdp_close_internal(csp_conn_t * conn, uint8_t closed_by, bool send_rst);
 
+static inline void csp_spin_lock(atomic_int * lock) {
+    int expected = 0;
+    /* Attempt to acquire the lock by setting it to 1 */
+    while (!atomic_compare_exchange_strong(lock, &expected, 1)) {
+        expected = 0; /* Reset expected value */
+        /* Retry until lock becomes available */
```
johandc: Here we need to insert a OS hook for yielding the current task and/or sleeping.
The problem with yielding is that if the router task has higher priority (which it typically does) the lower priority task would not wake up and release its lock. So this calls for a mutex with priority inheritance.

I will leave some more comments regarding this at the end of the review.

johandc (changes requested):
First of all thank you for producing a really nice and high quality pull request. The changes you have made is exactly what i also proposed. So big thumbs up for that. 

I have come to realise that the spinlock may be completely removed, because as far as i can tell, there is no conflict between the router task and the application task.

There are multiple entires to the rx/tx queues:
 1) The router task when dealing with a new packet
 2) The router task when calling csp_rdp_check_timeouts()
 3) The router task when calling csp_rdp_close()
 4) The application task when calling csp_close()
 3) The application task when calling csp_send() and csp_recv()

So lets quicly see if we can make a scheme, where all conflicts are avoided.

 1) I think we can say that the router task should have exclusive access to the lists :heavy_check_mark: 
 2) We already gave the router task exclusive access, we also have a guarantee that nobody else can call csp_rdp_check_timeouts() because it is only accessible from csp_rdp.h, which is in the src foler not the public include folder. Same goes for csp_conn_check_timeouts. :heavy_check_mark: 
 3) If the RDP connection gets closed by a broken pipe, the connections is markes af CONN_WAIT. It does not immediately flush or close the connection. So here there is not change of the lists. :heavy_check_mark: 
 4) In when csp_close is called on a RDP it is not actually closd, and it is deferred to a timeout in the router task to fully close the connection. As such it is never the application task that closes. Therefore rdp_flush_queue should be safe to call in csp_rdp_close (and should be moved to here in fact) :heavy_check_mark: 
 5) The application can trigger a window update csp_read() only refers to the connection RX queue, and calls csp_rdp_check_ack. This one does not access the out-of-order RX queue inside the RDP layer. So no need for a lock here. csp_send() this one adds a copy of the packet to the tx_queue. So here there is a conflict :warning: 

conclusion: We can already skip the RX locking. The TX lock however cannot be skipped without any further actions.

One idea could be to allow read-only from router task and read-write from application task. This means that the application task cleans out and frees transmitted elements when its calling csp_rdp_send(). The RDP task already knows to not retransmit them. Whenever there is room in the TX queue, we notify the user task on the tx_wait semaphore anyways.

With the aforementioned changes, we should be able to completely skip the spinlock :face_with_spiral_eyes: 

Thanks again.

johandc:
I would like to add, that adding this scheme for "protection" of the lists, requires a contract on who has access to modify the list and who can read-only. So we should put some usefull comments around the tx_list and rx_list fields, as well as some block comments in the area where the area accessed.

src/csp_conn.c
dimitrovand: You're right. Flushing in conn_close() is better (and is actually there). I added it to conn_new() as well, just as secondary protection, but moving csp_rdp_queue_flush() to csp_rdp_close() might be better, as you’ve pointed out.

src/csp_rdp.c
dimitrovand: You're absolutely right!

dimitrovand:
Thanks for the detailed feedback! 
Let me know what you think of the last 2 commits (they address rx locking and queue flushing). 
I'm a bit uncertain how we can completely avoid locking the tx list. My understanding of the RDP state machine is not very good.

dimitrovand:
@johandc I'm thinking of using mutexes to solve the tx queue concurrency issues. Let me know what you think.

johandc:
I have come to favor the old scheme a bit more without using linked lists. The reason is that linked lists uses dynamic pointer dereferencing and requires complex locking. The scheme used in the conn table, is much simpler, and only requires a single atomic lock which can be implemented by processer native instructions.

dimitrovand:
@johandc sure, you can decline the PR

PR #705, opened 2024-10-11
csp_rdp: Prevent NULL pointer dereference

dimitrovand:
Adds a check to ensure successful buffer allocation, preventing potential NULL pointer dereference when buffers are unavailable. Introduces the 'csp_buffer_copy' function to improve safety and handle buffers more reliably.

### DIFF
--- a/src/csp_buffer.c
+++ b/src/csp_buffer.c
@@ -130,20 +130,20 @@ void csp_buffer_free(void * packet) {
 	csp_queue_enqueue(csp_buffers, &buf, 0);
 }
 
-void * csp_buffer_clone(void * buffer) {
-
-	csp_packet_t * packet = (csp_packet_t *)buffer;
-	if (!packet) {
-		return NULL;
-	}
-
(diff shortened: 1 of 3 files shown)


src/csp_buffer.c
```diff
@@ -130,22 +130,21 @@ void csp_buffer_free(void * packet) {
...
+	if (packet) {
+		clone = csp_buffer_get(0);
+		csp_buffer_copy(packet, clone);
 	}
 
 	return clone;
 }
 
+void csp_buffer_copy(const csp_packet_t * src, csp_packet_t * dst) {
+	if (src && dst)
```
yashi: MISRA isn't enforced in our code base but it'd be better to compare thse two pointers against `NULL`. Also I'd like to have `{}` even with a single statement.
```C
if ((src != NULL) && (dst != NULL))  {
    (void)memcpy(...);
}
```

yashi (changes requested):
I left comment about syntax but overall it looks good. Thanks.

Issue #708, opened 2024-10-15
packet->cfpid is unexpectedly overwritten by csp_id_strip()


takumiando:
## Description

This is related to https://github.com/libcsp/libcsp/issues/531 and https://github.com/libcsp/libcsp/issues/695.

```cfpid``` has been broken with the commit https://github.com/libcsp/libcsp/commit/b74b25ec3757a4c8c39a54d25f4d9d8de731f6e9.  
So libcsp V1 cannot receive data over 2 bytes correctly.

Unless there is a strong reason, we shouldn't use union in ```csp_packet_s```.

**Comment by takumiando** [comment]
This bug is similar to https://github.com/libcsp/libcsp/issues/695.

The timing of the execution of ```csp_id_strip()```(this means ```timestamp_rx = 0```) differs between CSP V1 and V2.  
So ```cfpid``` will not be broken in V2.


If we can freely modify either struct in a union, there’s no reason to use union.

**Comment by yashi** [comment]
Please provide steps to reproduce the bug.

**Comment by takumiando** [comment]
@yashi 

### Apply this patch to use CSP v1

```diff
diff --git a/examples/csp_client.c b/examples/csp_client.c
index 37625a3..b59a804 100644
--- a/examples/csp_client.c
+++ b/examples/csp_client.c
@@ -139,6 +139,8 @@ int main(int argc, char * argv[]) {
        int ret = EXIT_SUCCESS;
     int opt;
 
+       csp_conf.version = 1;
+
        while ((opt = getopt_long(argc, argv, OPTION_c OPTION_z OPTION_R "k:a:C:tT:h", long_options, NULL)) != -1) {
         switch (opt) {
             case 'c':
diff --git a/examples/csp_server.c b/examples/csp_server.c
index e124c28..a42eb4a 100644
--- a/examples/csp_server.c
+++ b/examples/csp_server.c
@@ -184,6 +184,8 @@ int main(int argc, char * argv[]) {
        csp_iface_t * default_iface;
     int opt;
 
+       csp_conf.version = 1;
+
        while ((opt = getopt_long(argc, argv, OPTION_c OPTION_z OPTION_R "k:a:tT:h", long_options, NULL)) != -1) {
         switch (opt) {
             case 'c':
```

### Build examples

```shell
$ ./examples/buildall.py
```

### Run server/client examples

Pings don't return.

#### Server
```shell
$ ./build/examples/csp_server -c vcan0 -a 1
Initialising CSP
[...]

PR #709, opened 2024-10-15
Add support for testing with CSP version 1

takumiando:
- Add CSP version option to all the examples
- Add tests for CSP version 1

This patch set is to avoid bugs specific to CSP version 1.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -18,6 +18,9 @@ jobs:
             CXX: g++
           - CC: clang
             CXX: clang++
+        csp_version:
+          - 1
+          - 2
     runs-on: ${{ matrix.os }}
     steps:
       - name: Setup packages on Linux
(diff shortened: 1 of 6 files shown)


takumiando:
This is related to https://github.com/libcsp/libcsp/issues/708.

yashi:
Work on the commit message a bit.  "Modify for CSP version 1" doesn't tell you anything.  Any commit is a modification.

takumiando:
> Work on the commit message a bit. "Modify for CSP version 1" doesn't tell you anything. Any commit is a modification.

I changed PR title.

Currently, the added tests for CSP v1 fail because of the issue https://github.com/libcsp/libcsp/issues/708.

.github/workflows/build-test.yml
yashi, line 1: Can we do this with [github actions matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow)? Having copies of everything is tedious.

takumiando: I'll try it.

takumiando:
like this?
image

yashi:
Thanks. That's more manageable. LGTM but this need to wait for a bit to be merged.

yashi:
@takumiando Please rebase it.

Issue #712, opened 2024-10-21
__init__.py wrapper directory for PyCSP


kivkiv12345:
Hello good people of libcsp upstream.

A long time ago I created [PyCSH](https://github.com/spaceinventor/PyCSH), using [PyCSP](https://github.com/libcsp/libcsp/blob/develop/src/bindings/python/pycsp.c) to get started with the CPython API.

Since then I've learned quite a few things about building C extensions (with Meson), some of which are not demonstrated by PyCSP, but have been added to PyCSH.
Most relevant for PyCSP is probably an \_\_init__.py directory, which can house a .pyi file with API documentation.
This directory can also be used for higher-level APIs implemented in native Python, building off the lower-level bindings (although maybe this is not desired for PyCSP?).

I'll be happy to implement this in a PR, if it doesn't conflict with some existing philosophy regarding the generated .so Python module.

kivkiv12345:
Looking at the bindings/CSP API, it seems that they would go well with some context managers, for opening/closing connections and such.
It would be easy to write such a class in native Python with \_\_init__.py, but I dont know how people feel about such deviations from the C API.

Issue #636, opened 2024-08-11
Python to PyPI


FelixEL:
Will pythonic version of this library be published on pyPI?

yashi:
I'm not really a python guy.  A PR is always welcome :smile:

parker-research:
+1 for this. As long as there's no disagreement with it, I'd be happy to help prep this! Is there an account management system to do this? We'd need a `libcsp` PyPI.org account, of course.

Is the Python code versioned alongside the main `libcsp`, or should it get its own version numbers?

johandc:
I think we should aim for a Native Python implementation of CSP instead.

Lots of what is going on inside CSP, the router, the many network interfaces, the low level OS stuff, is really gone when inside python context. Most likely the main interface used with  python is going to be ZMQ which is already native python as well.

DeflateAwning:
I think this is good enough for now. Really targeting ease of use for ground stations would be huge

kivkiv12345:
A more immediate alternative could be to create a [pyproject.toml](https://github.com/spaceinventor/PyCSH/blob/master/pyproject.toml) file, allowing the repository itself to be pip3 installed.

I have tried this myself with the Space Inventor repository [PyCSH](https://github.com/spaceinventor/PyCSH), although here we mostly use an [./install](https://github.com/spaceinventor/PyCSH/blob/master/install) script to pip3 install a local build. But a GitHub URL can also be used.

DeflateAwning:
That is indeed the first step!

kivkiv12345:
I've made a PR for the pyproject.toml part:
https://github.com/libcsp/libcsp/pull/714
I took a creative liberty by bundling libcsp.so into the pip wheel,
but I think this will be greatly appreciated when users can install directly from PyPI without needing to install libcsp on their system first.

parker-research:
Would be awesome to get a backport to v1.6 for this, and have that lineage published to PyPI!

kivkiv12345:
I think you can configure the newest commit to CSP 1.6, right?

parker-research:
I'm just not too sure if all the other changes in that PR make sense to include in the backport; can you advise? Are any of them breaking changes?

kivkiv12345:
* I don't know whether the bindings API has seen breaking changes since 1.6 (outside of this PR), where the .pyi file would need to be backported.
[...]

PR #714, opened 2024-10-23
python3-bindings: Add pyproject.toml to support pip build and installation

kivkiv12345:
This PR proposes to add a pyproject.toml file, allowing the python3 bindings to be built and installed with pip.
```
# To build the bindings
pip3 wheel -w builddir-pip .
# To install the bindings
pip3 install $(ls builddir-pip/libcsp*.whl)
```

Please note that this initial suggested implementation uses [meson-python](https://pypi.org/project/meson-python/) as its build backend. I am unsure how we should structure the pyproject.toml to also support waf and cmake.
Also; please refer to this description from 93f4b5c, regarding concessions made to the meson.build script:
- 'pure' and 'copy' args used in meson.build,
   these were introduced after the minimum specified meson version,
   which we should therefore probably bump (unless workarounds can be found).
- Headers are now longer installed when building with enable_python3_bindings=true,
  as meson-python doesn't know what to do with them.
- configure_file('autoconfig.h') has explicit install: false,
  as it otherwise breaks meson-python.

Additionally, the resulting pip .whl installs as a Python package (directory with __init__.py).
A copy of libcsp.so is then bundled into this package, and will be used as a fallback to the system dependency.
This makes it possible to import the bindings without needing to worry LD_LIBRARY_PATH, PYTHONPATH or other system configuration.
It should also leave the library compatible with any previous use-cases.
Alternatively the bindings may easily be linked statically with libcsp.so, but this may conflict with some existing use-cases where the bindings are linked with existing symbols/namespace.

What do you guys think?

I also took the liberty of creating a .pyi file for intellisense, and fixed some minor errors in the bindings themselves.

I initially suggested an __init__.py wrapper in #712, and suggested a pyproject.toml in #636

### DIFF
--- a/examples/csp_server_client.py
+++ b/examples/csp_server_client.py
@@ -56,7 +56,7 @@ def client_task(addr: int, port: int) -> None:
         if conn is None:
             raise Exception('Connection failed')
 
-        packet = csp.buffer_get(0)
+        packet = csp.buffer_get()
         if packet is None:
             raise Exception('Failed to get CSP buffer')
 
--- a/examples/python_bindings_example_server.py
+++ b/examples/python_bindings_example_server.py
@@ -77,7 +77,7 @@ def csp_server():
 
                 # free up a buffer to hold the reply
                 # parameters: {buffer size (# of 4-byte doublewords)}
-                reply = libcsp.buffer_get(0)
+                reply = libcsp.buffer_get()
 
                 # store the data into the reply buffer
                 libcsp.packet_set_data(reply, data)
--- a/include/csp/meson.build
+++ b/include/csp/meson.build
@@ -1,1 +1,2 @@
-csp_config_h = configure_file(output: 'autoconfig.h', configuration: conf, install_dir: 'include/csp/')
+# pip3 freaks out if it thinks it must install autoconfig.h
+csp_config_h = configure_file(output: 'autoconfig.h', configuration: conf, install_dir: 'include/csp/', install: false)
--- a/meson.build
+++ b/meson.build
@@ -110,7 +110,7 @@ if get_option('enable_python3_bindings')
 	pydep = dependency('python3', version : '>=3.5', required : true)
 	py.extension_module('libcsp_py3', 'src/bindings/python/pycsp.c',
                     	c_args : csp_c_args,
-                    	dependencies : [csp_dep, pydep],
+                    	dependencies : [csp_dep.partial_dependency(links : false, includes : true), pydep],
 						subdir : 'libcsp_py3',
                     	install : true)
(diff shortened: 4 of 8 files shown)


kivkiv12345:
I am considering whether marking [pycsp_buffer_get()](https://github.com/libcsp/libcsp/blob/6a458dcf061c8681c05be3c150b3085df6eb0b92/src/bindings/python/pycsp.c#L650) with [METH_NOARGS](https://github.com/libcsp/libcsp/blob/6a458dcf061c8681c05be3c150b3085df6eb0b92/src/bindings/python/pycsp.c#L987) in [6a458dc](https://github.com/libcsp/libcsp/pull/714/commits/6a458dcf061c8681c05be3c150b3085df6eb0b92) was a mistake. It removes the option to call the function with the unused size argument, and therefore deviates from the C API.
It broke the [examples/tests](https://github.com/libcsp/libcsp/pull/714/commits/92ce2461fb69ec54937e1aa4cac90a6ccb16ab55), but I would argue that they should demonstrate that the argument is optional/unused anyway.
If corrective action is to be taken, I think we should leave the tests calling the function without the argument. But [pycsp_buffer_get()](https://github.com/libcsp/libcsp/blob/6a458dcf061c8681c05be3c150b3085df6eb0b92/src/bindings/python/pycsp.c#L650) should specifically only parse the unused size argument, such that it may still raise exceptions for incorrect or excessive arguments.

kivkiv12345:
I am beginning to doubt that a bundled libcsp.so can be made as backwards compatible as I had initially hoped.
The way the bindings are currently built, they will link against a system library for libcsp (eg. /usr/local/lib/libcsp.so).
The import will then fail if this file doesn't exist, no matter if all symbols are resolved through the bundled library.

It seems to me that the system library must be resolved in the \__init__.py, if we are to have a bundled version as a fallback.
But I am unsure if there is a clean way to abstract the location of the system dependency, keeping the path identical to those linked by meson.build.

kivkiv12345:
Some of the tests are failing due to the minimum Meson version being increased.
Is this ever permissible for a PR to do?

yashi:
Hi @kivkiv12345, 

Sorry to be late, but would you mind reabsing it? Thanks

kivkiv12345:
Sure thing @yashi!
I should also be able to make the meson.build part a little prettier,
based on my most recent experience from https://github.com/spaceinventor/pycsh

yashi:
> I should also be able to make the meson.build part a little prettier,

Sure. I can wait.

yashi:
Please re-open this PR whenever you are ready.  For issue and PR management purposes, I’m closing this for now.

PR #715, opened 2024-10-25
interfaces: udp: Fix packet size check

Nari-19:
In this commit, the comparison target of received_len was changed to header-size. 
In the previous case, CSP v1 and v2 have different header-size and csp_if_udp_rx_work does not work properly.
The inequality sign in the comparison was changed to receive packets with payload size of 0.


yashi:
Hi, thank you for your contribution!

Your commit log is a bit hard to understand, so I’d like to confirm my interpretation.

This commit actually addresses two issues:

1. A literal number `4` is used for both protocol v1 and v2. For v1, the header size is `4`, but for v2, it’s `6`, as defined by `CSP_ID1_HEADER_SIZE` and `CSP_ID2_HEADER_SIZE`, respectively.
2. Because `<=` is used in the comparison, it drops v1 packets that have no payload. In other words, it drops v1 packets with a size of `4`. Since v2 packets have at least `6` bytes, they are not affected by this bug. (Or bad packets are accepted reagardless).

Would you mind revising your commit log to clarify these points?

Thanks!

yashi:
@Nari-19 You still need to fix one GitLint error.

PR #729, opened 2024-11-12
Add __maybe_unused macro

yashi:
Add `__maybe_unused` to suppress warnings for variables that may be unused depending on build configurations, such as `NDEBUG`.


yashi:
`__maybe_unused` is kept explicitly to our code base and not exported to applications using libcsp. Zephyr has it's own `__maybe_unused` by itself, thus we can use it in the sample code for Zephyr.

Issue #734, opened 2024-11-14
csp_buffer: Is it considered normal to have access to the data of previously used buffers?


dimitrovand:
## Description
Is it considered normal to have access to the data of previously used buffers?

## Steps to Reproduce
Run this code with default CSP settings:
```c
#include <csp/csp.h>

int main() {
    csp_init();

    // use all buffer and free them
    for (unsigned int i = 0; i < CSP_BUFFER_COUNT; i++) {
        csp_packet_t * packet = csp_buffer_get_always();
        memset(packet->data, 0, sizeof(packet->data)); // clear buffer data
        memcpy(packet->data, "previous_data!!", i+1);  // put some data inside
        csp_buffer_free(packet);
    }

    // access the data of previously used buffers
    for (unsigned int i = 0; i < CSP_BUFFER_COUNT; i++) {
        csp_packet_t * packet = csp_buffer_get_always();
        csp_print("%lu: data=%s\n", (uintptr_t)packet, packet->data);
        csp_buffer_free(packet);
    }
}
```
## Expected behavior

Data, from previously used buffers, should not be visible.

## Actual Behavior

Data, from previously used buffers, is visible.

## Environment Details

- Build OS: any
- Target OS: any
- libcsp commit hash: any
- Meson version: any
- CMake version: any
- GNU Make version: any
- Ninja version: any

## Additional Context
Output from provided code snippet:
```
126213084001520: data=p
126213084001856: data=pr
126213084002192: data=pre
126213084002528: data=prev
126213084002864: data=previ
126213084003200: data=previo
126213084003536: data=previou
126213084003872: data=previous
126213084004208: data=previous_
126213084004544: data=previous_d
126213084004880: data=previous_da
126213084005216: data=previous_dat
126213084005552: data=previous_data
126213084005888: data=previous_data!
126213084006224: data=previous_data!!
```

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

yashi:
Hmm... What do you think? It's nice to clear the contents of a buffer before allocating. But that comes with a cost. In fact, if no MPU is involved, we can access anything anyway.

But, at least the pointers should be initialized, IMO.

[...]

Issue #731, opened 2024-11-12
frame_begin should be initialized after csp_buffer_get()


yashi:
As you can see from the following code, without `csp_id_prepend(packet);`, a segmentation fault occurs because `frame_begin` in the packet is not properly initialized but set to `0x0`.

```c
#include <csp/csp.h>
#include <csp/csp_debug.h>
#include <csp/csp_id.h>
#include <csp/crypto/csp_hmac.h>

int main(int argc, char * argv[])
{
	csp_packet_t * packet;

	csp_init();

	packet = csp_buffer_get_always();
	csp_id_prepend(packet);

	csp_hmac_append(packet, true);
	csp_hmac_verify(packet, true);

	return 0;
}
```

This issue was reported by @poomravint in #730.

yashi:
Fixed by #735

PR #735, opened 2024-11-15
Initialize packet buffer

yashi:
A packet buffer should be ready to be used when users acquired it from csp_buffer_get() or csp_buffer_get_always().

This PR has two commits a) variable renaming for easier reading, and b) initializing buffer.

It doesn't have memset() in it yet, which is raised by #734. I'm thinking about compile time configuration to switch off the memset() because some user might not care about the contents and don't want to spend time on erasing it.

This PR also addreses #731.


yashi:
@dimitrovand WDYT?

dimitrovand:
Looks good to me!

PR #742, opened 2024-11-22
Rework SFP protocol implementation, add tests and sample code

dimitrovand:
1. Removes the malloc usage by introducing a more generic interface for reading/writing from/to storage.
    The protocol functionality is not modified, so it remains fully backward compatible.
2. Introduce a test for the SFP protocol using CSP in loopback mode.
    The test sends random data to a client, verifying functionality by
    comparing the received size with the sent size and ensuring CRCs
    match on both ends.
    The backward compatibility test have been performed locally since there are no previous tests 
    in older versions to be used. SFP tests must be extended to also separate server/client to avoid such scenarios in future.
3. Introduce a sample demonstrating the use of SFP protocol for file transfer.


yashi (approved):
LGTM

yashi:
Actually the first commit is not "refactoring". It
- Removes `csp_sfp_send_own_memcpy()` and `csp_sfp_recv_fp()`
- Changes the API of `csp_sfp_send()` by introducing `csp_sfp_read_t` and removes `timeout`:
  - `int csp_sfp_send(csp_conn_t * conn, const void * data, unsigned int datasize, unsigned int mtu, uint32_t timeout)`
  - `int csp_sfp_send(csp_conn_t * conn, const csp_sfp_read_t * user, uint32_t totalsize, uint32_t mtu)`
- Changes the API of `csp_sfp_recv()` by introducing `csp_sfp_recv_t`:
  - `int csp_sfp_recv(csp_conn_t * conn, void ** dataout, int * datasize, uint32_t timeout)`
  - `int csp_sfp_recv(csp_conn_t * conn, const csp_sfp_recv_t * user, uint32_t timeout)`
  - `csp_sfp_read_t` is basically `data`, `datasize`, and `csp_memcpy_fuc_t` combined.
  
> code refactoring is the process of restructuring existing [...] without changing its external behavior
-- https://en.wikipedia.org/wiki/Code_refactoring

The first commit is a bit too big to review. Quoting your commit log:

> Removes the malloc usage by introducing a more generic interface
for reading/writing from/to storage.

`malloc()` is only used in `csp_sfp_recv_fp()`. Thus, if your commit log is right, it should only change `csp_sfp_recv_fp()`.

The point having `malloc()` in `csp_sfp_recv_fp()` is that `totalsize` is unknown at `csp_sfp_recv_fp()`, IIRC. And there are two options we can take (without using `malloc()`:

- a) Pass a best large buffer.
- b) Let user handle it.

Your implementation is basically (a). Then, what's the point of having `csp_sfp_recv_t`?
Also your test code is not testing error cases, IIRC.

For (a), you can also notify users that `totalsize` in the first packet is larger than given buffer. Just return `CSP_ERR_TOOBIG`
from `csp_sfp_recv_fp()` and let user call the function again with a larger buffer. I'm not sure it works as I imagine though.

For (b), you can also make the custom write function to return `void *` so that when the custom write function decided to `realloc()` the buffer bcause the original buffer is too small for incoming data it can do so and return the newlly allocated buffer.  I'm not saying this is needed but that's what the original code with `malloc()` is doing.

The complication of the functions are that it allows user to use a custom `memcpy()`-like functions. Maybe this is related to #694.

src/csp_sfp.c
```diff
@@ -1,191 +1,194 @@
...
- * The following functions are helper functions that handles the extra SFP
- * information that needs to be appended to all data packets.
+ * Helper function to remove the SFP header from a packet, performing bounds checking
+ * and endianness conversion.
+ *
+ * @param packet Pointer to the CSP packet.
+ * @return Pointer to the SFP header if valid, NULL otherwise.
  */
-static inline sfp_header_t * csp_sfp_header_add(csp_packet_t * packet) {
+static inline csp_sfp_hdr_t * csp_sfp_hdr_rem(csp_packet_t * packet) {
```
yashi: Please spell them all out `csp_sfp_header_remove()`.

yashi: You have removed `csp_sfp_header_add()` but you introduce `csp_sfp_header_remove()`? Can we keep `csp_sfp_header_add()`?

src/csp_sfp.c
```diff
@@ -1,191 +1,194 @@
...
+    csp_sfp_hdr_t * hdr = NULL;
+    const uint32_t hdr_sz = sizeof(* hdr);
 
-	sfp_header_t * header = (sfp_header_t *)&packet->data[packet->length];
-	packet->length += sizeof(*header);
-	return header;
-}
+    // Check if the fragment flag is set and length is sufficient
```
yashi: We tend to use old `/* */` comment style.

include/csp/csp_sfp.h
```diff
@@ -4,89 +4,108 @@
...
+     * User-defined function for reading data from any type of storage.
+     * 
+     * @param[in] data Pointer to user-specific data required for the read operation. Could be NULL (depends on user).
+     * @param[out] buffer Pointer to the buffer where the read data should be copied. Always valid.
+     * @param[in] offset The offset in the storage from where the data should be read.
+     * @param[in] size The number of bytes to read from the storage. Never 0.
+     * @return Read bytes, or a value different from \a size to terminate.
+     */
+    uint32_t (* read)(void * data, uint8_t * buffer, uint32_t offset, uint32_t size);
+} csp_sfp_read_t;
```
yashi: - Argument order of `read(2)` is `src fd`, `out buf`, `read count`.
- Argument order of `memcpy(3` is `out buf`, `in buf`, `read count`.

If you name it `read`, the current order make sense but if you are targetting memory location, latter is better.

Offset is kinda odd here. What is the usecases? 
We don't have any function with `offset` so it should be the last arg.

include/csp/csp_sfp.h
```diff
@@ -4,89 +4,108 @@
...
+    /** 
+     * User-defined function for writing data to any type of storage.
+     * 
+     * @param[in] data Pointer to user-specific data required for the write operation. Could be NULL (depends on user).
+     * @param[in] buffer Pointer to the buffer containing the data to be written. Always valid.
+     * @param[in] offset The offset in the storage where the data should be written.
+     * @param[in] size The number of bytes to write to the storage. Never 0.
+     * @return Written bytes, or a value different from \a size to terminate.
+     */
+    uint32_t (* write)(void * data, const uint8_t * buffer, uint32_t offset, uint32_t size);
```
yashi: Same as read,

- Argument order of `write(2)` is `dest fd`, `source buf`, `read count`
- Argument order of `memcpy(3` is `out buf`, `in buf`, `read count`.

> size The number of bytes to write to the storage.

This can be "size of valid date in the `buffer`. How many bytes the custom write function actually read from `buffer` and write to whereever it want to write is up to the function.

The return value only represents `success` or `failed` (or `keep iterate` and `abort`). Thus, can be `bool`.

yashi: How about

```c
bool (*func)(const uint8_t *buffer, size_t size, off_t offset, void* user);
```

`func()` is called up on receiving a packet to let user know that `size` bytes of data is arrived and stored in `buffer`. the arrived data should be at `offset`.  The `user` pointer was assosicated to it so I'm passing it to you.

include/csp/csp_sfp.h
dimitrovand: I'm ok with any order or parameters. The usecase is to enable reading from a file or buffer at a specified offset, sparing users the need to implement this functionality themselves.

yashi:
Sorry for the delay. I’ll work on this over the weekend or in the coming week.

yashi:
Hi @dimitrovand, I'm back to work on this. Would you mind to fix the conflicts? Thanks.

dimitrovand:
Hi @yashi, the ABI Compliance Check is failing as expected due to the changes introduced.

yashi:
Hi @dimitrovand Thanks for the quick update!

Would it be technically possible to leave the old API intact and add new one?  Like provide new APIs and make the old API to warp your new API?

If not, I feel like we wait v2.1 tagged. I'll make sure to merge this PR right after v2.1 so that you don't have to update yet again.

dimitrovand:
@yashi not sure, but lets wait until v2.1

yashi:
@dimitrovand Sorry to make you wait. But v2.1 is now tagged and I'd like to merge this.

yashi:
I'd like to merge this PR first because I don't want to make PR author to rebase again. I've alraedy asked him to rebase too many times. @dimitrovand, the last push has a merge commit from the `develop` branch. That's not what we need.

yashi:
Thanks @dimitrovand!

PR #743, opened 2024-11-25
csp_conn: Prevent usage of RDP active connections

dimitrovand:
Add an additional check for RDP state to prevent connections with `rdp.state != RDP_CLOSED` beeing reallocated. This prevents premature reuse of connections that are still being closed by RDP.

### DIFF
--- a/src/csp_conn.c
+++ b/src/csp_conn.c
@@ -94,7 +94,11 @@ csp_conn_t * csp_conn_find_existing(csp_id_t * id) {
 		/* Connection must be open */
 		if (conn->state != CONN_OPEN)
 			continue;
-
+#if (CSP_USE_RDP)
+		/* Connection must not be closed by RDP */
+		if (conn->rdp.state == RDP_CLOSED)
+			continue;
+#endif
 		/**
 		 * This search looks verbose, Instead of a big if statement, it is written out as
 		 * conditions. This has been done for clarity. The least likely check is put first
--- a/src/csp_conn.c
+++ b/src/csp_conn.c
@@ -166,11 +166,16 @@ csp_conn_t * csp_conn_allocate(csp_conn_type_t type) {
 	for (int j = 0; j < CSP_CONN_MAX; j++) {
 		i = (i + 1) % CSP_CONN_MAX;
 
-		int expected = CONN_CLOSED;
-		if (atomic_compare_exchange_strong(&arr_conn[i].state, &expected, CONN_OPEN)) {
-			conn = &arr_conn[i];
-			csp_conn_last_given = i;
-			break;
+#if (CSP_USE_RDP)
+		if (arr_conn[i].rdp.state == RDP_CLOSED) 
+#endif
+		{
+			int expected = CONN_CLOSED;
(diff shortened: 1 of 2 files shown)


src/csp_conn.c
```diff
@@ -166,11 +166,16 @@ csp_conn_t * csp_conn_allocate(csp_conn_type_t type) {
 	for (int j = 0; j < CSP_CONN_MAX; j++) {
 		i = (i + 1) % CSP_CONN_MAX;
 
-		int expected = CONN_CLOSED;
-		if (atomic_compare_exchange_strong(&arr_conn[i].state, &expected, CONN_OPEN)) {
-			conn = &arr_conn[i];
-			csp_conn_last_given = i;
-			break;
+#if (CSP_USE_RDP)
+		if (arr_conn[i].rdp.state == RDP_CLOSED)
```
yashi, line 174: `arr_conn` is a global array of connections.  Are you sure this is thread safe? What happens after this check some other thread changed the state?

dimitrovand: If arr_conn[i].rdp.state == RDP_CLOSED and arr_conn[i].state == CONN_CLOSED the connection is fully closed and won't be used by another thread. If arr_conn[i].rdp.state != RDP_CLOSED then it won't be allocated.
The check is needed because connections can be in CONN_CLOSED state but not in RDP_CLOSED. If both are not in CLOSED state, that means it is still in use. Without this check you can allocate a connection which is not yet freed completely.
arr_conn[i].rdp.state is not thread safe so it might be a good idea to make it atomic.

yashi: > If arr_conn[i].rdp.state == RDP_CLOSED and arr_conn[i].state == CONN_CLOSED the connection is fully closed and won't be used by another thread.

- Thread A with lower priority
- Thraed B with higher priority

1. Thread A calls `csp_conn_allocate()`, and check `if (arr_conn[i].rdp.state == RDP_CLOSED)`
2. Scheduler switchs to Thread B
3. Thread B calls `csp_conn_allocate()`, do RDP things, and the connection is now `CONN_CLOSED` but `!= RDP_CLOSED`
4. Thread B is done
5. Thread A takes the CPU, and find the same `conn` with `!= RDP_CLOSED`.

> The check is needed because connections can be in CONN_CLOSED state but not in RDP_CLOSED.

This is the root cause, isn't it? Can we fix the root cause?

> it might be a good idea to make it atomic.

If the states of both `conn.state` and `conn.rdp.state` can change without the other, atomic doesn't help.

dimitrovand: Did some more testing and it turns out that the opposite happens in 'csp_conn_find_existing' where `conn.state` might be in `CONN_OPEN` but `conn.rdp.state` might be `RDP_CLOSED`. We can use only 1 state variable to achieve the current functionality, but i don't think its going to fix the whole synchronization issue by itself.

yashi: If we must keep two independent variables in sync, we have to use some kind of locks, aka. Mutex.

dimitrovand: I agree. Adding mutexes will trigger a lot of changes so probably it should be done in steps. I'm going to close this PR and we should just keep in mind this issue is currently present.

PR #744, opened 2024-11-25
csp_rdp: Fix invalid connection creation

dimitrovand:
- Update SYN flag check to only pass if the SYN flag is explicitly set, aligning with how RDP is documented in RFC 908. Previously, SYN|ACK packets could be misinterpreted as SYN packets, causing incorrect interpretation of packet data as connection parameters. This issue could lead to excessive packet and connection allocation durations.
- Improve the return value of `csp_rdp_rx_queue_add` by replacing raw integers with error codes defined in `csp_error.h` for better consistency.
- Modify the return value check for `csp_conn_enqueue_packet` to compare against `CSP_ERR_NONE`, aligning with its usage of standard error codes.

### DIFF
--- a/src/csp_rdp.c
+++ b/src/csp_rdp.c
@@ -497,8 +497,12 @@ bool csp_rdp_new_packet(csp_conn_t * conn, csp_packet_t * packet) {
 		 */
 		case RDP_CLOSED: {
 
+			/* Clear ephemeral data added by csp_rdp_send_cmp(). 
+			   RDP flags are located in the lower 4 bits. */
+			uint8_t rx_header_flags = rx_header->flags & 0x0f;
+
 			/* No SYN flag set while in closed. Inform by sending back RST */
-			if (!(rx_header->flags & RDP_SYN)) {
+			if (rx_header_flags != RDP_SYN) {
 				csp_rdp_protocol("RDP %p: Not SYN received in CLOSED state. Discarding packet\n", (void *)conn);
 				csp_rdp_send_cmp(conn, NULL, RDP_RST, conn->rdp.snd_nxt, conn->rdp.rcv_cur);
 				goto discard_close;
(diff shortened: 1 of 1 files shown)


yashi:
Please split to three commits. These are orthogonal to each other.

yashi (changes requested):
Please split to three commits. These are orthogonal to each other.

src/csp_rdp.c
```diff
@@ -497,8 +497,12 @@ bool csp_rdp_new_packet(csp_conn_t * conn, csp_packet_t * packet) {
 		 */
 		case RDP_CLOSED: {
 
+			/* Clear ephemeral data added by csp_rdp_send_cmp(). 
+			   RDP flags are located in the lower 4 bits. */
+			uint8_t rx_header_flags = rx_header->flags & 0x0f;
+
 			/* No SYN flag set while in closed. Inform by sending back RST */
-			if (!(rx_header->flags & RDP_SYN)) {
+			if (rx_header_flags != RDP_SYN) {
```
yashi, line 505: You said in the commit log:

> Update SYN flag check to only pass if the SYN flag is explicitly set,

But the change is, as you documented in the comment, to ignore the ephemeral data added by `csp_rdp_send_cmp()`, and make sure the flags has `RDP_SYN` and only `RDP_SYN`.  The word "explicitly" might not express your intention.

Given that, it might be better to say:

> Update SYN flag check to ensure only the RDP_SYN flag is set, aligning with the RDP specification in RFC 908.
The previous implementation allowed SYN|ACK packets to be misinterpreted as SYN packets, leading to incorrect handling of packet data as connection parameters.
The updated check masks out the upper bits of the flags field and requires an exact match with RDP_SYN.
This change prevents unintended connection creation, reducing excessive packet and connection allocation durations.

Also you might want to reference "5.1  Connection Establishment" in the RFC because in other part of the RFC, it reads:

```
       If SYN set
```

Which feels like the code does exactly the RFC says. (but wrong).

BTW, we have so many `rx_header->flags & RDP_*` checks in our code base, do we have to fix other places as well?

yashi (commented):
Just curious, we have so many `rx_header->flags & RDP_*` check in our code base. Do we need to fix them all?

src/csp_rdp.c
dimitrovand: Not sure if there are other places.

yashi (approved):
Thanks. LGTM.

PR #746, opened 2024-11-27
Fix HMAC

yashi:
I've found a bug in HMAC code. I've added a unit test and sample code for HMAC. Please review.


yashi:
@poomravint Would you mind taking a look at it?

PR #760, opened 2024-12-04
cmake: Add support for FreeRTOS

dimitrovand:
Add build support for FreeRTOS in CMake.
The CSP_KERNEL_NAME variable is introduced to specify the kernel name since CMAKE_SYSTEM_NAME does not recognize FreeRTOS. This approach ensures flexibility for adding support for new kernels in the future.


yashi:
> This approach ensures flexibility for adding support for new kernels in the future.

@dimitrovand, sorry for my ignorance but does FreeRTOS has multiple kernels?

Or do you mean other RTOSes / kernels / OSes?

If you are referring other FreeRTOS kernels, the variable should be `CSP_FREERTOS_KERNEL_NAME`.

If you are referring other RTOSes, why not just set `CMAKE_SYSTEM_NAME`? Currently we set `CMAKE_SYSTEM_NAME` to `Zephyr` at the Zephyr's top level CMakeLists.txt:  https://github.com/libcsp/libcsp/blob/6de44fbc7f8842e006820da6dfb1be7a3ca0c2a3/contrib/zephyr/CMakeLists.txt#L2

We haven't have any problem setting a system name to `CMAKE_SYSTEM_NAME`. But do you have any consideration?

I found that a few references to `CMAKE_SYSTEM_NAME`  in FreeRTOS repository:
```
❯ ag CMAKE_SYSTEM_NAME
FreeRTOS/Demo/CORTEX_MPU_R5F_TI_RM57_HERCULES_GCC/CMakeLists.txt
15:SET(CMAKE_SYSTEM_NAME "Generic" CACHE STRING "Target system is a generic ARM Processor")

FreeRTOS/Demo/CORTEX_No_GIC_R5F_TI_RM57_HERCULES_GCC/CMakeLists.txt
15:SET(CMAKE_SYSTEM_NAME "Generic" CACHE STRING "Target system is a generic ARM Processor")

FreeRTOS/Demo/CORTEX_MPU_R4F_TI_RM46_HERCULES_GCC/CMakeLists.txt
15:SET(CMAKE_SYSTEM_NAME "Generic" CACHE STRING "Target system is a generic ARM Processor")

FreeRTOS-Plus/ThirdParty/mbedtls/library/CMakeLists.txt
206:if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")

FreeRTOS-Plus/ThirdParty/mbedtls/programs/test/CMakeLists.txt
31:   NOT ${CMAKE_SYSTEM_NAME} MATCHES "[Ww][Ii][Nn]")

FreeRTOS-Plus/ThirdParty/wolfSSL/CMakeLists.txt
43:#set(CMAKE_SYSTEM_NAME Linux)
```

Here is a search result of `set(CMAKE_SYSTEM_NAME FreeRTOS)` on GitHub.
https://github.com/search?q=%2Fset%5C%28CMAKE_SYSTEM_NAME+.*FreeRTOS%5C%29%2F&type=code

dimitrovand:
I'm referring to other kernels and operating systems. Using `CMAKE_SYSTEM_NAME` within `CSP` introduces challenges. It requires matching every possible `CMAKE_SYSTEM_NAME` value, which is often set by the top-level project and can be any string. Forcing the top-level project to set `CMAKE_SYSTEM_NAME` to a specific value isn't ideal. A clear example from this PR is the inconsistency when setting `CMAKE_SYSTEM_NAME` to `Generic`. When trying to integrate with one of the example projects, where `CMAKE_SYSTEM_NAME` is set to `FreeRTOS`, it won't work very well. Even if we set `CMAKE_SYSTEM_NAME` to `FreeRTOS` what stops users from integrating `CSP` into a project which sets `CMAKE_SYSTEM_NAME` to `Generic` or any other value meaningful to them.

dimitrovand:
Another example is when `CMAKE_SYSTEM_NAME` is required by `CSP` and another library, and they each expect different values.

yashi:
If we depend on `CMAKE_SYSTEM_NAME="Genric"` except for Linux, the PR isn't sufficient. We need to depend on top level `CMakeLists.txt` to set some variable(s). In your case, that's `CSP_KERNEL_NAME`.  We already have `CSP_FREERTOS`, `CSP_ZEPHYR`, and `CSP_POSIX`.

For Linux, I don't think users will set `CMAKE_SYSTEM_NAME` explicitly, but CMake will detect it.

For Zephyr and FreeRTOS, we must ask users to set some in top level `CMakeLists.txt`. What we want to do after all is to make conditions depends on the variable for various part of CMake configuration.

So, IMHO, `libcsp/CMakeLists.txt` will have:

```cmake
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
  set(CSP_POSIX 1)
  # and / or
  # set(CSP_KERNEL_NAME "Linux")
elseif(CMAKE_SYSTEM_NAME STREQUAL "Generic")
  if(CSP_KERNEL_NAME "FreeRTOS")
    set(CSP_FREERTOS 1)
  elseif(CSP_KERNEL_NAME "Zephyr")
    set(CSP_ZEPHYR 1)
  else()
    message(FATAL_ERROR "Unsupported CSP_KERNEL_NAME: ${CSP_KERNEL_NAME}")
  endif()
else()
  message(FATAL_ERROR "Unsupported CMAKE_SYSTEM_NAME: ${CMAKE_SYSTEM_NAME}")
endif()
```

Overall, it looks good and I can go with that. But I see a few issues:

1. The variable name with "KERNEL" is kind of misnominor. Do we really want to make conditions on kernels? We don't even care about OS but using "POSIX" right now.  Sure we "could" have Debian with Hurd or FreeBSD kernel but I double we want to do that right now.
2. What is the point of having both `CSP_KERNEL_NAME` and a set of `CSP_FREERTOS`, `CSP_POSIX` and `CSP_ZEPHYR` if what we want to do is to make conditions. We can either have a multi-values variable like `CSP_KERNEL_NAME`, or a set of variables. Or, some downstreams might already be using `CSP_FREERTOS` and `CSP_POSIX`. So should we keep them?

We have `arch/` but "Architecture" IMHO means CPU Architectures. FreeRTOS and Zephyr is a RTOS. Linux is acually a kernel, but these days nobody care even if we say "Linux is an OS". So, why not use "OS" for that? Or, follow CMake way and use `SYSTEM`?

- `CSP_KERNEL_NAME`
- `CSP_OS_NAME`
- `CSP_ARCH_NAME`
- `CSP_SYSTEM_NAME`
- `CSP_TARGET_OS`
- `CSP_TARGET_KERNEL`
- `CSP_TARGET_ARCH`
- `CSP_TARGET_SYSTEM`
- what else?

yashi:
Just to clearify. I'm not saying you need to port Zephyr in this PR. I can take care of that after merging this PR.

I want to agree on things I posted before merging it.

dimitrovand:
I was thinking of removing completely the usage of `CMAKE_SYSTEM_NAME`.
In top level `CMakeLists.txt` use something like this (note that CSP_PORT is just a name, I don't really know what is best suited for the purpose):
```
if(NOT DEFINED CSP_PORT)
  set(CSP_PORT "linux")
endif()

if(CSP_PORT STREQUAL "linux")
  set(CSP_POSIX 1)
elseif(CSP_PORT STREQUAL "zephyr")
  set(CSP_ZEPHYR 1)
elseif(CSP_PORT STREQUAL "freertos")
  set(CSP_FREERTOS 1)
else()
  message(FATAL_ERROR "Port not supported ${CSP_PORT}")
endif()
```
And in `arch/CMakeLists.txt` use something like:
`add_subdirectory(${CSP_PORT})`

And we won't have to think/deal with this ever again.

yashi:
> I was thinking of removing completely the usage of `CMAKE_SYSTEM_NAME`. 

Why? It's a standard variable in CMake.

In top level `CMakeLists.txt` use something like this:
> 
> ```cmake
> if(NOT DEFINED CSP_PORT)
>   set(CSP_PORT "linux")
> endif()
> ```

This means that a wrong setting will build Linux. I don't like that.

> ```cmake
> if(CSP_PORT STREQUAL "linux")
>   set(CSP_POSIX 1)
> elseif(CSP_PORT STREQUAL "zephyr")
>   set(CSP_ZEPHYR 1)
> elseif(CSP_PORT STREQUAL "freertos")
>   set(CSP_FREERTOS 1)
> else()
>   message(FATAL_ERROR "Port not supported ${CSP_PORT}")
> endif()
> ```
> 
> And in `arch/CMakeLists.txt` use something like: `add_subdirectory(${CSP_PORT})`

We also have `src/drivers/CMakeLists.txt`.

> And we won't have to think/deal with this ever again.

ha ha. we wish.

> (note that CSP_PORT is just a name, I don't really know what is best suited for the purpose)

Naming is hard.

dimitrovand:
I’m not certain what the best approach is, but FreeRTOS handles its various ports (compilers, architectures, operating systems, etc.) in a very effective/simple way. From what I can observe, the current management of CSP ports could be improved, and it might be beneficial to consider reusing their approach to address a similar problem.
Let me know what changes you’d like to see in this PR once you decide how you’d like to proceed.

yashi:
Let's use `CSP_SYSTEM_NAME` instead of `CSP_KERNEL_NAME` to resemble CMake's `CMAKE_SYSTEM_NAME`. Nothing else needs to be changed. I'll work on Zephyr size.

yashi:
@dimitrovand you still have one issue to fix.

yashi (approved):
LGTM

PR #769, opened 2024-12-26
Add missing `src/interfaces/csp_if_tun.c` in wscript file.

wookhyunshin:
While I install libcsp as shared library, I found that `csp_if_tun.c` is missing in script file.


yashi:
Hi @wookhyunshin,

Thank you for the PR! I've updated the commit log and merged. Thanks!

PR #786, opened 2025-01-24
python: Enhance Python Bindings and Examples for libcsp

moonlight83340:
This pull request introduces several improvements and fixes to the Python bindings, examples, and GitHub workflows for libcsp. Below is a summary of the changes made:

**Examples Improvements**:
- Added a test option to the client-server example, allowing users to run it in test mode for a specified duration.
- Ensure proper thread termination on program exit
- Fixed a typo in a `printf` statement in the Python examples ("Recieved" -> "Received").
- Removed unused imports in the Python example files for better code cleanliness.
- `route_start_task()` was  marked with `METH_VARARGS`, even though it didn't handle arguments. So change it to `METH_NOARGS `.

**GitHub Workflow Update**:
- Added a new loopback test for the Python bindings in the GitHub Actions workflow to enhance automated testing coverage.

---

If you prefer, I can split these changes into smaller, focused pull requests to make the review process easier. Let me know!

### DIFF
--- a/examples/csp_server_client.py
+++ b/examples/csp_server_client.py
@@ -4,7 +4,7 @@ Usage: LD_LIBRARY_PATH=build PYTHONPATH=build python3 ./examples/csp_server_clie
 import time
 import threading
 import libcsp_py3 as csp
-from typing import Any, Callable
+from typing import Callable
 
 
 def printer(node: str, color: str) -> Callable:
--- a/examples/csp_server_client.py
+++ b/examples/csp_server_client.py
@@ -32,7 +32,7 @@ def server_task(addr: int, port: int) -> None:
 
         while (packet := csp.read(conn, 50)) is not None:
             if csp.conn_dport(conn) == port:
-                _print('Recieved on {port}: {data}'.format(
+                _print('Received on {port}: {data}'.format(
                     port=port,
                     data=csp.packet_get_data(packet).decode('utf-8'))
                 )
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -963,7 +963,7 @@ static PyMethodDef methods[] = {
 	{"conn_src", pycsp_conn_src, METH_O, ""},
 	{"listen", pycsp_listen, METH_VARARGS, ""},
 	{"bind", pycsp_bind, METH_VARARGS, ""},
-	{"route_start_task", pycsp_route_start_task, METH_VARARGS, ""},
+	{"route_start_task", pycsp_route_start_task, METH_NOARGS, ""},
 	{"ping", pycsp_ping, METH_VARARGS, ""},
 	{"reboot", pycsp_reboot, METH_VARARGS, ""},
 	{"shutdown", pycsp_shutdown, METH_VARARGS, ""},
--- a/examples/python_bindings_example_client.py
+++ b/examples/python_bindings_example_client.py
@@ -10,7 +10,6 @@
 # $ LD_LIBRARY_PATH=build PYTHONPATH=build python3 examples/python_bindings_example_client.py -z localhost
 #
 
-import os
 import time
 import sys
 import argparse
--- a/examples/python_bindings_example_client.py
+++ b/examples/python_bindings_example_client.py
@@ -62,7 +62,6 @@ if __name__ == "__main__":
         # same format/use as line above
         libcsp.rtable_load(options.routing_table)
 
-    # Parameters: {priority} - 0 (critical), 1 (high), 2 (norm), 3 (low) ---- default=2
     # Start the router task - creates routing thread
     libcsp.route_start_task()
     time.sleep(0.2)  # allow router task startup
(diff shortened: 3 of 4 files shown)


src/bindings/python/pycsp.c
```diff
@@ -987,7 +987,7 @@ static PyMethodDef methods[] = {
 
 	/* csp/csp_buffer.h */
 	{"buffer_free", pycsp_buffer_free, METH_VARARGS, ""},
-	{"buffer_get", pycsp_buffer_get, METH_VARARGS, ""},
+	{"buffer_get", pycsp_buffer_get, METH_NOARGS, ""},
```
yashi: This breaks the existing python code, no?

examples/csp_server_client.py
yashi, line 1: This has too many changed than adding the test option. Please rework on it.

examples/csp_server_client.py
```diff
@@ -51,7 +51,7 @@ def server_task(addr: int, port: int) -> None:
 
         while (packet := csp.read(conn, 50)) is not None:
             if csp.conn_dport(conn) == port:
-                _print('Recieved on {port}: {data}'.format(
+                _print('Received on {port}: {data}'.format(
```
yashi, line 35: It's not technically in the function body of `printf()`.  But the message has a typo.

yashi (changes requested):

src/bindings/python/pycsp.c
moonlight83340: Yes, I remove it.

examples/csp_server_client.py
moonlight83340: I reworked and now there are 2 commits for this work.

examples/csp_server_client.py
moonlight83340: Yes I change my message.

moonlight83340:
workflow error are : 

[EOF Rule] File "src/arch/freertos/csp_hooks.c" has lint errors
[EOF Rule] File "src/interfaces/csp_if_tun.c" has lint errors

moonlight83340:
> workflow error are :
> 
> [EOF Rule] File "src/arch/freertos/csp_hooks.c" has lint errors [EOF Rule] File "src/interfaces/csp_if_tun.c" has lint errors

Could be fix by https://github.com/libcsp/libcsp/pull/791

examples/csp_server_client.py
yashi, line 1: Would you please wite a bit more commit log explaining why you need these changes?

examples/csp_server_client.py
```diff
@@ -3,9 +3,25 @@
 """
 import time
 import threading
+import sys
+import argparse
 import libcsp_py3 as csp
 from typing import Any, Callable
 
+server_received = 0
+server_received_lock = threading.Lock()
```
yashi: do you need to lock it in order to add options to this?

examples/csp_server_client.py
yashi, line 1: I'm still not sure why you need this.  The main thread knows both child threads. Why do you need to get an event? Plus you are using event as a normal variable. What's the point of using Event?

yashi, line 1: I can merge this.

yashi, line 1: I can merge this.

src/bindings/python/pycsp.c
```diff
@@ -963,7 +963,7 @@ static PyMethodDef methods[] = {
 	{"conn_src", pycsp_conn_src, METH_O, ""},
 	{"listen", pycsp_listen, METH_VARARGS, ""},
 	{"bind", pycsp_bind, METH_VARARGS, ""},
-	{"route_start_task", pycsp_route_start_task, METH_VARARGS, ""},
+	{"route_start_task", pycsp_route_start_task, METH_NOARGS, ""},
```
yashi, line 966: These also breaks the current python code, no?

yashi (changes requested):

examples/csp_server_client.py
```diff
@@ -68,18 +86,35 @@ def client_task(addr: int, port: int) -> None:
 
 
 def main() -> None:
+    global server_received
```
yashi: https://docs.python.org/3/reference/simple_stmts.html#index-43

> free variables may refer to globals without being declared global.

yashi (changes requested):

yashi:
I came to the conclusion that `csp_server_client.py` need not to be sync with the C version nor run as a test.  As noted in the initial commit of the file, this one was intended to be a simple example.  Rather than having this, it'd be better to have simple python samples for users telling how to use them.

The basic functionarity of the libcsp is tested by the tests under `unittests/` and other files under `examples/`.  This python conde does NOT add any test case except python bindings, which should be tested in other ways.

So if you agree, please drop any change to `csp_server_client.py` from this PR.

moonlight83340:
Rebase + drop `csp_server_client.py` workflow + big change.
Keep only fix typo and remove unused imports.

yashi (approved):

PR #789, opened 2025-01-31
contrib: zephyr: avoid csp_buffer_get_always()

moonlight83340:
PR update:
The sample now uses csp_buffer_get(0) instead of csp_buffer_get_always(),
which is not recommended in application-level code.


yashi (approved):

yashi (changes requested):
Actually I changed my mind. `csp_buffer_get_always()` should be called from application layer.  The function must be used in and only in the critical locations.

PR #790, opened 2025-01-31
interfaces: csp_if_tun: Remove check NULL for packet

moonlight83340:
Since csp_buffer_get_always() is used to get the csp packet, it should never return NULL but instead panic so we don't need to check if NULL is returned.


yashi (approved):

yashi:
As noted in #864, we should never use `csp_buffer_get_always()` in non-critical path.  Instead of removing check, please replace all `csp_buffer_get_always()` from non-critical path.  I'm closing this PR in favor of it.

PR #793, opened 2025-02-07
Sphinx docs build fix

IlievIliya92:
This PR fixes the Sphinx Documentation build issues related to the different libclang versions installed by the libclang-dev package on different Ubuntu releases. It provides support for building the documentation under Ubuntu 22.04 and the newer Ubuntu release - 24.04. It also updates the Docker image for the documentation build to use `ubuntu:latest` (Ubuntu 24.04).

### DIFF
--- a/doc/Dockerfile
+++ b/doc/Dockerfile
@@ -1,11 +1,11 @@
-FROM ubuntu:22.04
+FROM ubuntu:latest
 ARG DEBIAN_FRONTEND=noninteractive
 MAINTAINER "Iliya Iliev"
 
 # Install dependencies
 RUN apt-get update && \
 	apt-get install --no-install-recommends --no-install-suggests -y sudo git python3-pip cmake \
-	libclang-14-dev build-essential locales locales-all
+	libclang-dev build-essential locales locales-all
 
 ENV LC_ALL en_US.UTF-8
 ENV LANG en_US.utf8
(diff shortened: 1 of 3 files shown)


doc/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM ubuntu:22.04
+FROM ubuntu:24.04
```
moonlight83340: You're using the 24.04 version strictly, could you pass the ubuntu version as an argument by the workflow ? it could fix it even if the lastest version is changing ?

IlievIliya92: The Docker image is not used directly by the github workflow. It has been added for convenience to use it for local builds of the documentation. We can also use `ubuntu-latest` as base for the Docker image of the docs if this is what you are referring to ?

moonlight83340 (commented):
LGTM! Just a small suggestion: I think it's better to always include a commit body, even for small changes, rather than just a summary. It makes the history clearer.

Ref: https://github.com/libcsp/libcsp/blob/develop/doc/git-commit.md
(PS: I'm not a maintainer, just my two cents.)

yashi:
Hi @IlievIliya92 Thaks for the PR. Would you please add commit message to the 2nd and 3rd commits? Thanks.

yashi (approved):
Super! Thanks!

PR #796, opened 2025-02-12
Reset rtable_inptr in csp_rtable_free() to avoid null pointer dereferencing

clorch:
If csp_rtable_free() is called, the rtable_inptr is not reset to 0.

 If later the rtable is used again, this can lead to null pointer dereferencing in other parts of the code.


yashi:
Hi @clorch,

Thanks for the PR! It's LGTM except that your commit message is too long.
Please fold at or less than 80.

Thanks.

clorch:
Hi @yashi,

I tried to change the commit message. I hope it worked, to me it looks better now.

yashi (approved):
Thanks!

PR #797, opened 2025-02-12
service: peek/poke v2 with 64 bit memory support

troelsjessen:


johandc:
This is a duplicate of an already pulled request

PR #799, opened 2025-02-12
Fixed compilation in release mode

troelsjessen:
Fails due to warning about unused variable

Do we have another consolidated solution to this topic? I presume it cannot be the first time we have such a situation.


yashi:
You mean #806?

troelsjessen:
Yes, the approach look very similar to #806, that is good!
This PR can be merged along with #806, as that one does not include this specific file handled in this PR, likely due to the optional inclusion of the ZMQ interface.

yashi:
I'm merging this for now but `csp_if_zmqhub.c` really needs a lot of work.

PR #800, opened 2025-02-13
Zephyr: Use C linkage for C++ builds

surligas:
Zephyr CAN driver fails to build for Zephyr projects using C++. This commit adds C linkage in the Zephyr CAN driver in case C++ is used.


yashi (approved):
Ah great! Thanks!

PR #804, opened 2025-03-05
build: warn about and fix missing prototypes

6arms1leg:
This PR turns on `-Wmissing-prototypes` for all builds and fixes all occurrences to improve code safety.
Be less strict about the Python C extension module since it is not flight code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,7 @@ if(CSP_POSIX)
     -Wcast-align
     -Werror
     -Wextra
+    -Wmissing-prototypes
     -Wpedantic
     -Wpointer-arith
     -Wshadow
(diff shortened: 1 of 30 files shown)


6arms1leg:
The API compatibility check fails.  This is to be expected as this change set makes "private" functions `static`.
I assumed those functions (15 in total) to be "private" because they are not exposed in the API (header files) and not used outside their modules (implementation files).
So IMO, the PR is clean and the failing test should be ignored.  Let me know if you want anything changed, though.

yashi:
LGTM.  Would you please fix the conflicts?

6arms1leg:
> LGTM. Would you please fix the conflicts?

Thanks for the review!
I rebased and resolved the conflicts.

yashi:
Thanks. Merged.

PR #807, opened 2025-03-09
build: make shared lib. build optional for CMake/Linux

6arms1leg:
This PR adds an option to the CMake build process to build a static library for GNU/Linux instead of the default shared library.

I have an embedded project that links libCSP as a static library from a Git submodule.  However, I also build that project for GNU/Linux for easy integration/end-to-end/functional testing purposes.  Without an option to disable the shared library build on GNU/Linux, I currently need to manually modify the libCSP submodule.
It would be great if this was possible without the need to modify libCSP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,9 @@ cmake_minimum_required(VERSION 3.20)
 project(CSP VERSION 2.1)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-  set(BUILD_SHARED_LIBS ON)
+  set(DEFAULT_BUILD_SHARED_LIBS ON)
 endif()
+option(BUILD_SHARED_LIBS "Build using shared libraries" ${DEFAULT_BUILD_SHARED_LIBS})
 
 add_library(csp)
 set_target_properties(csp PROPERTIES C_STANDARD 11)


yashi:
Hi @6arms1leg Thanks for the PR.

Just curious, why don't we build both by default? does it make sense?

6arms1leg:
Hello @yashi, thanks for your review!
Sure, that would be another alternative.  Do you prefer it that way?
I thought that usually you just want one or the other, not both.

yashi:
@6arms1leg It only takes a fraction of a second to generate shared object in addition to generate a static library.  And we can control which one to link. So, let's generate both.

6arms1leg:
> @6arms1leg It only takes a fraction of a second to generate shared object in addition to generate a static library. And we can control which one to link. So, let's generate both.

I just had a look at this again and it's not as straight forward as it initially seems.
We cannot simply compile the object files once and link them into a static and shared library, respectively.
The one main reason for that is PIC.  See [here](https://stackoverflow.com/questions/2152077/is-it-possible-to-get-cmake-to-build-both-a-static-and-shared-library-at-the-sam).

IMO, the original PR therefore still implements the best solution, so I just rebased it.

yashi:
OK.

Then, how about like this:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 06abab70..4e4a94d2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.20)
 project(CSP VERSION 2.1)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-  set(BUILD_SHARED_LIBS ON)
+  set(DEFAULT_BUILD_SHARED_LIBS ON)
 endif()
 
 add_library(csp)
@@ -15,8 +15,9 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE MinSizeRel CACHE STRING "Build Type" FORCE)
 endif()
 
-option(CSP_REPRODUCIBLE_BUILDS "Reproducible build" OFF)
+option(BUILD_SHARED_LIBS "Build using shared libraries" ${DEFAULT_BUILD_SHARED_LIBS})
 
+option(CSP_REPRODUCIBLE_BUILDS "Reproducible build" OFF)
 option(CSP_HAVE_STDIO "OS provides C Standard I/O functions" ON)
 option(CSP_ENABLE_CSP_PRINT "Enable csp_print() function" ON)
 option(CSP_PRINT_STDIO "Use vprintf() for csp_print() function" ON)
```

The base idea is the same; Let user decide what to build.

But I use the standard `BUILD_SHARED_LIBS` as the option instead of our own `CSP_BUILD_SHARED_LIB_LINUX` so that users can do either:

```
cmake -B builddir -DBUILD_SHARED_LIBS=1
```

or

```
cmake -B builddir -DBUILD_SHARED_LIBS=0
```

I've moved the `option()` down to where others are. That makes it look clean.  `add_library()` doesn't care about where the `option()`s are set.

6arms1leg:
Sure.  Either way is fine.
The only functional difference I see is that your suggestion allows to build the static library for any target, not just GNU/Linux.
I'm not sure how useful that is for embedded targets but I guess it's good to have the option.

I made the changes you requested.

yashi:
My bad. [CMake document](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) says:

> If adding such an option to the project, do so in the top level `CMakeLists.txt` file, before any [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) calls.

You need to move `option()` before the `add_library()`:


```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 86dff82c..8a3ed3b9 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,9 @@ cmake_minimum_required(VERSION 3.20)
 project(CSP VERSION 2.1)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
-  set(BUILD_SHARED_LIBS ON)
+  set(DEFAULT_BUILD_SHARED_LIBS ON)
 endif()
+option(BUILD_SHARED_LIBS "Build using shared libraries" ${DEFAULT_BUILD_SHARED_LIBS})
 
 add_library(csp)
 set_target_properties(csp PROPERTIES C_STANDARD 11)
```

When I suggested the change in my other comment, I had only tested it with the variable explicitly set, which worked fine. However, leaving it implicit doesn’t work, as the CI build revealed.

yashi:
> The only functional difference I see is that your suggestion allows to build the static library for any target, not just GNU/Linux.
> I'm not sure how useful that is for embedded targets

That’s true, but I see no point in prohibiting it. If a user wants to build a shared library for their target board, I’d say: let them do it. :wink:

6arms1leg:
Alright, done.

yashi:
Thanks!

PR #811, opened 2025-03-10
workflow: codespell: Add codespell to the github workflow

moonlight83340:
Add GitHub Actions workflow for spelling check using Codespell.

Set up a workflow to automatically check for spelling mistakes in the codebase
using Codespell.

Configured Codespell to skip waf files.

Uses quiet-level=2 to minimize false positives and ensure only critical spelling
mistakes are reported.

Created a `.codespell-ignore` file and added "re-use" to the ignore list to
prevent false positives.

The word "re-use" is valid and should not be corrected to "reuse".

Added it to the ignore list to prevent false positives.

Typo fix are in another PR : https://github.com/libcsp/libcsp/pull/810
Merging the PR with typo fix will fix the current failing of the codespell workflow here.


yashi:
why did you add `re-use` to be ignored? How did you interpreted the following text?

> @return Packet (free with csp_buffer_free() or re-use packet), NULL on error or timeout.

moonlight83340:
> why did you add `re-use` to be ignored? How did you interpreted the following text?
> 
> > @return Packet (free with csp_buffer_free() or re-use packet), NULL on error or timeout.

We can either free or reuse the readed packet.
https://en.wikipedia.org/wiki/Reuse
https://forum.wordreference.com/threads/re-use-vs-reuse-noun-form.2085633/

yashi:
I'm sorry, I don't get what you mean. Why did you put "re-use" in the ignore?

moonlight83340:
"Re-use" is a commonly accepted alternative spelling of "reuse", so I added it to the ignore list to avoid false positives from codespell.

yashi:
Ok, in that case I accept it. Thank you for elaborating me.

PR #812, opened 2025-03-11
interfaces: udp: Improve UDP socket creation error handling and bind error handling

moonlight83340:
Ensure `socket()` failure is properly detected before proceeding
Change loop condition from `ifconf->sockfd == 0` to `ifconf->sockfd < 0` to handle properly failing `socket()` calls.
Update error message to clarify socket creation failure.
Avoid using an invalid `sockfd` value in subsequent operations.
Close `sockfd` and reset it to `-1` if `bind()` fails.
Add retry mechanism with a delay when `bind()` fails.
Improve error messages for better debugging.

### DIFF
--- a/src/interfaces/csp_if_udp.c
+++ b/src/interfaces/csp_if_udp.c
@@ -69,9 +69,15 @@ void * csp_if_udp_rx_loop(void * param) {
 	csp_iface_t * iface = param;
 	csp_if_udp_conf_t * ifconf = iface->driver_data;
 
-	while (ifconf->sockfd == 0) {
+	do {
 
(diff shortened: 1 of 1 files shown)


yashi:
In fact, the loop should be remove.  "loop until we get a socket" isn't proper handling.

yashi (changes requested):
In fact, the loop should be remove. "loop until we get a socket" isn't proper handling.

moonlight83340:
That make sense, even more in space usage program.
I can take care this from this PR, if you want or I can close it and make a new one.

yashi:
Let's close this. Open a new issue with a new implementation. you can always reference this from the new one.

PR #813, opened 2025-03-13
samples: udp: Add simple sample for UDP

moonlight83340:
This PR aim to : 
- Add simple send UDP sample
- Add UDP interface support to csp_server example
- Add UDP interface support to csp_client example
- Add UDP interface testing on workflow

PS: I added `wait $!` to wait for the background program because the background program does not exit immediately. During testing, we keep the background program running for 10 seconds while the foreground program runs for 3 seconds. So, when the foreground program finishes, the next test starts while the background program is still running.

I noticed this behavior specifically with UDP, as a bind failure causes the test to fail in the workflow, whereas other interfaces do not have the same issue. However, I was able to confirm that the background program kept running for those interfaces as well, even though we would expect it to terminate. This is something that should be investigated.

I think we should think about waiting the background program end.

### DIFF
--- a/samples/posix/simple-send-udp/src/main.c
+++ b/samples/posix/simple-send-udp/src/main.c
@@ -1,0 +1,60 @@
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+
+#include <csp/csp.h>
+#include <csp/interfaces/csp_if_udp.h>
+
+#define CLIENT_ADDR 2
+
+#define SERVER_ADDR 1
+#define SERVER_PORT 10
+
+#define DEFAULT_UDP_ADDRESS     "127.0.0.1"
+#define DEFAULT_UDP_REMOTE_PORT 1501
+#define DEFAULT_UDP_LOCAL_PORT  1500
+
+int main(int argc, char * argv[]) {
+	csp_conn_t * conn;
+	csp_packet_t * packet;
+	int ret;
+
+	/* init */
+	csp_init();
+
+	/* Interface config */
+	csp_iface_t iface;
+	csp_if_udp_conf_t conf = {
+		.host = DEFAULT_UDP_ADDRESS,
+		.lport = DEFAULT_UDP_LOCAL_PORT,
+		.rport = DEFAULT_UDP_REMOTE_PORT};
+
+	csp_if_udp_init(&iface, &conf);
+	iface.is_default = 1;
+	iface.addr = CLIENT_ADDR;
+
+	/* connect */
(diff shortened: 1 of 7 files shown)


samples/posix/simple-send-udp/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+
+#include <csp/csp.h>
+#include <csp/interfaces/csp_if_udp.h>
+
+#define CLIENT_ADDR 2
+
+#define SERVER_ADDR 1
+#define SERVER_PORT 10
+
+#define DEFAULT_UDP_ADDRESS     "0.0.0.0"
```
yashi: Why 0.0.0.0?

samples/posix/simple-send-udp/src/main.c
```diff
@@ -0,0 +1,55 @@
...
+
+	/* connect */
+	conn = csp_connect(CSP_PRIO_NORM, SERVER_ADDR, SERVER_PORT, 1000, CSP_O_NONE);
+	if (conn == NULL) {
+		csp_print("Connection failed\n");
+		return 1;
+	}
+
+	/* prepare data */
+	packet = csp_buffer_get_always();
```
yashi: Please do not use `csp_buffer_get_always()` in samples. See #864.

yashi (changes requested):

samples/posix/simple-send-udp/src/main.c
moonlight83340: Test purpose, I tought I wrote `127.0.0.1` in the first place.

samples/posix/simple-send-udp/src/main.c
```diff
@@ -0,0 +1,60 @@
...
+
+	/* prepare data */
+	packet = csp_buffer_get(0);
+	if (packet == NULL) {
+		csp_print("Failed to get buffer\n");
+		csp_close(conn);
+		return 1;
+	}
+	memcpy(packet->data, "abc", 3);
+	packet->length = 3;
```
yashi, line 51: Does our server adds `\n` when prints? Mine has:

```diff
+	/* prepare data */
+	packet = csp_buffer_get_always();
+	csp_id_prepend(packet);
+	memcpy(packet->data, "abc\n", 4);
+	packet->length += 4;
+	packet->frame_length += 4;
```

yashi: Other samples only have `"abc"` with length of 3. hmm..

PR #814, opened 2025-03-18
interfaces: udp: Add return error codes with csp_if_udp_init

moonlight83340:
Refactored `csp_if_udp_init` to return an error code instead of `void`, allowing better error handling.

- Updated function signature to return `int`.
- Added return values for error cases:
  - `CSP_ERR_INVAL` if the peer address is invalid.
  - `CSP_ERR_DRIVER` if thread creation or attribute initialization fails.
- Improved documentation for `csp_if_udp_init` due to function signature change.
- Adjusted debug messages for consistency.

This change improves robustness by making errors explicit and enabling calling functions to handle failures properly.

### DIFF
--- a/include/csp/interfaces/csp_if_udp.h
+++ b/include/csp/interfaces/csp_if_udp.h
@@ -29,6 +29,8 @@ typedef struct {
 } csp_if_udp_conf_t;
 
 /**
+ * @brief Initialize a UDP interface for CSP.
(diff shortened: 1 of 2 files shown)


yashi:
This must be after v2.1 because of API change. If you want to merge non-API change before v2.1, please create another PR.

PR #818, opened 2025-03-25
doc: Clarify ETH driver compatibility with Linux

moonlight83340:
Updated the documentation to replace "Posix ETH driver" with "Linux ETH driver" since the driver is currently only usable on Linux.

The previous name suggested POSIX compatibility, which could be misleading.

### DIFF
--- a/doc/api/drivers/eth_linux_h.rst
+++ b/doc/api/drivers/eth_linux_h.rst
@@ -1,4 +1,4 @@
-Posix ETH driver
+Linux ETH driver
 ================
 
 .. autocmodule:: drivers/eth_linux.h
--- a/include/csp/drivers/eth_linux.h
+++ b/include/csp/drivers/eth_linux.h
@@ -1,7 +1,7 @@
 /****************************************************************************
  * **File:** csp/drivers/eth_linux.h
  *
- * **Description:** Posix ETH driver
+ * **Description:** Linux ETH driver
  *
  * .. note:: Using this driver require user elevation. Guideline for doing this
  *           is given at run-time


yashi:
Are you sure you tested against other POSIX OSes? Which one did you try? and tell me why it doesn't work with them.

moonlight83340:
While writing a Zephyr driver for the Ethernet interface, I tried to reuse the same file using Zephyr's POSIX layer. 
However, I found that some definitions and functions depend on Linux-specific headers.

So even though it may seem POSIX-compatible at first glance, the current implementation actually relies on Linux-specific interfaces, which prevents it from being used as a generic POSIX driver.

The code uses several Linux-specific headers and system calls, such as `<linux/if_packet.h>`, `AF_PACKET`, `struct sockaddr_ll`, and `SO_BINDTODEVICE`.
These elements are not defined by POSIX and are therefore not portable to other POSIX-compliant systems.

yashi:
But other than those pointed out, we can reuse the code, no? The goal is not creating pure-POSIX drivers. We are more pragmatic. If MacOS or NuttX is ported we could reuse it by refarctoring the code.  Or, maybe not. It might not worth poring it to other POSIX systems. I don't know. But that's our intention.

yashi:
A bad thing about the current form is that the header is named "linux".  In C, especially with portable or muti-platform library, we should use non arch specific types and functions. So that we can use arch agnostic API while backend can be implemented in arch specific way.

yashi:
I'll close this for now. We might change our mind and just declear it's only for Linux, until then I'll leave it alone.

PR #819, opened 2025-04-01
interfaces: zmq: Destroy thread attributes after pthread_create

moonlight83340:
Destroy the thread attributes after creation, as the attributes are no longer needed once the thread is created.

Kept the assert(ret == 0) pattern for consistency with the existing ZMQ interface code.


yashi:
Please rebase it. Thanks.

moonlight83340:
rebase

PR #824, opened 2025-04-08
Allow CAN broadcast reception when not configuring interface to promisc

troelsjessen:
Previously, posix instances using CAN did not accept L2 and L3 broadcast due to HW filtering


yashi:
Hi @troelsjessen, care to elaborate a bit more?

troelsjessen:
@yashi Sure :-)

Some physical interfaces (including CAN) allow for the HW controller to filter incoming messages without invoking the SW. To do that, filters must be configured to accept messages actually designated for the instance representing the destination address. That is already implemented in the posix CAN driver included in libcsp.

But on top of the listening to traffic designated to the specific node address, every CSP endpoint is also expected to listen in and reply to the broadcast nodes. This counts for the global broadcast node (2¹⁴-1 = 16383) as well as the local broadcast node (e.g. 127 if having an adress in the 64/8 segment).

My commit is adding filters for those two node values in the posix CAN driver, allowing posix CSP endpoints to also reply to broadcast messages and in that way and achieving the same behavior as when having CAN interfaces configured as promisc.

PR #825, opened 2025-04-08
github: template: Add issue template

moonlight83340:
This PR aim to add new issue template :

1. Add RFC & proposal template issue template.
2. Add feature request template issue template.
3. Add enhancement template issue template.

There are some cancelled checks : 

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details

Fix with https://github.com/libcsp/libcsp/pull/826

### DIFF
--- a/.github/ISSUE_TEMPLATE/03_rfc_proposal.md
+++ b/.github/ISSUE_TEMPLATE/03_rfc_proposal.md
@@ -1,0 +1,41 @@
+---
+name: RFC / Proposal
+about: Submit an RFC / Proposal for changes or new features in `libcsp`.  
+title: '' 
+labels: RFC  
+---
+
+## Introduction
+
+### Problem Description
+<!--
+What issue in `libcsp` are we addressing? Is it a bug, missing feature, or performance issue?
+-->
+
+### Proposed Change
+<!--
+A brief overview of the proposed change and its impact on `libcsp`.
+-->
+
+## Details
+
+### Proposed Change (Detailed)
+<!--
+Describe the change in more detail. What will it modify or add in `libcsp`? 
+How will it be implemented?
+-->
+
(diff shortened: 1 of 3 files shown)


yashi:
Do we really need these? what's the diff between them all three are about new things.  Plus, why RFC is also used as new features.

moonlight83340:
Thanks for the feedback!

You're right, all three templates are about introducing "new things", but the intention behind splitting them was to make it easier for contributors and users to categorize their ideas more accurately:

- **RFC / Proposal**: This is meant for bigger or potentially breaking ideas that need discussion before being implemented. It’s more of a design or concept phase — useful for core developers or maintainers.

- **Feature Request**: This is for end-users or community members who want to request new functionality, but don’t necessarily plan to implement it themselves.

- **Enhancement**: For smaller, well-scoped changes to existing functionality — usually things the contributor already knows how to implement.

At the time I created this, my thinking was to simplify issue creation for users by giving them clear entry points depending on their goal.
I agree there's overlap, especially between feature requests and enhancements, so I’m open to merging or renaming some of them if that makes things clearer.
Let me know what you’d prefer!

yashi:
can you list how many isses are there (open or closed) for each? if the number is big, let's make a template. If the number of the type / label are small or used only by a certain ppl, we don't care.

BTW, the reason I've created a few templates was that I had to ask same thing over and over again. That took me time. If you experience the same, please share.

moonlight83340:
My initial motivation was the feeling that these templates could help new users by guiding them better when creating issues.
However, after taking a closer look, I realized that these labels are mainly used by maintainers, and therefore the impact or usefulness of adding templates for them would be quite limited for now.
I'll close the PR for the time being. Thanks again for the feedback!

PR #835, opened 2025-04-14
drivers: socketcan: add CAN FD support to driver

### LINKED ISSUE #794: libcsp 2.0 CAN FD support
Hello everyone, 

I was wondering if libcsp 2.0. supports CAN FD and in the case it doesn't what modifications would be required to use this protocol under Linux and FreeRTOS?

My current understanding is that CAN 2.0 is supported only and in order to use the FD extension both the `interfaces/csp_if_can.c` and the `drivers/can_socketcan.c` modules would require some extra functionality to be added?

moonlight83340:
This commit adds support for CAN FD (Flexible Data-rate) frames in the SocketCAN driver implementation.

The driver now properly handles both standard CAN 2.0 and CAN FD frames.

This enhancement allows CSP nodes using SocketCAN to benefit from higher throughput and payload sizes offered by CAN FD, while maintaining backward compatibility with legacy CAN networks.

Tested on a virtual interface. I would like to test on a physical interface as soon as possible.

To enable virtual can FD
```
ip link add dev vcan0 type vcan
ip link set vcan0 mtu 72
ip link set up vcan0
```
I can provide a sample or example, in another commit.
I can add documentation, in another commit.

Could fix #794

### DIFF
--- a/include/csp/drivers/can_socketcan.h
+++ b/include/csp/drivers/can_socketcan.h
@@ -30,6 +30,22 @@ extern "C" {
 int csp_can_socketcan_open_and_add_interface(const char * device, const char * ifname, unsigned int node_id, int bitrate, bool promisc, csp_iface_t ** return_iface);
 
 /**
+ * Open CAN FD socket and add CSP interface.
+ *
+ * Parameters:
+ * @param[in] device CAN device name (Linux device).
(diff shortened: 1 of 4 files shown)


moonlight83340:
Closing for now, it more like a test on can FD.

PR #841, opened 2025-04-19
drivers: can_socketcan: Add delay when write return EAGAIN/EINTR

moonlight83340:
Previously, when write() returned EAGAIN or EINTR, the transmit loop would
increment the waiting time but continue retrying without any delay.

This could result in busy looping, unnecessarily consuming CPU cycles.

This commit adds a small sleep (5 ms) when such conditions are encountered,
ensuring a more cooperative retry mechanism and consistent behavior with how
ENOBUFS is already handled.

Then in a second commit : 
ENOBUFS, EAGAIN, and EINTR are all considered temporary errors when writing CAN
frames.

Previously, they were handled in separate branches with identical logic.

This commit merges them into a single branch for improved clarity and
maintainability, without changing behavior.

### DIFF
--- a/src/drivers/can/can_socketcan.c
+++ b/src/drivers/can/can_socketcan.c
@@ -124,12 +124,8 @@ static int csp_can_tx_frame(void * driver_data, uint32_t id, const uint8_t * dat
 		
 		written = write(ctx->socket, (void *)pdata, length);
 		if (written < 0) {
-			if (errno == ENOBUFS) {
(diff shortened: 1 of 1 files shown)


yashi (commented):
I feel like having `EAGIN` and `EINTR` in the same condition is the root cause. Have you checked both cases why each case happen and how soon the case is resolved?

moonlight83340:
Let me check and compare both cases. If it turns out to be necessary, I’ll reopen the issue.

PR #845, opened 2025-05-10
arch: add RTEMS port

6arms1leg:
This PR ports libCSP to RTEMS.


moonlight83340:
Hi, 
Thanks a lot for your contribution!

I just wanted to double-check that I understood your intention correctly, from what I see in the RTEMS-specific files, it looks like you're only using POSIX calls. In that case, I'm wondering why the existing posix/ implementation couldn't be reused directly, instead of duplicating it under rtems/.

Is there something I'm missing that requires a separate RTEMS-specific implementation?

I'm not a maintainer, so feel free to take this as just a personal observation, but I was curious about the reasoning behind this approach. Thanks again!

yashi:
Hi @6arms1leg,

I'd love to merge this PR! Thanks!
BTW, 

- Would you please provide us how to build and run it with RTEMS? So that I can test it locally or create GitHub Action to test.
- Would you want to use POSIX pthread instead of the native thread API? I thought RTMS has one, no?

6arms1leg:
Thanks for reviewing this!
You are both right.  This was quite an old change set that I pushed here a year later without properly reviewing it again.
Sorry about that!  I will have another look.

yashi:
You can reopen this PR whenever you’re ready. We’d all welcome an RTEMS port. (I actually built RTEMS with the RTEMS Source Builder for the first time the other week.) For issue and PR management purposes, I’m closing this for now.

PR #869, opened 2025-07-21
arch: posix: refactor pthread_queue_create error handling

yashi:
Error paths in `pthread_queue_create()` were deeply nested conditions with duplicated cleanup code, making it hard to follow and easy to introduce leaks.

Refactor the code to use a linear initialization sequence with goto-based cleanups. Now each failure jumps to a single well-defined cleanup point, ensuring proper resource release, reducing duplication, and improving readability.

### DIFF
--- a/src/arch/posix/pthread_queue.c
+++ b/src/arch/posix/pthread_queue.c
@@ -52,44 +52,57 @@ static inline int init_cond_clock_monotonic(pthread_cond_t * cond) {
 
 pthread_queue_t * pthread_queue_create(int length, size_t item_size) {
 
-	pthread_queue_t * q = malloc(sizeof(pthread_queue_t));
+	int ret;
+	pthread_queue_t * q;
 
-	if (q != NULL) {
-		q->buffer = malloc(length * item_size);
-		if (q->buffer != NULL) {
-			q->size = length;
-			q->item_size = item_size;
-			q->items = 0;
-			q->in = 0;
-			q->out = 0;
-
-			int ret_val = pthread_mutex_init(&(q->mutex), NULL);
-			if (ret_val == 0) { /* Proceed? */
-				ret_val = init_cond_clock_monotonic(&(q->cond_full));
-				if (ret_val != 0) {
-					(void)pthread_mutex_destroy(&(q->mutex)); /* Cleanup */
-				}
-			}
-			if (ret_val == 0) { /* Proceed? */
-				ret_val = init_cond_clock_monotonic(&(q->cond_empty));
-				if (ret_val != 0) {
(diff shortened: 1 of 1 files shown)


yashi:
@6arms1leg @moonlight83340 WDYT?

moonlight83340 (approved):
LGTM

6arms1leg:
Safety-critical projects (common in the aerospace sector) often have requirements that dictate MISRA C-compliance.  IMO, MISRA is a great rule set to make the C language safe (to the degree possible).  MISRA advises not to use `goto` statements (rule 15.1).
Anyway, if this project is not concerned with such strict coding guidelines and does not plan to cater to safety-critical applications, it LGTM too.

yashi:
We care deeply about functional safety and recognize many of the benefits in the MISRA guidelines. However, we have no intention of mindlessly following every rule without first evaluating its merit.  I know libcsp has not yet defined which rules to adopt. Zephyr RTOS conducted a thorough MISRA discussion—their resulting coding guidelines are available at https://docs.zephyrproject.org/latest/contribute/coding_guidelines/index.html and I’m considering using that as our reference.

yashi:
Now, let's talk about `goto` statement specifically. As you can see from the diff, using `goto` makes the code much cleaner and streamlined.  Many projects love this idea.

MISRA-C 2023 Rule 15.1 says "The goto statement _should_ not be used" and goes:

(Emphasis mine)

> Unconstrained use of goto can lead to programs that are unstructured and extremely difficult to
understand.
> [...] if this rule is not followed, the restricted use of goto is allowed where that use follows the
guidance in Rule 15.2 and Rule 15.3.

And these two rules say: 

> Rule 15.2 The goto statement _shall_ jump to a label declared later in the same function

> Rule 15.3 Any label referenced by a goto statement shall be declared in the same block, or in any block enclosing the goto statement

Which is what I have in this PR.

Moreover "SEI CERT C Coding Standard" says: 

> Consider using a goto chain when leaving a function on error when using and releasing resources

https://wiki.sei.cmu.edu/confluence/display/c/MEM12-C.+Consider+using+a+goto+chain+when+leaving+a+function+on+error+when+using+and+releasing+resources

6arms1leg:
As I said, LGTM too, and I do see the appeal. :+1: 
I was just mentioning MISRA C because space projects are often bound to it by requirements.  In such cases it's often easier to be MISRA-compliant than to go through the formal process of justifying the deviation--especially since this is a library and every downstream project will have to deal with the non-compliance again.
Since such coding guidelines are not yet defined for this project, to me the MR looks fine.

PR #870, opened 2025-07-22
github: workflow: restrict workflow permissions

moonlight83340:
Set permissions workflow to limit access to only what is strictly necessary.

This improves security by disabling all other permissions by default.

### DIFF
--- a/.github/workflows/abi-checker.yml
+++ b/.github/workflows/abi-checker.yml
@@ -1,5 +1,7 @@
 name: ABI Compliance Check
 on: [pull_request]
+permissions:
+  contents: read
 jobs:
   abi-check:
     strategy:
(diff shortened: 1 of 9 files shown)


yashi:
Why do we need this? What are you trying to protect?

moonlight83340:
The goal here isn't to fix a specific vulnerability but to apply the principle of least privilege, in the workflow.
This change reduces those by only granting the minimum required permissions at the job level.
https://github.blog/changelog/2023-02-02-github-actions-updating-the-default-github_token-permissions-to-read-only/
https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

yashi (approved):

PR #783, opened 2025-01-18
interfaces: eth: eth_linux: Add waf build support + eth interface example build

moonlight83340:
- waf: interfaces: Add build recipe for eth_linux driver
- example: Add Ethernet interface support to CSP example
- samples: Add sample code for sending via ETH
- github: workflow: Add ETH interface testing
- github: workflow: Exclude meson for ETH interface testing
- pycsp: Add csp_eth_init to the python binding
- examples: python: server: Add command line options for ETH interface
- examples: python: client: Add command line options for ETH interface
- github: workflow: python: Add ETH interface testing

Exclude setcap usage with Meson builds due to shared library error

When using `setcap` on binaries compiled with Meson, running the binary results
in the following error:
    error while loading shared libraries: libcsp.so: cannot open shared object
    file: No such file or directory

This issue occurs because `setcap` disables the use of `LD_LIBRARY_PATH`, which
Meson relies on to locate shared libraries during runtime.

As a temporary workaround, I have excluded the use of `setcap` with Meson-built
binaries.

Future work:
- Investigate alternative methods to specify library paths when
using `setcap`.
- Explore potential fixes to allow seamless integration of
`setcap` with Meson builds.

I think the error is related with that issue :
See : https://stackoverflow.com/questions/61796975/setting-ld-library-path-after-using-setcap

I'm currently investigating the problem, but since it only occurs with the Meson build, I have excluded Meson for now to find a workaround.

github workflow log :
```
RUN ETH server test

Run ./build/examples/csp_server -e veth1 -a 1 -T 10 -v 1 &
  ./build/examples/csp_server -e veth1 -a 1 -T 10 -v 1 &
  ./build/examples/csp_client -e veth0 -a 2 -C 1 -t -v 1
  shell: /usr/bin/bash -e {0}
./build/examples/csp_server: error while loading shared libraries: libcsp.so: cannot open shared object file: No such file or directory
./build/examples/csp_client: error while loading shared libraries: libcsp.so: cannot open shared object file: No such file or directory
Error: Process completed with exit code 127.
```

Things I have consider : 
https://unix.stackexchange.com/questions/83191/how-to-make-sudo-preserve-path
https://man7.org/linux/man-pages/man8/sudo.8.html

PR #873, opened 2025-07-24
doc: Enhance example usage documentation

moonlight83340:
- Add “Available Example Programs” section
- Clarify build script behavior and output directory

The csp_server_client example has been removed from the documentation.
On platforms without memory protection, CSP uses the loopback
interface (lo) with shared memory for communication.
Because of this mechanism, csp_client and csp_server cannot
communicate over lo on Linux due to a limitation on Linux ports.

Depend on https://github.com/libcsp/libcsp/pull/783

### DIFF
--- a/doc/example.md
+++ b/doc/example.md
@@ -1,9 +1,33 @@
 # Client and server example
 
-The example in `examples/csp_server_client.c` provides a
-simple server/client setup, where the client sends a request to the
-server and receives a reply. The code can be compiled to an executable
-using `./examples/buildall.py`.
+This page explains how to build and run the example programs provided with `libcsp`.
+
+## Building the examples
+
+To compile all the examples, simply run:
+
+`./examples/buildall.py`
+
+This script builds all the example binaries using your system’s default compiler and 
+places the resulting executables in the `build/examples/` directory.
+
+## Available Example Programs
+
+The following example programs are included:
+
+- **`csp_server`**  
+  A server that listens for incoming packets.
+
+- **`csp_client`**  
+  A client that sends packets to a specified destination.
+
+- **`zmqproxy`**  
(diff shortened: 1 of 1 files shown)


doc/example.md
```diff
@@ -1,9 +1,36 @@
...
+
+This script builds all the example binaries using your system’s default compiler and 
+places the resulting executables in the `build/examples/` directory.
+
+## Available Example Programs
+
+The following example programs are included:
+
+- **`csp_server_client`**  
+  A single binary that starts both a client and a server using the loopback interface.
```
yashi: I feel like it's not right. it is an implementation of server and client using threads.  It can handle multiple interfaces, no?

moonlight83340: Not yet. While the architecture could support multiple interfaces, currently the implementation does not configure any. So it falls back to the loopback interface by default. With some modifications, it could indeed support more.

yashi: Indeed. I stand corrected.

Can we just drop csp_server_client?

moonlight83340: I guess we could, I'm not sure it add anything more than csp_client & csp_server.

yashi: It's just a historical thing. Drop it from the doc and we'll merge this.

yashi:
Would you mind adding an output from `csp_server` and `csp_client` at where `csp_server_client` was?

yashi:
> Would you mind adding an output from `csp_server` and `csp_client` at where `csp_server_client` was?

Did you add the output?

moonlight83340:
> > Would you mind adding an output from `csp_server` and `csp_client` at where `csp_server_client` was?
> 
> Did you add the output?

I have removed the csp_server_client example from the documentation.
On platforms without memory protection, CSP uses the loopback interface (lo) with shared memory for communication.
Because of this mechanism, csp_client and csp_server cannot communicate over lo on Linux due to a limitation on Linux ports.

PR #875, opened 2025-07-29
interfaces: csp_if_udp: Change inet_aton to inet_pton

moonlight83340:
inet_aton isn't posix api so change it to inet_pton, to be posix compliant.

https://man7.org/linux/man-pages/man3/inet_aton.3.html
https://man7.org/linux/man-pages/man3/inet_pton.3.html
https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_ntop.html

### DIFF
--- a/src/interfaces/csp_if_udp.c
+++ b/src/interfaces/csp_if_udp.c
@@ -112,8 +112,9 @@ void csp_if_udp_init(csp_iface_t * iface, csp_if_udp_conf_t * ifconf) {
 
 	iface->driver_data = ifconf;
 
-	if (inet_aton(ifconf->host, &ifconf->peer_addr.sin_addr) == 0) {
-		csp_print("  Unknown peer address %s\n", ifconf->host);
+	if (inet_pton(AF_INET, ifconf->host, &ifconf->peer_addr.sin_addr) == 0) {
+		csp_print("Invalid peer address: %s", ifconf->host);
+		return;
 	}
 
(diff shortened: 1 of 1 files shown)


yashi:
It does indeed switch to POSIX but that's it.  The error handing is incomplete for both `inet_aton()` and `inet_pton()`.

moonlight83340:
Yes, if an invalid peer address is passed, we should return immediately at this point.

yashi:
AFAIK, we never declared POSIX compliance. I believe `inet_aton()` is from 4.3BSD, and POSIX simply didn’t include it.  
That said, it’s better to check whether FreeRTOS and Zephyr support `inet_pton()` before we change APIs.

- Zephyr `inet_pton()` (POSIX layer): https://docs.zephyrproject.org/apidoc/latest/inet_8h.html
- FreeRTOS: I'm not sure about it.

Just saying “because it’s a POSIX function” doesn’t buy us anything. I’m not saying I don’t want to move toward POSIX, but we must make sure current users won’t be surprised.

moonlight83340:
I just wanted to clarify my initial intentions with this PR.
My idea was that with POSIX support, we could enhance the libcsp interface by improving portability and integration with standard system calls.

Currently, we do not compile the CSP UDP interface for Zephyr or FreeRTOS.
https://github.com/libcsp/libcsp/blob/develop/src/interfaces/CMakeLists.txt

Zephyr also don't have `inet_aton()` only `inet_pton()` with the posix layer.

PR #878, opened 2025-08-03
interfaces: csp_if_udp: Refactor and harden UDP interface initialization

### LINKED ISSUE #863: infinite loop
We have a few infinite loop like `while (true)` in our code base. But in space industry, it not a favourite thing to have:

- [NASA/JPL The Power of 10](https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code): _All loops must have fixed bounds. This prevents runaway code._
- MISRA C: _Rule 14.3Controlling expressions shall not be invariant_
- [SEI CERT C Coding Standard](https://wiki.sei.cmu.edu/confluence/display/c/SEI+CERT+C+Coding+Standard): [MSC21-C. Use robust loop termination conditions](https://wiki.sei.cmu.edu/confluence/display/c/MSC21-C.+Use+robust+loop+termination+conditions)

I'm not a big fan of MISRA C; rather most of the time I hate it, but there is some idea we can learn from it.

Having easy-to-use function for applications or critical allocation path in the library is ok but it's not an everyday function in the library. WDYT?

moonlight83340:
### Fix a long time bug with sockfd
0 is a valid file descriptor, so comparing with 0 may skip valid descriptor.

socket() returns -1 on failure; 0 is a valid file descriptor, so comparing
with 0 may skip valid descriptor.

This PR brings several important improvements and refactorings to the UDP interface implementation in libcsp:

    Deterministic socket creation and binding during interface init
    Moved socket creation and bind logic from the RX thread loop to the synchronous csp_if_udp_init() function.
    This avoids silent retries and spinning in the RX thread, making initialization fail-fast and deterministic, 
    which is critical for embedded and space applications.

    Improved resource cleanup on initialization failure
    Introduces a consistent goto-based cleanup path in csp_if_udp_init() to close sockets 
    and destroy pthread attributes on failures, preventing resource leaks and leaving the interface config
    in a consistent state.

    Check socket validity before transmission
    Adds a socket validity check before attempting to send packets. 
    If the socket is invalid (sockfd < 0), the packet is safely dropped with an error log, 
    preventing undefined behavior or leaks.

Together, these changes increase robustness, correctness, and maintainability of the UDP interface code, 
aligning with CSP’s philosophy of reliable, predictable initialization and operation.

Partially fix https://github.com/libcsp/libcsp/issues/863
Inspired by discussion from this closed PR https://github.com/libcsp/libcsp/pull/812

### DIFF
--- a/src/interfaces/csp_if_udp.c
+++ b/src/interfaces/csp_if_udp.c
@@ -73,25 +73,9 @@ static void * csp_if_udp_rx_loop(void * param) {
 	csp_iface_t * iface = param;
 	csp_if_udp_conf_t * ifconf = iface->driver_data;
 
-	ifconf->sockfd = -1;
-
-	while (ifconf->sockfd < 0) {
-
-		ifconf->sockfd = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
-
-		struct sockaddr_in server_addr = {0};
-		server_addr.sin_family = AF_INET;
-		server_addr.sin_addr.s_addr = htonl(INADDR_ANY);
-		server_addr.sin_port = htons(ifconf->lport);
-
-		bind(ifconf->sockfd, (struct sockaddr *)&server_addr, sizeof(server_addr));
-
-		if (ifconf->sockfd < 0) {
-			csp_print("  UDP server waiting for port %d\n", ifconf->lport);
-			sleep(1);
-			continue;
-		}
-		break;
+	if (ifconf->sockfd < 0) {
+		csp_print("csp_if_udp_rx_loop: invalid socket\n");
+		return NULL;
 	}
 
 	while (1) {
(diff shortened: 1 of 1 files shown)


yashi:
would you please make sure the markup in descript is right?

moonlight83340:
I add some modification on the description, there was in fact an error I wrote :

> If the socket is invalid (sockfd <= 0)

But it should be 

if the socket is invalid (sockfd < 0)

src/interfaces/csp_if_udp.c
```diff
@@ -23,15 +21,13 @@ static int csp_if_udp_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packe
...
+	if (ifconf->sockfd < 0) {
+		csp_print("Sockfd invalid\n");
 		csp_buffer_free(packet);
 		return CSP_ERR_NONE;
 	}
 
 	csp_id_prepend(packet);
-	ifconf->peer_addr.sin_family = AF_INET;
-	ifconf->peer_addr.sin_port = htons(ifconf->rport);
 	sendto(ifconf->sockfd, packet->frame_begin, packet->frame_length, MSG_CONFIRM, (struct sockaddr *)&ifconf->peer_addr, sizeof(ifconf->peer_addr));
```
yashi, line 33: You can send to a different address using `sendto(2)`.

yashi (changes requested):

src/interfaces/csp_if_udp.c
moonlight83340: Could you clarify what you mean? I’m aware that `sendto(2)` can be used to send to different addresses, but our current implementation always uses a fixed peer address.

yashi: > our current implementation always uses a fixed peer address.

Does it? Then why do we have the `dest` arg as the second parameter to `csp_sendto()`?

moonlight83340: My misunderstanding, sorry.
Let me check this PR one more time.

PR #881, opened 2025-08-05
contrib: zephyr: Simplify test-mode exit handling

moonlight83340:
The previous version used a goto to jump to a common 'end:' label,
but both branches already followed the same path.

This change removes the redundant goto and label, restoring a
straightforward break and exit(ret) sequence.

The ret variable is now initialized to 1 by default for defined
behavior.

### DIFF
--- a/contrib/zephyr/samples/server-client/main.c
+++ b/contrib/zephyr/samples/server-client/main.c
@@ -137,7 +137,7 @@ void client(void) {
 /* main - initialization of CSP and start of server/client tasks */
 int main(void) {
 
-	int ret;
+	int ret = 1;
 	uint8_t uart_address = 1;
 	uint8_t can_address = 10;
 	const char * kiss_device = NULL;
@@ -253,18 +253,14 @@ int main(void) {
 
 		if (test_mode) {
 			/* Test mode is intended for checking that host & client can exchange packets over loopback */
-			if (server_received < 5) {
-				LOG_INF("Server received %u packets", server_received);
-				ret = 1;
-				goto end;
+			if (server_received >= 5) {
+				ret = 0;
 			}
 			LOG_INF("Server received %u packets", server_received);
-			ret = 0;
-			goto end;
+			break;
 		}
 	}
 
-end:
 	if (IS_ENABLED(CONFIG_CSP_HAVE_CAN)) {
 		csp_can_stop(can_iface);
 	}


yashi:
Let's define `ret` to `1` by default.  And set to `0` when we are sure everything is OK.

yashi:
With the current implementation, it will always be a failure if test mode isn't activated, no?

moonlight83340:
Without test_mode, the program doesn't fail, but it will loop indefinitely until interrupted (e.g., via SIGINT).

yashi:
Have you tested?

contrib/zephyr/samples/server-client/main.c
```diff
@@ -253,13 +253,10 @@ int main(void) {
...
 		if (test_mode) {
 			/* Test mode is intended for checking that host & client can exchange packets over loopback */
-			if (server_received < 5) {
-				LOG_INF("Server received %u packets", server_received);
-				ret = 1;
-				goto end;
+			if (server_received >= 5) {
+				ret = 0;
 			}
 			LOG_INF("Server received %u packets", server_received);
```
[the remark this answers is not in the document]
moonlight83340:
```
		if (test_mode) {
			/* Test mode is intended for checking that host & client can exchange packets over loopback */
			if (server_received < 5) {
				LOG_INF("Server received %u packets", server_received);
				ret = 1;
				goto end;
			}
			LOG_INF("Server received %u packets", server_received);
			ret = 0;
			goto end;
		}
```
It seems Copilot didn't understood the change.

```
		if (test_mode) {
			/* Test mode is intended for checking that host & client can exchange packets over loopback */
			if (server_received >= 5) {
				ret = 0;
			}
			LOG_INF("Server received %u packets", server_received);
			goto end;
		}
```
Simplify test_mode result handling
The same LOG_INF message was duplicated in both branches.
Now we log once after updating ret (set to 0 on success, otherwise stays 1),
then goto end since both cases end execution.

contrib/zephyr/samples/server-client/main.c
```diff
@@ -253,13 +253,10 @@ int main(void) {
 
 		if (test_mode) {
 			/* Test mode is intended for checking that host & client can exchange packets over loopback */
-			if (server_received < 5) {
-				LOG_INF("Server received %u packets", server_received);
-				ret = 1;
-				goto end;
+			if (server_received >= 5) {
+				ret = 0;
```
yashi, line 257: This is the only place to set `ret` to `0`. This means that if you are not in the test mode, you always fail. Because the program is infinite loop at the end, `ret` actually not used. But still I feel it's not a good practice.

contrib/zephyr/samples/server-client/main.c
```diff
@@ -253,13 +253,10 @@ int main(void) {
...
 			}
 			LOG_INF("Server received %u packets", server_received);
-			ret = 0;
 			goto end;
```
yashi: This is the only `goto end` to break out from the while loop, why not use `break` instead.

yashi (changes requested):

PR #903, opened 2025-10-24
bindings: python: Fix error handling in PyArg_ParseTuple

moonlight83340:
Replace incorrect uses of Py_RETURN_NONE with `return NULL` after PyArg_ParseTuple() failures.

Returning NULL correctly propagates the Python exception raised by the argument parser, while Py_RETURN_NONE would silently ignore the error and return None, masking invalid argument issues.

https://docs.python.org/3/extending/extending.html
> [PyArg_ParseTuple()](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple) returns true (nonzero) if all arguments have the right type and its components have been stored in the variables whose addresses are passed. It returns false (zero) if an invalid argument list was passed. In the latter case it also raises an appropriate exception so the calling function can return NULL immediately (as we saw in the example).


yashi:
It seems we should return `NULL`.
https://docs.python.org/3/extending/extending.html

But I'm not sure. I'm not a python guy. Hope someone charm in.

PR #906, opened 2025-10-26
github: workflows: build-test: Remove trailing white spaces

yashi:
Remove the trailing white spaces.


yashi:
I'm not sure this is related to this but..
image

EDIT: It seems it IS related.

Issue #907, opened 2025-10-27
CI: SFP Test fails with “SFP protocol error or inconsistency” and CRC mismatch


moonlight83340:
## Description

Description:
During the GitHub Actions CI run, the SFP example test (`csp_sfp_server_client`) fails with multiple errors.

CI Run:
[Link to the failed run](https://github.com/libcsp/libcsp/actions/runs/18832827306/job/53727361133?pr=886)

## Expected behavior

The SFP test should complete successfully without protocol errors or CRC mismatches.

## Actual Behavior:

```
Run ./build/examples/csp_sfp_server_client --mtu 1 --size 1
  ./build/examples/csp_sfp_server_client --mtu 1 --size 1
  ./build/examples/csp_sfp_server_client --mtu 1 --size 128
  ./build/examples/csp_sfp_server_client --mtu 128 --size 1
  ./build/examples/csp_sfp_server_client --mtu 128 --size 10
  ./build/examples/csp_sfp_server_client --mtu 128 --size 100
  ./build/examples/csp_sfp_server_client --mtu 128 --size 1000
  ./build/examples/csp_sfp_server_client --mtu 128 --size 10000
  ./build/examples/csp_sfp_server_client --mtu 128 --size 100000
  shell: /usr/bin/bash -e {0}
Test completed!
Test completed!
Test completed!
Test completed!
Test completed!
Test completed!
Test completed!
line 247: Failed csp_sfp_recv dew to: SFP protocol error or inconsistency
line 211: Failed csp_sfp_send dew to: Not enough memory
line 322: CRC mismatch
Error: Process completed with exit code 1.
```

## Environment Details:

Provide information about your environment, including:

- Build OS:  Ubuntu 22.04
- CI: GitHub Actions
- meson
-  gcc-12,
-  g++-12
- csp version 2
- libcsp commit hash: https://github.com/libcsp/libcsp/pull/886

## Logs or console output from the very beginning to the end
Every logs :
https://gist.github.com/moonlight83340/c6ef0a43860342aea8dc923a7601eb67

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

moonlight83340:
[...]

PR #910, opened 2025-10-28
sfp: example: Fix buffer starvation in UDP mode

dimitrovand:
Previously, a hardcoded timeout was used to avoid buffer starvation when running in UDP mode. This update replaces the fixed delay with a check for available buffers, waiting until at least (CSP_BUFFER_COUNT / 2) + 1 are free before continuing. The +1 provides an extra safety margin.

Fix for https://github.com/libcsp/libcsp/issues/907

### DIFF
--- a/examples/csp_sfp_server_client.c
+++ b/examples/csp_sfp_server_client.c
@@ -259,11 +259,17 @@ static int loopback_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packet,
     (void)iface;
     (void)via;
     (void)from_me;
-
-    /* add some sleep to avoid starving the system when RDP is not used */
-    if (!test_options.rdp)
-        usleep(1000);
+    
     csp_qfifo_write(packet, &csp_if_lo, NULL);
+    
+    /* add some sleep to avoid starving the system when RDP is not used */
+    if (!test_options.rdp) {
+        const int min_free_buffers = (CSP_BUFFER_COUNT / 2) + 1;
+        while (min_free_buffers > csp_buffer_remaining()) {
+            usleep(1000);
+        }
+    }
+
     return CSP_ERR_NONE;
 }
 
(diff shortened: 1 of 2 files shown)


yashi (commented):
@dimitrovand Thanks for the fix. I’m worried the change is ad-hoc; the buffer threshold feels arbitrary and not tied to packet arrival/pacing.

By sleeping, what do you want to gain?

AFAICS, even if you wait longer, the system might receive a new packet and use more memory. IOW, you never know when the system has enough free memory to send.

Do I understand the root cause correctly?

dimitrovand:
Hi @yashi,
The fix specifically targets this test case (SFP + UDP + loopback). The buffer threshold is arbitrary, but I wanted to at least base it on a CSP-provided constant.
By adding a sleep in the loopback TX path, we intentionally slow down the sender (only when necessary) until the receiver has processed enough of the sender’s data. Since this test runs in isolation (no external interference), we know packets are processed only by the receiver thread, so it’s safe for the sender to wait.
You got the problem right and in real-world SFP over UDP should be avoided - it’s just unpredictable.

yashi:
hmm... I red a few times before replying this but it still does't convince me.

1. The starvation can happen to any configuration. It doesn't depend on UDP nor loopback (nor any protocol / interface combination).  Thus, ` if (!test_options.rdp)` is bogus.
1. You said "The fix specifically targets this test case", and yes in some sense it's only for test (both are an example and a sample). But the purpose of sample programs are to give working sample code for those who wants to learn how to use libcsp.  If `simple-sfp-send-recv` has the block or code, it will be copied to all other sfp related code. It shouldn't have situation specific code.
1. The reason of the sleep, you said, is to "intentionally slow down the sender [...] until the receiver has processed enough of the sender’s data".  As you said, this _is_ the test specific.  We never know how telemetries and commands are designed for real spacecraft and we can't assume that tx and rx is balanced or not. it might be used for rx intense case. If you mimic ping-pong style of telecommand, why don't we just make the length of the sleep a test parameter? I don't see any valid reason to use `CSP_BUFFER_COUNT`.

johandc:
I think we should be building simple flow control / rate limit into sfp. We have a similar concept at spaceinventor we call it dtp, which we will be looking into making available.

dimitrovand:
@yashi @johandc sure, you can decline the PR

PR #912, opened 2025-11-01
zephyr: Replace symbolic link to Zephyr module

### LINKED ISSUE #911: Zephyr build on Windows does not work without Git symlink support enabled
## Description

The libcsp repository currently symlinks the `zephyr` folder containing `module.yml` to `contrib/zephyr`.
Git installations on Windows don't have symlinks enabled by default, causing the Zephyr build to break because it can't find `zephyr/module.yml` in the repository root.

## Steps to Reproduce

On Windows:
1. Create a new Zephyr project using a west workspace setup like the [Zephyr Example Application](https://github.com/zephyrproject-rtos/example-application)
2. Add libcsp as a west module:

```
manifest:

  remotes:
    ...
    - name: libcsp
      url-base: https://github.com/libcsp

  ...
  projects:
    - name: libcsp
      remote: libcsp 
      revision: develop
      path: modules/lib/libcsp

```
3. Try to enable libcsp by setting `CONFIG_LIBCSP=y`
4. build.

## Expected behavior

The build system should find libcsp and find the Kconfig and CMakeLists. The build might still fail at a later stage but that's not the point.

## Actual Behavior:

Instead of putting a symlink there, git just puts a text file with the path of where the symlink would point if there was one. Zepyhr will not detect it as a west module and will not find CMakeLists and Kconfig.

## Environment Details:

- Build OS: Windows
- Target OS: Zephyr
- libcsp commit hash: develop (f02d28e8)
- Meson version: -
- CMake version: 3.25.3
- GNU Make version: -
- Ninja version: -

## Logs or console output from the very beginning to the end

-- Generated devicetree_generated.h: C:/Users/.../build/zephyr/include/generated/zephyr/devicetree_generated.h

C:/Users/.../app/prj.conf:13: warning: attempt to assign the value 'y' to the undefined symbol LIBCSPParsing C:/Users/.../app/Kconfig

## Additional Context:

Symlinks are disabled by default.

Image

## Ways to fix:

Option 1:
Remove the symlink and create a new `zepyhr` folder with only `module.yml` inside at the root of the repository.

Option 2:
Move the whole `zephyr` folder from `contrib/zephyr` to `zephyr`, including the samples.

I like option 1 because it only moves the essential files for the build system to the root of the repository.
Everything else can stay in contrib.

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

stadlero:
Fixes Windows builds on machines with symbolic links disabled.
 
Removes the symbolic link from `zephyr`-> `contrib/zephyr`
and moves the `module.yml` file into a new `zephyr` directory.

Fixes #911


yashi:
Thank you @stadlero for raising this issue. Right now we don't have a native build support for Windows. We had Windows "port" a while back but not now neither. Just curious, how do you build libcsp and zephyr? Does the current CMake just works? If you build your project using libcsp + Zephyr on Windows, I'd definitely love to support it.

The idea of using symlink was @johandc 's  (https://github.com/libcsp/libcsp/pull/222#issuecomment-927324082). His reasoning was that he doesn't like a congested project root directory. So I'd like to wait for him what he thinks.

BTW, do you think you can add a Github action to test build on a Windows runner?

stadlero:
Hi @yashi 
Thanks for thinking about adding support!

I’m sorry, I wasn’t aware that Windows isn’t supported as a build host.
It just worked after I reinstalled Git with the correct settings, so I didn’t question it :)

We are using a Zephyr repository setup that has the same structure as the [Zephyr Example Application](https://github.com/zephyrproject-rtos/example-application)

We add libcsp as a Zephyr module to our `west.yml` like this:

```
manifest:

  remotes:
    ...
    - name: libcsp
      url-base: https://github.com/libcsp

  ...
  projects:
    - name: libcsp
      remote: libcsp 
      revision: develop
      path: modules/lib/libcsp
```

And it builds just fine for our STM32 boards.

I saw the commit from @johandc
And I agree that cluttering the root with stuff should be avoided.

I would argue that having a symlink there is "just as ugly" as having a real folder.
But the real folder would help compatibility.
And I just put the necessary file there. Everything else stays well organized in `contrib`

And I am happy to work on the Github Action.
I will come back to you when I have a working version.
That test would be running in [libcsp-zephyr](https://github.com/libcsp/libcsp-zephyr) I assume?

stadlero:
Update:

- Added two PRs to the test repo: libcsp/libcsp-zephyr#5, libcsp/libcsp-zephyr#6
- Confirmed that the build fails on Linux and Windows if symlinks are disabled: https://github.com/stadlero/libcsp-zephyr/actions/runs/19050476910

```text
INFO    - The following issues were found (showing the top 10 items):
INFO    - 1) libraries.libcsp on mps2/an385 error (CMake build failure - undefined symbol LIBCSP)
INFO    - 2) libraries.libcsp on qemu_cortex_m3/ti_lm3s6965 error (CMake build failure - undefined symbol LIBCSP)
```

- Successfully tested this PR with symlinks **enabled**: https://github.com/stadlero/libcsp-zephyr/actions/runs/19051251781
- Successfully tested this PR with symlinks **disabled**: https://github.com/stadlero/libcsp-zephyr/actions/runs/19051285684

yashi:
That's super! Thank you!

Given that the current main targets (Linux, FreeRTOS, and Zephyr), if there is a real user of Windows environment, I don't think @johandc will object. He is just too busy these days, so give him a week or so. If he didn't response, I'll merge your change.

yashi:
I'd like to keep build essential files under `zephyr/`, that is `CMakeLists.txt`, `Kconfig`, and `module.yml`. No need for modification. Just rename them.

We can leave `contrib/zephyr/samples` as is.

stadlero:
Moved the files.
Had to make a small modification to the `CMakeLists.txt` because the relative path to the root changed.
Build still works on the runners and on our project.

yashi (approved):

yashi:
Thank you @stadlero again for the work. Merged.

PR #914, opened 2025-11-07
python: Added conditional compile flags for socketcan functions.

olmanqj:
I added conditional compile flags for `socketcan` related functions in python bindings. This is useful when I won't use `socketcan` with the python bindings, so I don't need to install `libsocketcan` just for using the bindings.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ libcsp 2.2, xx-xx-202x
 ----------------------
 - new: csp_sfp_send(), csp_sfp_conn_max_mtu(), csp_sfp_opts_max_mtu()
 - improvement: Small Fragmentation Protocol has been reworked to remove malloc() (#742)
+- improvement: PyCSP can now be built without SocketCAN (#638)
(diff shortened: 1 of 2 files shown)


moonlight83340:
Hi, thank you for your contribution!

I think you can squash your commits into one.
Could you also review your commit summary and add a proper commit body?

Here are the triggered errors:
```
1: T1 Title exceeds max length (83>72): "python: Added conditional compile flags for socketcan functions in python bindings."
1: T3 Title has trailing punctuation (.): "python: Added conditional compile flags for socketcan functions in python bindings."
3: B6 Body message is missing
```


I think a commit message like this would be fine:
```
pycsp: Add conditional compile flag for SocketCAN support

The SocketCAN-related functions are now only compiled when
CSP_HAVE_LIBSOCKETCAN is defined.
This ensures that the Python bindings can be built even on systems
without libsocketcan installed.
```

Also, don’t forget to add your Signed-off-by line 👀

moonlight83340:
Something went wrong you still have 5 commits :sweat_smile: 
@olmanqj

PR #915, opened 2025-11-11
interfaces: csp_if_kiss: Make KISS CRC optional

### LINKED ISSUE #892: Request for KISS CRC move
Hello,

I was trying to implement CSP (KISS) parser in wireshark, and stumbled upon what seems to be a confusing implementation. 
In the wireshark, protocols are dissected in layers. And so from USB BLK transfer I get data payload, which is a KISS packet. 
After parsing 0xC0 and etc., in the KISS payload, I call another layer parser, which is CSP. However CSP gets additional 4 bytes of payload, which happens to be a CRC.
The problem is, that in CSP header, CRC flag is turned off; so CSP layer parser does not expect to have 4 CRC bytes.
Standard KISS (AX.25) however, does not describe CRC usage at all. CSP calculates this CRC only for **its payload**; so it is actually a **CSP packet CRC**. 

To fixup this confusion, it would be much clearer to remove CRC calculation from KISS layer, but add a mandatory CRC Flag in the CSP header when using KISS. This way the CSP engine itself will deal with CRC's, and leave KISS only for transport. 
Any thoughts appreciated.

moonlight83340:
Introduce a new compile-time option `CSP_ENABLE_KISS_CRC` to allow enabling or disabling the CRC in the KISS interface.

Default behavior preserves legacy operation (CRC enabled) Add support for this option in:
  * CMake (`CMakeLists.txt`)
  * Waf (`wscript`)
  * Meson (`meson.build`)

Updated autoconfig headers and build system defines to propagate the option

This makes the CRC in KISS optional while keeping legacy builds fully compatible. Future builds can enable or disable CRC as needed.

Fix https://github.com/libcsp/libcsp/issues/892

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 libcsp 2.2, xx-xx-202x
 ----------------------
 - new: csp_sfp_send(), csp_sfp_conn_max_mtu(), csp_sfp_opts_max_mtu()
+- new: Make KISS interface CRC optional via CSP_ENABLE_KISS_CRC (#892)
 - improvement: Small Fragmentation Protocol has been reworked to remove malloc() (#742)
 - removed: csp_sfp_send_own_memcpy()
 - break; csp_sfp_recv_fp()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,8 @@ option(CSP_BUILD_SAMPLES "Build samples and examples by default" OFF)
 
 option(CSP_FIXUP_V1_ZMQ_LITTLE_ENDIAN "Use little-endian CSP ID for ZMQ with CSPv1" ON)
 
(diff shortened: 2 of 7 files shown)


DjLukas74 (approved):

yashi:
Please add documentation and an entry to CHANGELOG.

yashi:
@tibor0991, do you have any comment?

tibor0991:
> @tibor0991, do you have any comment?

Late reply, sorry, LGTM.

Just one question: is it fine to leave it enabled by default for the next releases?

yashi:
@tibor0991, Don't be sorry. It was me who accidentally merged this before your comment. I applogize.

I'm thinking about disabling it before release if everybody is ok with it. This makes the option for Waf to be renamed to `--enable-kiss-crc`, IIRC.

PR #921, opened 2025-12-02
Fix typo in example code

wookhyunshin:
There is a discrepancy in the csp_server example: the code specifies a 50ms timeout for csp_read, while the comment indicates 100ms

### DIFF
--- a/examples/csp_server.c
+++ b/examples/csp_server.c
@@ -61,7 +61,7 @@ static void * server(void * param) {
 			continue;
 		}
 
-		/* Read packets on connection, timeout is 100 mS */
+		/* Read packets on connection, timeout is 50 mS */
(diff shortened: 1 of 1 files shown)


moonlight83340:
Hi,
Thank you for your PR.
Can you resolve your commits please ? You should'nt have merged commit here.
You should have only the last one [Fix typo in example code](https://github.com/libcsp/libcsp/pull/921/commits/f19ddd7af0cd89035a0b55a7de18bba5b85cffb9)

You could write your commit log like this :
```
examples: csp_server: Fix `csp_read` timeout value in comment

Timeout for `csp_read` is set to 50 ms in the code, but the comment
still indicated 100 ms.
Update the comment to reflect the correct value.
```


You can also add your sign-off if possible.

PR #924, opened 2025-12-16
Allow setting "mask" and "is_default" itf attributes from Python bindings

jeanbaptistelab:
This change extends the "zmqhub_init", "socketcan_init" & "kiss_init" bindings
to accept new (optional) "is_default" and "mask" positional parameters that are
then applied to the corresponding added interface, allowing more flexibility in
interface management from Python

### DIFF
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -840,15 +840,21 @@ static PyObject * pycsp_cmp_clock_get(PyObject * self, PyObject * args) {
 #if CSP_HAVE_LIBZMQ
 static PyObject * pycsp_zmqhub_init(PyObject * self, PyObject * args) {
 	uint16_t addr;
-	char * host;
-	if (!PyArg_ParseTuple(args, "Hs", &addr, &host)) {
+	const char * host = "localhost";
+	int is_default = 0;
+	uint16_t mask = 8;
+	if (!PyArg_ParseTuple(args, "H|spH", &addr, &host, &is_default, &mask)) {
 		return NULL;  // TypeError is thrown
 	}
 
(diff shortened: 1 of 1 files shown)


yashi:
Thanks, @jeanbaptistelab.

The reason some tests are faling is that you've updated Meson version. Do you need a newer version? If so, please describe it in your commit log. Thanks.

jeanbaptistelab:
The reason I updated the meson version requirement was to get rid of warnings that have been around for quite some time, for instance here:
[WARNING: Project specifies a minimum meson_version '>=0.53.2' but uses features which were added in newer versions:
0.63.0: {'python_installation.extension_module with implicit dependency on python'}](https://github.com/libcsp/libcsp/actions/runs/20204299385/job/58000567182)

I'll gladly revert the meson version requirement

yashi:
Thanks. Please squash the two commits to one.

Regarding Meson version, unfortunately Ubuntu 22.04 only has Meson v0.61.2, which is not enough for the currently required 0.63.0 for python bindings.  I should disable Python binding check with Ubuntu 22.04.

jeanbaptistelab:
> Thanks. Please squash the two commits to one.
> 
> Regarding Meson version, unfortunately Ubuntu 22.04 only has Meson v0.61.2, which is not enough for the currently required 0.63.0 for python bindings. I should disable Python binding check with Ubuntu 22.04.

You could instead update the `build-test-python.yml`, there is already some special casing for python >= 3.12, why not simply remove the `Setup build system packages on Linux` step entirely and remove the `&& matrix.python-version >= '3.12'` clause from the `Install the latest Meson using Pip if Python is 3.12 or later` step ?

PR #933, opened 2026-01-13
Service transaction

drowzie:
Due to the potential failure of service_transaction or csp_send i created the additional functions 
service_transaction_persistent_retry, and _reconnect.

Retry provides functionality for retrying to send N amount of packets,
and Reconnect provides reconnection on failure, and retries combined.

Considering packet loss can occur, and a loss of connection then these seemed like nice additions.


drowzie:
Not intended for the PR in this repo...

PR #948, opened 2026-03-04
Use csp_buffer_get_always only when incoming packets is not to be routed

troelsjessen:
The functionality is used to stall a module in case it is running out of buffers, potentially causing a reboot. Nevertheless, running out of buffers is a valid situation if a module is routing between two physical networks with different capatilities (e.g. CAN vs ETH).


troelsjessen:
Apologies, this pull request was opened by mistake

PR #960, opened 2026-04-17
examples: simple-send-usart: Fix send length

ajrgallego:
The packet length was previously hardcoded to "3" for the "abc" example string, excluding the null terminator (\0). This could result in undefined behaviour on the receiver side. The hardcoded values have now been removed for future implementations.


yashi (commented):
This is actually intentional. The example demonstrates sending arbitrary-length data, not a string. That is why it uses `memcpy()` instead of `strcpy()`.

Because `packet->length` is set to `3`, the receiver should treat the received data as exactly three bytes. The sample happens to use printable ASCII, but the payload can be anything, including `0x0`.

PR #961, opened 2026-04-20
samples: posix: README: Clarify payload data handling

yashi:
The sample payload is raw data, not a null-terminated string. Update the README files to describe the transmitted data as a 3-byte payload and note that the receiver should interpret it using `packet->length`.

@ajrgallego, WDYT?

Closes #960

### DIFF
--- a/samples/posix/simple-send-canbus/README.md
+++ b/samples/posix/simple-send-canbus/README.md
@@ -1,7 +1,11 @@
 # Simple Send via CANbus
 
-This is a simple sample code to send `"abc"` to a server via the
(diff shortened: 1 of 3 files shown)


ajrgallego:
> The sample payload is raw data, not a null-terminated string. Update the README files to describe the transmitted data as a 3-byte payload and note that the receiver should interpret it using `packet->length`.
> 
> @ajrgallego, WDYT?
> 
> Closes #960

Hi @yashi, good point about updating the READMEs to clarify the senders. Thank you.

Issue #966, opened 2026-05-01
Security: RDP SYN options are parsed without payload length validation


ladderlogix:
## Description

When an RDP SYN is received in the closed state, `csp_rdp_new_packet()` reads six `uint32_t` option fields from `packet->data32[0]` through `packet->data32[5]`.

There is no check that the packet payload contains the full 24 byte option block before those reads.

## Impact

A short RDP SYN can cause out of bounds reads from the packet buffer. The parsed values are then stored into the connection RDP state, including `window_size`, `conn_timeout`, `packet_timeout`, `delayed_acks`, `ack_timeout`, and `ack_delay_count`.

This can cause incorrect connection behavior, long lived stale connections, or denial of service depending on the values read or supplied.

Provisional CVSS v3.1: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

CWE: `CWE-125`

## Affected versions

Observed on develop commit `4352143edba2af45afeb32d5b62ea5746b93a26a`.

Tag review indicates this is present from `v2.0` through `v2.1` and current develop. RDP was not present in the older `v1.x` tags I checked.

## Code locations

- `src/csp_rdp.c:528-533`

## Steps to Reproduce

1. Enable RDP support.
2. Deliver an RDP SYN with fewer than 24 bytes of SYN option payload.
3. Route the packet into `csp_rdp_new_packet()` while the connection is in the closed state.
4. The function reads six option words without first validating the payload length.

## Expected behavior

The RDP SYN path should reject packets shorter than the full SYN option block before parsing any option field.

## Actual Behavior

The code reads all six option words unconditionally.

## Environment Details

- Build OS: source review
- Target OS: targets with `CSP_USE_RDP`
- libcsp commit hash: `4352143edba2af45afeb32d5b62ea5746b93a26a`
- Meson version: not applicable
- CMake version: not applicable
- GNU Make version: not applicable
- Ninja version: not applicable

## Additional Context

I searched open and closed issues and PRs for this RDP SYN option length issue and did not find an existing report.

A minimal fix is to check `packet->length >= sizeof(rdp_header_t) + 6 * sizeof(uint32_t)` before reading `packet->data32[0..5]`.

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
[...]

Issue #969, opened 2026-05-01
Security: HMAC verification uses a 4 byte tag and memcmp


ladderlogix:
## Description

The HMAC verification path truncates the HMAC to 4 bytes and compares it with `memcmp()`.

There is also no guard that prevents HMAC append or verification before `csp_hmac_set_key()` has initialized the static key. In that case the static key storage starts as zero initialized memory.

## Impact

This weakens packet authentication in deployments that rely on CSP HMAC. The 4 byte tag gives a small authentication margin for long lived systems, and `memcmp()` is not appropriate for comparing authentication tags because it can return early.

Provisional CVSS v3.1: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

CWE: `CWE-347`

## Affected versions

Observed on develop commit `4352143edba2af45afeb32d5b62ea5746b93a26a`.

Tag review indicates the 4 byte tag, `memcmp()` comparison, and static HMAC key storage are present from `v1.0` through `v2.1` and current develop.

## Code locations

- `include/csp/crypto/csp_hmac.h:19`
- `src/crypto/csp_hmac.c:15`
- `src/crypto/csp_hmac.c:168`
- `src/crypto/csp_hmac.c:180`

## Steps to Reproduce

1. Review `CSP_HMAC_LENGTH` in `include/csp/crypto/csp_hmac.h`.
2. Review the verification branches in `src/crypto/csp_hmac.c`.
3. Observe that only 4 bytes are checked and that `memcmp()` is used for the comparison.
4. Review the static key storage and note that append and verify do not require a prior successful key setup.

## Expected behavior

Authentication tag comparison should be constant time, the tag length should provide an appropriate security margin, and HMAC use should fail until a key has been explicitly configured.

## Actual Behavior

The code compares a 4 byte tag with `memcmp()` and can operate with the zero initialized static key if no key is configured.

## Environment Details

- Build OS: source review
- Target OS: targets using `CSP_USE_HMAC`
- libcsp commit hash: `4352143edba2af45afeb32d5b62ea5746b93a26a`
- Meson version: not applicable
- CMake version: not applicable
- GNU Make version: not applicable
- Ninja version: not applicable

## Additional Context

I searched open and closed issues before filing. I did not find a report for the HMAC tag length, comparison, and key initialization behavior.

A candidate fix would use a constant time comparison, reject HMAC operations before key setup, reject empty keys, and consider a longer tag for new deployments.

[...]

Issue #968, opened 2026-05-01
Security: CSP_REBOOT uses a public magic value without authentication


ladderlogix:
## Description

The built in `CSP_REBOOT` service accepts a public magic word and then calls `csp_reboot_hook()` or `csp_shutdown_hook()`.

The magic word is a guard against accidental reboot requests, but it is not authentication. If this service is reachable from an untrusted link, a remote peer can request reboot or shutdown.

## Impact

The impact is availability. A remote peer that can reach the service can reboot or shut down the node if the deployment exposes this service across an untrusted boundary.

Provisional CVSS v3.1: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

CWE: `CWE-306`

## Affected versions

Observed on develop commit `4352143edba2af45afeb32d5b62ea5746b93a26a`.

Tag review indicates the reboot magic behavior is present from `v1.0` through `v2.1` and current develop.

## Code locations

- `src/csp_service_handler.c:295-310`

## Steps to Reproduce

1. Register the built in `csp_service_handler()`.
2. Send a packet to the `CSP_REBOOT` service with the documented reboot or shutdown magic value.
3. Observe that the service calls the reboot or shutdown hook.

## Expected behavior

Reboot and shutdown should require an explicit authentication decision, not only a public magic value.

## Actual Behavior

The public magic value is sufficient if the packet reaches the service.

## Environment Details

- Build OS: source review
- Target OS: targets exposing `CSP_REBOOT`
- libcsp commit hash: `4352143edba2af45afeb32d5b62ea5746b93a26a`
- Meson version: not applicable
- CMake version: not applicable
- GNU Make version: not applicable
- Ninja version: not applicable

## Additional Context

I searched open and closed issues before filing. I did not find a specific issue for `CSP_REBOOT` authentication.

I am not claiming this is unsafe for deployments where all CSP traffic is already authenticated and trusted. The concern is that the default service behavior does not enforce that boundary itself.

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

johandc:
[...]

Issue #967, opened 2026-05-01
Security: callback services do not require HMAC


ladderlogix:
## Description

Packets delivered through the callback path in `csp_route_work()` are checked with `CSP_SO_CRC32REQ`, but not with the socket or service option that would require HMAC.

This means built in service callbacks can be reached without HMAC even when the deployment expects authenticated service traffic.

## Impact

If the built in services are registered through the callback path and the network is reachable from an untrusted peer, management services may be reachable without the expected HMAC requirement.

Provisional CVSS v3.1: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H`

CWE: `CWE-306`

## Affected versions

Observed on develop commit `4352143edba2af45afeb32d5b62ea5746b93a26a`.

Tag review indicates this callback path issue is present in `v2.0` through `v2.1` and current develop.

## Code locations

- `src/csp_route.c:181-191`

## Steps to Reproduce

1. Register a built in service through the callback path.
2. Send a packet to that service without `CSP_FHMAC`.
3. Observe that the callback path calls `csp_route_security_check(CSP_SO_CRC32REQ, ...)`.
4. The check requires CRC32 but does not require HMAC.

## Expected behavior

The callback path should enforce the same authentication policy that the deployment expects for service traffic.

## Actual Behavior

The callback path uses `CSP_SO_CRC32REQ` and does not require HMAC.

## Environment Details

- Build OS: source review
- Target OS: targets using callback registered services
- libcsp commit hash: `4352143edba2af45afeb32d5b62ea5746b93a26a`
- Meson version: not applicable
- CMake version: not applicable
- GNU Make version: not applicable
- Ninja version: not applicable

## Additional Context

I searched open and closed issues before filing. I did not find a report for the callback path HMAC requirement bypass.

This appears related to the broader issue of secure by default management service handling.

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

johandc:
[...]

PR #970, opened 2026-05-01
Validate RDP packet lengths

### LINKED ISSUE #963: Security: short RDP packets are processed before header length validation
## Description

`csp_rdp_new_packet()` takes a reference to the RDP header before checking that the packet is long enough to contain one. The helper `csp_rdp_header_ref()` computes the header pointer as `packet->data[packet->length - sizeof(*header)]`.

If `packet->length` is less than the RDP header size, the unsigned subtraction underflows and the returned pointer is outside the packet payload. The caller then reads and writes through that pointer while converting `ack_nr` and `seq_nr`.

## Impact

A short packet with the RDP flag set can cause out of bounds memory access in the router task. On targets using the static CSP buffer pool, this can corrupt adjacent packet buffer state or crash the node.

Provisional CVSS v3.1: `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H`

CWE: `CWE-191`

## Affected versions

Observed on develop commit `4352143edba2af45afeb32d5b62ea5746b93a26a`.

Tag review indicates this is present from `v2.0` through `v2.1` and current develop. RDP was not present in the older `v1.x` tags I checked.

## Code locations

- `src/csp_rdp.c:76-79`
- `src/csp_rdp.c:441-443`

## Steps to Reproduce

1. Enable RDP support.
2. Deliver a CSP packet with the RDP flag set and a CSP payload shorter than `sizeof(rdp_header_t)`.
3. Route the packet into `csp_rdp_new_packet()`.
4. The function calls `csp_rdp_header_ref()` before validating the length and then reads and writes the derived header fields.

## Expected behavior

`csp_rdp_new_packet()` should reject packets shorter than `sizeof(rdp_header_t)` before taking a header reference.

## Actual Behavior

The header pointer is derived from an underflowed length and is used before any minimum length check.

## Environment Details

- Build OS: source review
- Target OS: targets with `CSP_USE_RDP`
- libcsp commit hash: `4352143edba2af45afeb32d5b62ea5746b93a26a`
- Meson version: not applicable
- CMake version: not applicable
- GNU Make version: not applicable
- Ninja version: not applicable

## Additional Context

I searched open and closed issues and PRs for RDP short packet and header length handling. I did not find an existing report for this specific issue.

A minimal fix is to return early when `packet->length < sizeof(rdp_header_t)`.

## Checklist

- [x] I've searched in "Issues" for similar issues, both open and closed.
- [x] I've used GitHub Flavored Markdown (GFM) to format my text.
- [x] I've avoided using screenshots for console output or logs
- [x] I've included all relevant details and steps to reproduce the issue.
- [x] I've provided information about my environment.
- [x] I've attached necessary logs or code snippet for minimal reproducible example.

ladderlogix:
## Summary

This PR is narrowed to RDP packet length validation.

Changes included:

- Reject short RDP packets before taking an RDP header reference.
- Reject RDP SYN packets that do not contain the full option block.

Fixes #963 and #966.

## Compatibility notes

This revision does not change CSP authentication policy, HMAC behavior, buffer exhaustion behavior, CMP behavior, or wire format. It only rejects malformed RDP packets before parsing fields that are not present.

## Testing

- Meson library build with Python bindings disabled.
- CMake library build with samples and Python bindings disabled.

### DIFF
--- a/src/csp_rdp.c
+++ b/src/csp_rdp.c
@@ -27,6 +27,7 @@
 #define RDP_ACK 0x04
 #define RDP_EAK 0x02
 #define RDP_RST 0x01
+#define RDP_SYN_OPTIONS_LENGTH (6U * sizeof(uint32_t))
 
 #ifndef CSP_USE_RDP_FAST_CLOSE
 #define CSP_USE_RDP_FAST_CLOSE 1
@@ -437,6 +438,12 @@ bool csp_rdp_new_packet(csp_conn_t * conn, csp_packet_t * packet) {
 
 	bool close_connection = false;
 
+	if (packet->length < sizeof(rdp_header_t)) {
+		csp_rdp_error("RDP %p: Packet too short for RDP header\n", (void *)conn);
+		csp_buffer_free(packet);
+		return false;
+	}
+
 	/* Get RX header and convert to host byte-order */
 	rdp_header_t * rx_header = csp_rdp_header_ref(packet);
 	rx_header->ack_nr = be16toh(rx_header->ack_nr);
@@ -512,6 +519,11 @@ bool csp_rdp_new_packet(csp_conn_t * conn, csp_packet_t * packet) {
 			}
 
 			csp_rdp_protocol("RDP %p: SYN-Received\n", (void *)conn);
+
+			if (packet->length < (sizeof(rdp_header_t) + RDP_SYN_OPTIONS_LENGTH)) {
+				csp_rdp_error("RDP %p: SYN packet too short for options\n", (void *)conn);
+				goto discard_close;
+			}
 
 			/* Setup TX seq. */
 			unsigned int seed = csp_get_ms();


yashi:
Hi @ladderlogix,

Thanks for the issues and the PR. Would you mind splitting the PR into multiple commits?

ladderlogix:
I split the PR into focused commits by issue area: 
- HMAC verification: #969 
- RDP packet length validation: #963 and #966 
- Buffer exhaustion handling: #962
- Built in service authentication: #965, #967, and #968 
- CMP reply lengths: #964

ladderlogix:
@yashi broken up them

kopi3daniel:
merci pour informations

johandc:
Thanks for the contribution, but...

A lot of these look like AI generated socalled "Security" issues. - CSP is not a hardened protocol, and security is not in any way shape or form a concern. The Entire memory is accessible through "PEEK" and "POKE". So theres that. I an not at all against using AI for code review, and i can definitely see why an AI would flag common security issues, but we need to instruct the AI that CSP is not concerned about security as such.

Security for satellites need to be on the interface. I.e. over the air protocols is hardened by encryption on the radio. Not on the backbone, where stability and efficiency is more of concern.

I will walk through them later to see if any of these could be applied, without sacrifice.

src/csp_route.c
```diff
@@ -87,6 +87,11 @@ static int csp_route_security_check(uint32_t security_opts, csp_iface_t * iface,
 		iface->autherr++;
 		return CSP_ERR_HMAC;
 	}
+#else
```
johandc: It does not make sense to chek for HMAC if csp is compiled without HMAC

src/csp_route.c
```diff
@@ -181,7 +186,12 @@ int csp_route_work(void) {
 	csp_callback_t callback = csp_port_get_callback(packet->id.dport);
 	if (callback) {
 
-		if (csp_route_security_check(CSP_SO_CRC32REQ, input.iface, packet) != CSP_ERR_NONE) {
+		uint32_t callback_opts = CSP_SO_CRC32REQ;
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: HMACREQ is a connection based flag, so it is not controlled at compile time but by the service that creates a socket. So it should not be a compile time overrule for all connections.

src/csp_service_handler.c
```diff
@@ -11,6 +11,17 @@
 #include <csp/csp_id.h>
 #include <csp/arch/csp_time.h>
 
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
+static int csp_service_require_auth(csp_packet_t * packet) {
```
johandc: HMACREQ is a connection based flag, so it is not controlled at compile time but by the service that creates a socket. So it should not be a compile time overrule for all connections.

src/csp_service_handler.c
```diff
@@ -201,6 +212,12 @@ static int csp_cmp_handler(csp_packet_t * packet) {
 	int ret = CSP_ERR_INVAL;
 	struct csp_cmp_message * cmp = (struct csp_cmp_message *)packet->data;
 
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: We have csp_route_security_check for this

src/csp_service_handler.c
```diff
@@ -273,6 +295,12 @@ void csp_service_handler(csp_packet_t * packet) {
 			break;
 
 		case CSP_PS: {
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: This has already been checked by the incoming handler in route.c so dont add it inside the service handler too.

src/csp_service_handler.c
```diff
@@ -281,6 +309,12 @@ void csp_service_handler(csp_packet_t * packet) {
 		}
 
 		case CSP_MEMFREE: {
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: This has already been checked by the incoming handler in route.c so dont add it inside the service handler too.

src/csp_service_handler.c
```diff
@@ -293,6 +327,12 @@ void csp_service_handler(csp_packet_t * packet) {
 		}
 
 		case CSP_REBOOT: {
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: This has already been checked by the incoming handler in route.c so dont add it inside the service handler too.

src/csp_service_handler.c
```diff
@@ -310,6 +350,12 @@ void csp_service_handler(csp_packet_t * packet) {
 		}
 
 		case CSP_BUF_FREE: {
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: This has already been checked by the incoming handler in route.c so dont add it inside the service handler too.

src/csp_service_handler.c
```diff
@@ -318,6 +364,12 @@ void csp_service_handler(csp_packet_t * packet) {
 		}
 
 		case CSP_UPTIME: {
+#if (CSP_REQUIRE_AUTHENTICATED_SERVICES)
```
johandc: This has already been checked by the incoming handler in route.c so dont add it inside the service handler too.

CMakeLists.txt
```diff
@@ -30,9 +30,11 @@ set(CSP_BUFFER_SIZE 256 CACHE STRING "Bytes in each packet buffer")
 set(CSP_BUFFER_COUNT 15 CACHE STRING "Number of total packet buffers")
 set(CSP_RDP_MAX_WINDOW 5 CACHE STRING "Max window size for RDP")
 set(CSP_RTABLE_SIZE 10 CACHE STRING "Number of elements in routing table")
+set(CSP_HMAC_LENGTH 16 CACHE STRING "Number of HMAC bytes appended to authenticated packets")
 
 option(CSP_USE_RDP "Reliable Datagram Protocol" ON)
 option(CSP_USE_HMAC "Hash-based message authentication code" ON)
+option(CSP_REQUIRE_AUTHENTICATED_SERVICES "Require HMAC on built-in management services" ON)
```
johandc: I think i want to remove HMAC from CSP alltogether and make a document on how to secure the satellite network through encrypted interfaces instead.

src/csp_buffer.c
```diff
@@ -187,26 +187,15 @@ int csp_buffer_remaining(void) {
 }
 
 /* CSP will use every remaining buffer in an attempt to allocate a packet
- * buffer. Including the last two, that is normally reserved.
- * This can be important for ensuring a node is always reachable, so a
- * failure to allocate will cause a panic and a reboot */
+ * buffer, including the last two that are normally reserved. Callers must
+ * handle NULL and drop the frame when no buffers are available. */
```
johandc: Dont agree. We deliberately panic when the final two packets get used up. This is to safeguard against buffer starvation. Critical interfaces would not be able to work, if there is no more memory. So the system must reboot as a result, and to make the buffer situation known. The last thing a system should do is to reject incoming traffic.

src/interfaces/csp_if_can_pbuf.c
```diff
@@ -52,6 +52,9 @@ csp_packet_t * csp_can_pbuf_new(csp_can_interface_data_t * ifdata, uint32_t id,
 	uint32_t now = (task_woken) ? csp_get_ms_isr() : csp_get_ms();
 
 	csp_packet_t * packet = (task_woken) ? csp_buffer_get_always_isr() : csp_buffer_get_always();
+	if (packet == NULL) {
```
johandc: User should NOT check return on buffer_get_always. That is part of the contract with the function cannot return NULL.

src/interfaces/csp_if_can.c
```diff
@@ -50,6 +50,10 @@ static int csp_can1_rx(csp_iface_t * iface, uint32_t id, const uint8_t * data, u
 	if (packet == NULL) {
 		if (CFP_TYPE(id) == CFP_BEGIN) {
 			packet = csp_can_pbuf_new(ifdata, id, task_woken);
+			if (packet == NULL) {
```
johandc: csp_can_pbuf_new can NOT return NULL. it is guaranteed to provide a buffer, or panic the system. This is by design.

src/crypto/csp_hmac.c
```diff
@@ -90,6 +91,16 @@ static int csp_hmac_done(hmac_state * hmac, uint8_t * out) {
 	return CSP_ERR_NONE;
 }
 
+static int csp_hmac_constant_time_cmp(const uint8_t * a, const uint8_t * b, size_t len) {
```
johandc: This is a good idea 👍 

But HMAC is not strong enough for actual security. So i think i want to remove it from CSP, and provide better hooks for custom security frameworks.

include/csp/crypto/csp_hmac.h
```diff
@@ -16,7 +16,13 @@ extern "C" {
 /**
  * Number of bytes from the HMAC calculation, that is appended to the CSP message.
  */
-#define CSP_HMAC_LENGTH	4
+#ifndef CSP_HMAC_LENGTH
+#define CSP_HMAC_LENGTH	16
```
johandc: This would be breaking, and is not accepted.

johandc (changes requested):

ladderlogix:
Thank you for the review. I understand the model you are describing: CSP itself is not intended to be the security boundary, and authentication policy belongs at the interface or service configuration layer.

So because of that I reworked the PR around that. The branch now removes the global authenticated service option, the HMAC length change, the unset key behavior change, and the buffer exhaustion contract change.

What remains is malformed input handling and local memory safety work: RDP length checks, CMP reply lengths, pbuf cleanup traversal, CAN frame length validation, CSP 2 header field masking, bounded connection table formatting, and the non breaking HMAC compare cleanup.

johandc (commented):
Pbuf free was fixed here as well
https://github.com/libcsp/libcsp/commit/bfdb6e133a8758a2a5b9e423268c1c4caf61943e

ladderlogix:
Thanks, I missed that pbuf commit. I have dropped the overlapping pbuf cleanup from this PR.

I also removed the other cleanup items so this branch is now limited to the RDP length validation for #963 and #966. That keeps the PR away from the CSP security boundary discussion and from behavior changes outside the RDP parser.

PR #973, opened 2026-05-13
csp_conn: validate reply source to prevent cross-talk

dsabala:
Hello,

I observed a CSP node cross-talk issue caused by late replies landing in recycled connection slots, since connections are matched only by the reply destination port.

I'm adding dynamic node discovery to my application. The other nodes are passive listeners by contract, so the main node (ID 1) has to periodically poll all configured CSP node IDs (in my case 10, 11, 12, 13, 14, 15, 30) every second.

I'm currently testing with a single node, ID 30. When node 30 boots, its bootloader uses the CPU to compute some hashes, which takes a few seconds, and during that time the device's CAN peripheral queues at least two identification requests. Once the CPU is ready, it processes them. By then, however, the main node has already recycled the connection slot that used port 24, so the late responses to the original node 30 queries are parsed as responses from other nodes (e.g. node 11).

I'm attaching a Wireshark capture, please look at these packets:

- packet 591 (30.603s): node 1 queries node 30 for identification (port 18)
- packet 605 (31.606s): node 1 queries node 30 for identification (port 17)
- packet 619 (32.603s): node 1 queries node 30 for identification (port 24)
- packet 653 (35.105s): node 30, now several seconds after power-up, responds to the old identification query (port 18)
- packet 654 (35.105s): node 1 queries node 11 for identification (reuses port 24)
- packet 659 (35.106s): node 30 responds to the old identification query (port 17)
- packet 663 (35.107s): node 30 responds to the identification query on port 24, and node 1 parses it as a response from node 11

For reference, Linux does the same thing for connected UDP sockets: after connect(2) on SOCK_DGRAM the kernel only delivers datagrams from the connected peer (see compute_score() in net/ipv4/udp.c, or the connect(2) man page). This patch makes CONN_CLIENT behave the same way, while keeping the broadcast exception.

The fix doesn't fully eliminate the race: a late reply from node X can still be matched to a new connection also targeting X if the slot was recycled in between. Closing that would require an application-level transaction ID, which CSP doesn't provide at the transport layer.

[cross-talk-test.zip](https://github.com/user-attachments/files/27715927/cross-talk-test.zip)

### DIFF
--- a/src/csp_conn.c
+++ b/src/csp_conn.c
@@ -90,7 +90,7 @@ csp_conn_t * csp_conn_find_dport(unsigned int dport) {
 	return NULL;
 }
 
-csp_conn_t * csp_conn_find_existing(csp_id_t * id) {
+csp_conn_t * csp_conn_find_existing(csp_id_t * id, csp_iface_t * iface) {
 
 	for (int i = 0; i < CSP_CONN_MAX; i++) {
 		csp_conn_t * conn = &arr_conn[i];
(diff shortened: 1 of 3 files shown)


johandc:
Excellent. I completely agree in matching on source ids as well. I dont know yet if calling `csp_id_is_broadcast` would catch all exceptions. But lets find out and test it.

dsabala:
Thanks, glad it makes sens to you :)

One caveat: I cant reliably test broadcast branch myself, my setup is unicast only, so this branch hasnt been exercised in practice on my end. If you spot any edge cases during tests, I am happy to iterate.

dsabala:
@johandc Hello, just a ping to see if you've had a chance to test this out, or if you need anything else from my side?

johandc:
No, but we might move it to spaceinventor/develop to test it out there.

Issue #753, opened 2024-12-02
SFP is disabled in python bindings


lu-maca:
Hi,

Is there a particular reason why csp_sfp_* functions are not compiled in the [python bindings](https://github.com/libcsp/libcsp/blob/v2.0/src/bindings/python/pycsp.c#L206)?

Thanks,
Luca

yashi:
maybe because nobody has implemented yet.

lu-maca:
Implementation is there, look at the link I put in the first post:

```c
#if 0
static PyObject* pycsp_sfp_send(PyObject *self, PyObject *args) {
    PyObject* conn_capsule;
    Py_buffer data;
    unsigned int mtu;
    uint32_t timeout = 1000;
    if (!PyArg_ParseTuple(args, "Ow*II|I", &conn_capsule, &data, &mtu, &timeout)) {
        return NULL;
    }

    csp_conn_t* conn = get_obj_as_conn(conn_capsule, false);
    if (conn == NULL) {
        return NULL;
    }

    int res;
    Py_BEGIN_ALLOW_THREADS;
    res = csp_sfp_send(conn, data.buf, data.len, mtu, timeout);
    Py_END_ALLOW_THREADS;
    if (res != CSP_ERR_NONE) {
        return PyErr_Error("sfp_send()", res);
    }

    return Py_BuildValue("i", res);
}

static PyObject* pycsp_sfp_recv(PyObject *self, PyObject *args) {
    PyObject* conn_capsule;
    uint32_t timeout = 500;
    if (!PyArg_ParseTuple(args, "O|I", &conn_capsule, &timeout)) {
        return NULL; // TypeError is thrown
    }
    csp_conn_t* conn = get_obj_as_conn(conn_capsule, false);
    if (conn == NULL) {
        return NULL;
    }
    void *dataout = NULL;
    int return_datasize = 0;
    int res;
    Py_BEGIN_ALLOW_THREADS;
    res = csp_sfp_recv(conn, &dataout, &return_datasize, timeout);
    Py_END_ALLOW_THREADS;

    if (dataout == NULL) {
        Py_RETURN_NONE;
    }

    if (res != CSP_ERR_NONE) {
        return PyErr_Error("sfp_recv()", res);
    }

    return PyCapsule_New(dataout, PACKET_CAPSULE, pycsp_free_csp_buffer);
}
#endif
```

yashi:
[...]

PR #974, opened 2026-05-25
pycsp: implement new SFP functions in python bindings 

LewentalDigital:
Python bindings for SFP send and receive functions were disabled with a `#if 0` and commented out of the python methods list. This commit enables and implements the `sfp_send` and `sfp_recv` functions in pycsp; compatible with the new SFP rework.

This PR solves #753

### DIFF
--- a/src/bindings/python/pycsp.c
+++ b/src/bindings/python/pycsp.c
@@ -209,60 +209,90 @@ static PyObject * pycsp_send(PyObject * self, PyObject * args) {
 	Py_RETURN_NONE;
 }
 
-#if 0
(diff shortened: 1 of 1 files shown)


johandc:
Why were they disabled in the first place?

LewentalDigital:
> Why were they disabled in the first place?

I don't know.
The implementation was almost functioning prior to the SFP rework, only a small error in binding arguments types; maybe because of this they were commented and waiting for a fix.

I can provide my fix for release v2.1 or other in another PR. It would also come in handy for me.

Issue #956, opened 2026-04-13
csp_can_pbuf_free() going into an endless loop


brookbot:
## Description
csp_can_pbuf_free() is going into an endless loop.
I'm not exactly sure what sequence of evens cause this, but its clear that this function is looping forever.
` 13   Thread 0x7e96820186c0 (LWP 125) "core-cpu2"       0x00007e968282a7ac in csp_can_pbuf_free () from ./apps/cfs_io_lib.so`
and it seems to be happening on our simulated CAN bus rx thread and is triggered by various tests and rather inconsistent
```
[Switching to thread 13 (Thread 0x7e96820186c0 (LWP 125))]
#0  0x00007e968282a7ac in csp_can_pbuf_free () from ./apps/cfs_io_lib.so
(gdb) bt
#0  0x00007e968282a7ac in csp_can_pbuf_free () from ./apps/cfs_io_lib.so
#1  0x00007e9682829cb4 in csp_can1_rx () from ./apps/cfs_io_lib.so
#2  0x00007e968282a6f9 in csp_can_rx () from ./apps/cfs_io_lib.so
#3  0x00007e96828291e1 in socketcan_rx_thread () from ./apps/cfs_io_lib.so
#4  0x00007e9682be9aa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#5  0x00007e9682c76a64 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:100
```

## Steps to Reproduce

I've instrumented the code to print both the address of the buffer being freed and the packet addresses its checking. This produced output such as:
```
CSP: Freeing buffer 0x7c7f13e535a0
CSP: Checking packet 0x7c7f13e535a0
CSP: Freeing buffer 0x7c7f13e53300
CSP: Checking packet 0x7c7f13e53060
CSP: Checking packet 0x7c7f13e52c70
CSP: Checking packet 0x7c7f13e52f10
CSP: Checking packet 0x7c7f13e53450
CSP: Checking packet 0x7c7f13e53840
CSP: Checking packet 0x7c7f13e53990
CSP: Checking packet 0x7c7f13e53300
```
Which looks fine - although I notice that it continues checking packet addresses after the buffer has been freed.
But then I get the following output:
```
CSP: Checking packer 0x7b9026e3e450
CSP: Checking packer 0x7b9026e3e450
CSP: Checking packer 0x7b9026e3e450
CSP: Checking packer 0x7b9026e3e450
[...]

PR #977, opened 2026-06-08
Correct possible use-after-free conditions in PBUFs

troelsjessen:
This is fixing #956


johandc:
Thanks! This one is included in the si-sync-2026-06

Issue #975, opened 2026-05-28
Feature Request: Support for mixed CSP versions (v1/v2) on the same node


Rhyn-muller:
Hi.

I’m working on a comms relay hardware that sits on a mixed-version network. It runs CSP v2 natively but bridges CAN buses where legacy v1.6 nodes still live. Because it acts as a relay, it inherently needs to speak multiple versions simultaneously. 

Currently, `csp_conf.version` is process-global. While it's technically possible to switch the global version right before each send, it causes a few concrete issues for a bridge/relay setup:
* **Thread-safety:** It creates a race condition between two tasks trying to send to different-version destinations at the same time. Two tasks can interleave a version-1 send and a version-2 send, producing a packet with the wrong header.
* **Complexity:** It requires callers to know the destination's version and carefully manage the toggle externally.
* **Future-proofing:** It breaks if CSP ever uses `csp_conf.version` in an async path (e.g., retransmissions).

### Proposed fix: Per-interface version field
A nice solution given the existing API shape might be adding an override to the interface struct:

```c
// csp_interface.h
struct csp_iface_s {
    uint16_t addr;
    uint16_t netmask;
    const char * name;
    void * interface_data;
    void * driver_data;
    nexthop_t nexthop;
    uint8_t is_default;
    uint8_t version;   // <-- add: 0 = inherit from csp_conf, 1 or 2 = override
    /* ... stats, next ... */
};
```

Then, in `csp_id_prepend()` (or the nexthop call chain), the iface version can be evaluated `iface->version ?: csp_conf.version`. If it's zero, global default can be used "use global default", and non-zero overrides it. This is strictly backwards-compatible, as existing code that never sets `iface->version` should keep working exactly as it does today.

### Alternative: Route table version
If you prefer to keep the interface struct clean, an alternative could be to have the `csp_rtable` entry carry a version hint per destination address. Relay nodes already know which destinations are v1 vs v2 from their routing tables anyway.

Please let me know if there is a reason why version is global-only which I'm missing and if the per-interface override PR is something that would be accepted. Thanks!

johandc:
Interesting we are working in exactly that at the moment. The solution we are implementing is per-interface version field, along with some other fixes where id strip and prepend were using globals.

Rhyn-muller:
> Interesting we are working in exactly that at the moment. The solution we are implementing is per-interface version field, along with some other fixes where id strip and prepend were using globals.

[...]

PR #982, opened 2026-06-18
Version field added to csp_packet_t and csp_iface_t

hendav76:
This can be used for having support for both CSPv1 and CSPv2 at the same time on different interfaces. Only implemented for CAN and the csp_id_ functions that CAN uses.

### DIFF
--- a/include/csp/csp_id.h
+++ b/include/csp/csp_id.h
@@ -17,11 +17,28 @@ void csp_id_prepend(csp_packet_t * packet);
 
 /**
  * Extract CSP header from a 4 (CSP1) or 6 (CSP2) byte array.
+ * Uses csp_conf.version to determine the CSP version
  *
  * @param The first bytes of a CSP packet, representing the CSP header structure.
  * @return The extracted header ID structure.
  */
 csp_id_t csp_id_extract(const uint8_t * data);
+
+/**
+ * Extract CSP header from a 4 (CSP1) byte array.
+ *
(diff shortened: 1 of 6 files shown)


yashi:
@hendav76  would you please rebase on `develop`?

yashi:
Hi @hendav76,

I might be missing something, but libcsp does not currently support selecting the CSP version on a per-interface basis, does it?

In #975, @johandc said that CSP v1 and v2 can be mixed, but I cannot find where per-interface version selection is currently implemented.

Could you clarify the intention behind this PR?

hendav76:
Hi @yashi 

The concept is that the interface is responsible for setting the version field in the csp_iface_t struct. Consequently, csp_if_can.c (which is the only interface updated in this pull request) will set the packet's version field to match the version of the receiving interface. This allows the functions csp_id_prepend, csp_id1_extract, csp_id2_extract, csp_id_strip, and csp_id_setup_rx to handle both CSP version 1 and version 2 packets. Furthermore, csp_if_can.c will call the appropriate functions csp_can1_rx, csp_can2_rx, csp_can1_tx, or csp_can2_tx depending on the version of the interface.

To support both CSP version 1 and version 2 concurrently, the following must be implemented outside of libcsp:
 - The interface implementation must explicitly set the version field (if not set csp_conf.version will be used) and call either csp_id1_extract or csp_id2_extract depending on the version.
 - CSP addresses must be converted between version 1 and version 2 if needed.

PR #991, opened 2026-07-02
drivers: eth: eth_linux: Initialize SO_REUSEADDR option

yashi:
Initialize the integer value passed to setsockopt() for SO_REUSEADDR.

SO_REUSEADDR expects an integer boolean option value. Set it to 1 so address reuse is enabled instead of passing an uninitialized value.


yashi:
@troelsjessen Would you please take a look?

troelsjessen (approved):

PR #993, opened 2026-07-06
drivers: usart_linux: Add termios2 support for arbitrary baud rates

clorch:
Use termios2 with BOTHER on Linux to allow configuring arbitrary USART baud rates in csp_usart_open(). This removes the previous limitation to the fixed set of baud rate constants exposed through termios and makes it possible to use device-specific or non-standard serial speeds.

Keep the existing termios-based path for systems where termios2 is not available. On those platforms, baud rate selection still uses the existing mapping to standard termios baud rate constants and rejects unsupported values as before.

termios2 has been available since Linux 2.6.20 from 2007. You could also consider dropping the support for kernels before that i order to simplify the code.


src/drivers/usart/usart_linux.c
```diff
@@ -13,6 +13,30 @@
 #include <csp/csp.h>
 #include <pthread.h>
 
+#if defined(__linux__)
+#  include <sys/ioctl.h>
+#  define CSP_HAVE_TERMIOS2 1
+/* Define termios2 without pulling in <asm/termbits.h> to avoid
+   struct termios redefinition conflict with <termios.h> */
+struct termios2 {
```
yashi: Why are you defining `struct termios2` instead of including `termbits.h`? You said _to avoid struct termios redefinition conflict_ but that's exactly what you are doing, "redefining".

src/drivers/usart/usart_linux.c
```diff
@@ -137,13 +162,40 @@ int csp_usart_open(const csp_usart_conf_t * conf, csp_usart_callback_t rx_callba
...
+	options.c_cflag |= CS8;
+	options.c_lflag &= ~(ECHO | ECHONL | ICANON | IEXTEN | ISIG);
+	options.c_iflag &= ~(IGNBRK | BRKINT | ICRNL | INLCR | PARMRK | INPCK | ISTRIP | IXON);
+	options.c_oflag &= ~(OCRNL | ONLCR | ONLRET | ONOCR | OFILL | OPOST);
+	options.c_cc[VTIME] = 0;
+	options.c_cc[VMIN] = 1;
+	if (ioctl(fd, TCSETS2, &options) != 0) {
+		csp_print("%s: Failed to set attributes on device: [%s], errno: %s\n", __func__, conf->device, strerror(errno));
+		close(fd);
+		return CSP_ERR_DRIVER;
```
yashi, line 105: So many dup code.

src/drivers/usart/usart_linux.c
```diff
@@ -13,6 +13,30 @@
 #include <csp/csp.h>
 #include <pthread.h>
 
+#if defined(__linux__)
```
yashi: This file is for linux. I don't think you need this.

src/drivers/usart/usart_linux.c
```diff
@@ -13,6 +13,30 @@
...
+#  include <sys/ioctl.h>
+#  define CSP_HAVE_TERMIOS2 1
```
yashi: This means you are forcing all linux compilation to use `CSP_HAVE_TERMIOS2`.  `CSP_HAVE_TERMIOS2` should be defined by the build systems.

yashi (changes requested):

src/drivers/usart/usart_linux.c
clorch: You cannot include `temios.h` and `asm/termbits.h` at the same time, because if you do this, you get a redefinition conflict. That is why i chose to define everything I need from termios2 here manually. This is also the same approach as e.g. [picocom](https://github.com/npat-efault/picocom/blob/master/termbits2.h) chose.

That being said, it can also be done differently. As I said in my initial comment, we could also consider dropping support for old termios entirely which then simplifies the code a lot.

src/drivers/usart/usart_linux.c
clorch: Yes, you are right. My reasoning here was that the code is overall cleaner if I completely separate the termios and termios2 paths. If you do not like it, then I can either (as said above) drop termios support or do a mixed approach and reducing the duplicate lines by having more `#ifdef` sections.

src/drivers/usart/usart_linux.c
clorch: Ok, great, I'll remove it.

src/drivers/usart/usart_linux.c
clorch: Yes, correct. In the end this is your decision. I can elevate the decision between termios and termios2 to the build system or drop support for termios. What do you prefer?

src/drivers/usart/usart_linux.c
yashi: I hope you did not copy or adapt any code from picocom. Picocom is licensed under GPLv2-or-later, while libcsp is licensed under MIT. We cannot accept code derived from picocom unless it is available under a compatible license.

Leaving picocom aside, why not include the appropriate header conditionally? `<termios.h>` and `<asm/termbits.h>` do not need to be included in the same build:

```c
#ifdef CSP_HAVE_TERMIOS2
#include <asm/termbits.h>
#else
#include <termios.h>
#endif
```

The build system can select the implementation, so there is no need to copy the architecture-dependent definition of `struct termios2` into libcsp.

src/drivers/usart/usart_linux.c
yashi: we can do something like https://gist.github.com/yashi/b26baf0e1d6912b193ce5cc98dda2c26

But if you want to separate completely, we should have another file for termios2. we can default to termiso2 but if user wants, we can enable or switch to POSIX termios. WDYT?

yashi: I'd rather make `src/drivers/usart/usart_linux.c` with termios2 and rename `src/drivers/usart/usart_linux.c` to `src/drivers/usart/usart_posix.c`.

src/drivers/usart/usart_linux.c
clorch: No, I did not copy anything from picocom. I was just looking for some project that does the same thing as an example.

clorch:
As you suggested, I have now migrated the USART driver in `usart_linux.c` to termios2 and also include `asm/termbits.h` there. The old driver is now in (`usart_posix.c`) and can be enabled through a build option. The new driver is the default. I have added this build option to waf, cmake and meson.

Some checks failed with "Error: Service Unavailable". This probably has nothing to do with my code.

yashi (changes requested):
Thanks. I've rerun the checks, and they pass now. GitHub Actions had an outage yesterday, so that likely caused the failure.

As for the code, I'd like to use `termios2` by default on Linux. It was added in Linux 2.6.20 in 2007, and I don't think we need to support kernels that old.

We should also provide a compile-time option, `CSP_USE_POSIX_TERMIOS`, for non-Linux systems. This means that we can unconditionally select `usart/usart_linux.c` when `host_machine.system() == 'linux'`.

Thank you for the time you've spent on this.

clorch:
I agree that termios2 should be the default on Linux. I think this is also how I have implemented it in the latest commit. The three different build systems handle this somewhat differently now. In waf there is already the `--with-driver-usart` option, so it felt natural to add a third option there. For meson and cmake I added the `have_termios2` (meson) or `CSP_HAVE_TERMIOS2` (cmake) options with true being the default value. But I am also not really experienced with how you handle the build systems in this projects so I can also change this to something else if that fits better, of course.

yashi:
Thanks. What I had in mind is a `CSP_USE_POSIX_TERMIOS` override, rather than a `HAVE_TERMIOS2` option.

For CMake, it would be a Boolean option that is disabled by default. A user could enable it with:

```console
cmake -DCSP_USE_POSIX_TERMIOS=1 ...
```

It would also be available through `ccmake` or `cmake-gui`.

For Meson, the corresponding `use_posix_termios` option would be disabled by default and could be changed with `meson configure`, for example:

```console
meson configure -Duse_posix_termios=true build
```

For Waf, it should be a separate option from `--with-driver-usart`, since that option selects whether and which USART driver is built.

The normal source selection should be automatic. Linux should use `usart_linux.c` with `termios2`, while non-Linux POSIX systems should use `usart_posix.c`. Enabling `CSP_USE_POSIX_TERMIOS` should force Linux to use `usart_posix.c`; on non-Linux POSIX systems, it would effectively be a no-op.

If this is too much to ask, I can handle all three build systems and prepare the tree so that your change can be merged easily. I can rename the current `usart_linux.c` to `usart_posix.c`, update the build systems, merge your new `usart_linux.c`, and make it the default on Linux.

Let me know how I can help!

clorch:
I have now changed the build options as you proposed:
- for cmake it is now `CSP_USE_POSIX_TERMIOS` instead of `CSP_HAVE_TERMIOS2`
- for meson `use_posix_termios` instead of `have_termios2`
- for waf `--with-driver-usart=posix --use-posix-termios` instead of `--with-driver-usart=posix` and `--with-driver-usart=posix` instead of `--with-driver-usart=linux`

PR #984, opened 2026-06-22
fix: compilation error when CSP_ENABLE_CSP_PRINT=0

LewentalDigital:
As reported in #947, compiling with CSP_ENABLE_CSP_PRINT set to false fails.

The reason for the fail is that compiler reports unused arguments in functions that use `csp_print` as when CSP_ENABLE_CSP_PRINT=0 `csp_print` is a no-op and arguments passed to it by other functions are unused.

To fix this I changed the disabled macro implementation in `include/csp_debug.h` to:
```
do { if (0) csp_print_func(__VA_ARGS__); } while (0)
```
and `csp_hex_dump` in `include/csp.h` to: 
```
inline void csp_hex_dump(const char *desc, void *addr, int len) { (void)desc; (void)addr; (void)len; }
```
to keep the arguments visible to the compiler but do nothing with them.

PR #994, opened 2026-07-08
github: workflow: build-test: Add print options test

LewentalDigital:
As requested in https://github.com/libcsp/libcsp/pull/984#issuecomment-4864768204 I added a new test to github workflow `build-test.yml` that checks failures in compilations with options `CSP_ENABLE_CSP_PRINT` and `CSP_HAVE_STDIO`.

To cover all cases, the tests compiles with all combinations enabled/disabled of the two options.

Since `buildall.py` script used in the workflow does not accept extra options the test only compiles with meson.

### DIFF
--- a/.github/workflows/check-compile-options.yml
+++ b/.github/workflows/check-compile-options.yml
@@ -1,0 +1,34 @@
+name: Check Compile Options
+on: [push, pull_request]
+permissions:
+  contents: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+jobs:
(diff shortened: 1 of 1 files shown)


yashi:
I don't think it's a good place to put it because the following tests depends on the plain build. Maybe it's a good idea to create yet another workflow for various compile time options. WDYT?

LewentalDigital:
Yeah I get it. I can define a `check-compile-options.yml` that tests compilation with a defined set of options in the matrix.
That would also be useful for other tests with options.

yashi:
@LewentalDigital Thank you for the test! I've already merged last night and I'm sure we can add more tests on the workflow. Thanks!

PR #996, opened 2026-07-10
Refactor csp service handler

yashi:
This PR cleans up `csp_service_handler()`. It is a pure refactoring, and no functionality has been changed.


yashi:
@gigioneggiando Would you mind to take a look?

gigioneggiando (approved):
LGTM: faithful, behavior-preserving refactor, and `set_u32_reply()` is a nice cleanup.

The only real change is dropping the final `if (packet != NULL)` guard, which is safe: `packet` is already dereferenced at the top in `switch (packet->id.dport)`, so a NULL would crash on entry anyway and no case reassigns it. Types line up too (good explicit cast on `csp_buffer_remaining()`). I don't have anything else to add!

PR #978, opened 2026-06-11
More python bindings

guillaume-jcb:
Hi,

This pull request makes two kind of changes:
* Set the SOVERSION on the compiled library. This is needed to be able to package the library in a Yocto recipe.
* Add some missing python bindings, in particular those necessary to register a custom interface and many functions to inspect / manipulate the content of a packet.

I hope that these changes may benefit other user; let me know if you can think of any improvements.

PR #997, opened 2026-07-11
Add soversion

yashi:
This PR also set SO Name for waf and meson build.

This is related to a patch from #978.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -101,6 +101,7 @@ csp_lib = library('csp',
 	c_args : csp_c_args,
 	install : true,
 	pic:true,
+	version : meson.project_version(),
 )
 
 # The following dependency variable is for parent projects to link
--- a/wscript
+++ b/wscript
@@ -245,7 +245,8 @@ def build(ctx):
                   name='csp_shlib',
                   target='csp',
                   use=['csp_h', 'util_shlib'],
-                  lib=ctx.env.LIBS)
+                  lib=ctx.env.LIBS,
+                  vnum=VERSION)
 
     # Build Python bindings
     if ctx.env.HAVE_PYEXT:


meson.build
```diff
@@ -101,6 +103,8 @@ csp_lib = library('csp',
 	c_args : csp_c_args,
 	install : true,
 	pic:true,
+	version : csp_version,
+	soversion : csp_soversion,
```
guillaume-jcb: If left unspecified, `soversion` is set to the major version number by default: 

https://mesonbuild.com/Reference-manual_functions_library.html#library_soversion

So it _needs_ not be set, but I guess it does not hurt to make it explicit.

wscript
```diff
@@ -245,7 +245,9 @@ def build(ctx):
                   name='csp_shlib',
                   target='csp',
                   use=['csp_h', 'util_shlib'],
-                  lib=ctx.env.LIBS)
+                  lib=ctx.env.LIBS,
+                  vnum=VERSION,
+                  cnum=VERSION.split('.')[0])
```
guillaume-jcb: Likewise, it seems that `cnum` is set to `vnum.split('.')[0]` by default as long as `vnum` is sepcified.

https://github.com/ArduPilot/waf/blob/35eadbb64e2052099a853b571e507c33032b392c/waflib/Tools/ccroot.py#L592

guillaume-jcb (approved):
Thanks for taking care of this :-)

LewentalDigital:
Hi, sorry to revive this but so version for CMake is still missing right? CMake build does not output libcsp.so.2 and libcsp.so.2.2, CMakeLists.txt its missing `set_target_properties(... SOVERSION ...)`.

I was reading https://github.com/libcsp/libcsp/pull/978#issue-4641115233 since I use Yocto; I had to work around the missing so version to get a working Yocto libcsp recipe; so version in cmake would benefit to me.

Can we implement so version for CMake too?

yashi:
Yes. #978 includes the CMake fix, and it has now been merged. :wink:

PR #998, opened 2026-07-14
cmake: add install rule for libcsp_py3 python bindings

LewentalDigital:
CMakeLists is missing the rule to install libcsp_py3 target.

This PR introduces it and enables its installation in correct python packages directory by running `cmake --install build/`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.20)
 project(CSP VERSION 2.2)
+include(GNUInstallDirs)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set(DEFAULT_BUILD_SHARED_LIBS ON)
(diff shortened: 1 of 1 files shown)


yashi (changes requested):
I assume you added `Interpreter` to obtain `Python3_SITEARCH`, but I do not think `Python3_SITEARCH` is appropriate here. It is normally an absolute path, so using it as the installation destination prevents the installation from being relocated with:

```console
cmake --install build --prefix /tmp/libcsp-install
```

Could you please use a prefix-relative destination instead?

If possible, could you add a separate commit that tests both:

```console
cmake --install build
cmake --install build --prefix /tmp/libcsp-install
```

For the first case, the build can be configured with a temporary `CMAKE_INSTALL_PREFIX`. The tests should also verify that the module was installed under the expected prefix, so that they catch an absolute destination silently ignoring `--prefix`.

Also, if Interpreter remains a requested component, please check Python3_FOUND rather than only Python3_Development.Module_FOUND. Otherwise, configuration may continue even when the interpreter was not found.

LewentalDigital:
I looked into this a bit and there is no easy, simple way, without trade offs to manage python library installation with relative path.

A valid solution is to install it to 
```
${CMAKE_INSTALL_LIBDIR}/python${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}/site-packages
```
which is a relative path that works with `--prefix`; but the trade off is that python destination is incompatible with Debian distros that install python libraries to `dist-packages` instead of `site-packages`.

This does not need `Interpreter` but uses `include(GNUInstallDirs)`.

What do you think @yashi?
Also, in which workflow should I add the tests?

Issue #944, opened 2026-02-23
CANFD support


flotter:
Hi.

A quick search on the latest development branch shows that the CAN frame size is hardcoded to 8 bytes.

I am planning on adding CAN FD support. Are there any thoughts on how this should be done to get support merged back ?

Should CSP support mixing of CAN and CANFD interfaces, which feels like some frame fragmentation/reassembly would need to be supported. Or, is it OK to start with enforcing a CSP global interface frame size, perhaps part of the csp init config?

Regards,
Fred

johandc:
Hi Fred,

The mix of CAN and CAN-FD should be no problem if the underlying hardware supports it. But most CAN devices are configured to send NACK when they see an invalid frame. It is considered a feature that every node is required to ACK in unison. So in order to open up for mixed, you need to find all CAN devices and instruct them to ignore faults. I am in doubt if this in iself could be a concern. But at least its non Trivial to obtain full co existence. However if you have control of the nodes on the bus, and have access to change the ack behaviour CAN-FD is compatible with CAN and vice versa.

One of the strengths CSP/CAN is that it has a fixed speed of 1 Mbaud. This means no concerns about Configuration. By starting an official CAN-FD interface i have been wondering if we could find and agree on what this speed should be for FD. 

A market survey has shown FD to go up to 8 Mbaud, but far from all transceivers support this. There is a big portion that does support 5. And almost the majority supports 4 Mbaud.

So my thought was tro go for a conservative 4 Mbaud. With FD the majority of the gain is not the speed alone, but the increase in MTU and reduction in overhead. What is your thoughts on the speed?

As for the implementation, i think that we should add another interface to deal with the CAN-FD. csp_if_canfd.c and the underlying cfp layer be extended with some options for the frame length.

The CAN-FD can still follow the same header format and specification. So much of the code could be reused between the interfaces.

CubecomAdriaan:
Hi @flotter 

Have you had time to work on this?
I'm interested in this feature as well and would like to contribute where I can (without "reinventing the wheel" in my own fork).
I agree with all of the suggestions made by @johandc above (4 Mbaud speed and implemented as a separate interface).

flotter-cs:
I need to make some progress on this, at least at the design / agreement level. So I will prioritize looking at this.

@CubecomAdriaan - since we are neighbours, I will reach out - we can compare some notes and get the ball rolling - driving the conversation here with @johandc and the community to reach an agreement.

[...]

PR #1001, opened 2026-08-09
can: add CAN FD support with a fixed 1M/4M profile

flotter:
Transport CSP over ISO CAN FD frames (up to 64 data bytes) with one fixed bus profile, 1 Mbit/s nominal / 4 Mbit/s data phase, keeping the CFP 2.x wire format and classic CAN behavior unchanged.

The driver decides the frame size per interface by setting csp_can_interface_data_t::max_frame_size to CSP_CAN_FRAME_SIZE or CSP_CANFD_FRAME_SIZE; no other sizes are accepted, keeping buses interoperable. CAN FD is requested through a new fd parameter on the SocketCAN and Zephyr open functions (API change), which the driver accepts or rejects based on link and controller capability. Fragments only use DLC representable sizes, so controller padding never reaches the wire. CAN FD requires CSP version 2.

Test matrix:
- Unit: virtual CAN bus rejecting non DLC representable frames, all packet lengths round tripped (CAN FD and classic)
- Hardware, 2x PCAN-USB FD at 1M/4M: 100% delivery, zero errors, 6.7x goodput vs classic; BRS and padding free frames on candump
- Mixed: classic interop with a classic-only controller, FD cleanly rejected on non FD links
- Routing, classic CAN <-> router <-> CAN FD, both directions: works as is (both segments CFP2). With #982 applied (per interface CFP version), CFP1 <-> CFP2 cross version routing also verified, using a router hook that clears packet->version for re-stamping at egress
- Zephyr: compile coverage via CI; SocketCAN is the hardware validated reference

Ref #944
Ref #982


flotter:
@flotter-cs

yashi:
It looks like you originally had multiple commits, but this PR has been squashed into a single huge commit.

Could you keep the commits separated? This change is too large to review effectively as one commit. Having logical changes split into separate commits would make it much easier to understand and review.

flotter:
@yashi Hi!

Let me break this up into smaller pieces, and also look into the CI issues ... respin coming up over the next few days

johandc:
Istead of making one csp_if_can support both fd and non-fd, would it make sense to make two files, csp_if_can and csp_if_canfd separately?

Issue #313, opened 2021-12-07
Split up CSP services


johandc:
I think the future is that some features, such as the time syncronization, and the memory access (peek/poke) gets separated into different smaller CSP server/client examples that are started and listens each on it's own port using the callback pattern. Each of these services should maintain their own protocol version, so it's possible to upgrade the time management service, without it affecting the CSP version each time a new feature is added or removed.

This basically means CSP will become the absolute basic core stack, and other "application" like services actually become applications. I still would like them to be part of the open source CSP library, and i don't think everybody should run out and make their own peek/poke or time sync implementations. But i don't see why everything has to be inside one large csp_serivce_handler. That makes changes very difficult because of backwards compatibility. 

Final thing is that each service should foresee protocol updates buy including a version number in the request and the replies. Something that we have forgot to add when originally making these.

Complete list of services ordered by group:

**Internals**
 - `CSP_CMP_ROUTE_SET_V1`
 - `CSP_CMP_ROUTE_SET_V2`
 - `CSP_BUF_FREE`
 - `CSP_CMP_IF_STATS`
 - Missing service: Access to the new error counters, and csp errno.

**Memory access**
 - `CSP_CMP_PEEK`
 - `CSP_CMP_POKE`

**Time manager**
 - `CSP_CMP_CLOCK`

**Discovery service**
 - `CSP_CMP_IDENT`
 - `CSP_PING`

**OS dependant**
 - `CSP_MEMFREE`
 - `CSP_PS`
 - `CSP_REBOOT`
 - `CSP_UPTIME`

PR #1011, opened 2026-08-18
csp: Make callback CRC32 requirements service-specific

### LINKED ISSUE #972: csp_route_work() enforces CRC32 for all callback-bound ports, breaking legitimate use cases
libcsp version: v2.1 (commit 92846355)

Description:

In csp_route.c:182, when a packet is destined for a port registered via csp_bind_callback(), the security check is called with CSP_SO_CRC32REQ hardcoded:

```
if (csp_route_security_check(CSP_SO_CRC32REQ, input.iface, packet) != CSP_ERR_NONE) {
    csp_buffer_free(packet);
    return CSP_ERR_NONE;
}
```
This rejects all packets that don't have the CSP_FCRC32 flag set, even when the caller explicitly does not request CRC32 (e.g. csp_transaction() with default opts = 0).

The behavior was introduced in commit 92846355 ("Demand CRC32 on all packets when used with csp connection less callbacks"), which changed the previous CSP_SO_NONE to CSP_SO_CRC32REQ.

Why this is problematic:

1. CRC32 should be opt-in, not mandatory. The caller already controls CRC32 via CSP_O_CRC32/CSP_FCRC32. Enforcing it at the routing level removes that choice and breaks any existing code that uses csp_bind_callback() without CRC32.
2. Integrity protection can be provided at other layers (link-layer CRC, transport-layer checksums, application-level authentication). Mandating CRC32 at the CSP routing layer is redundant in many deployment scenarios and adds unnecessary overhead (computation + 4 bytes per packet).
3. Inconsistent with socket-based connection-less handling. For sockets using CSP_SO_CONN_LESS (line 199), the security check uses the socket's own opts, not a hardcoded CSP_SO_CRC32REQ. Callbacks should follow the same pattern.

Suggested fix:

Replace CSP_SO_CRC32REQ with CSP_SO_NONE on line 182, reverting to the pre-92846355 behavior. Users who need CRC32 can enable it per-packet via CSP_O_CRC32 or set csp_conf.conn_dfl_so |= CSP_SO_CRC32REQ globally.

```
- if (csp_route_security_check(CSP_SO_CRC32REQ, input.iface, packet) != CSP_ERR_NONE) {
+ if (csp_route_security_check(CSP_SO_NONE, input.iface, packet) != CSP_ERR_NONE) {
```

yashi:
Callback delivery unconditionally requires CRC32, rejecting legitimate traffic when integrity is already provided by another layer. The rule was introduced to protect built-in services from corrupted packets, but applying it to every callback also removes the caller's choice.

Keep validating CRC32 whenever a packet carries it, but stop requiring it for generic callbacks. For CSP v2, require CRC32 in memory-access, route, and clock CMP handlers and for reboot requests. Preserve CSP v1 behavior and add coverage for optional callbacks and protected services.

Fixes #972 

@johandc Is this what you want?

### DIFF
--- a/include/csp/csp.h
+++ b/include/csp/csp.h
@@ -343,6 +343,7 @@ void csp_bridge_work(void);
  * a switch/case statement in a CSP listener task.
  * In order to listen to csp service ports, bind your listener to the specific services ports #csp_service_port_t or
  * use #CSP_ANY to all ports.
+ * With CSP v2, CMP memory, route, and clock requests, and reboot requests without CRC32 are discarded.
  *
  * @param[in] packet first packet, obtained by using csp_read()
  */
--- a/src/cmp/csp_cmp_clock.c
+++ b/src/cmp/csp_cmp_clock.c
@@ -7,6 +7,10 @@
 
 int csp_cmp_clock_handler(csp_packet_t * packet) {
 
+	if (csp_cmp_check_crc32(packet) != CSP_ERR_NONE) {
+		return CSP_ERR_CRC32;
+	}
+
 	struct csp_cmp_clock_msg * cmp = (struct csp_cmp_clock_msg *)packet->data;
 
 	if (csp_cmp_check_len(packet, sizeof(*cmp)) != CSP_ERR_NONE) {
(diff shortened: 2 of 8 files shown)


yashi:
The check is intentionally implemented separately in each service, in line with #313.

johandc:
I will need to think a little about this :-)

I do prefer the services to be independent, but without a connection handle, there is no method to tell CSP how to security check the packets. So i kind of agree with the implemented solution, that the crc32 check can be deferred to the service itself if required. 

I think all packets that have the flag set should be checked by the router, and rejected. But the service could look and check if the flag is set, instead of re-computing the checksum again.

PR #1012, opened 2026-08-20
doc: Document CMP PEEK and POKE security model

yashi:
CMP PEEK and POKE intentionally provide memory access and assume trusted management peers. Document this trust boundary, deployment responsibility, consequences of exposing the service, and the distinction between intended capability and implementation defects.

Add the security page to the generated documentation navigation and provide a stable target for direct links.


johandc:
Agreed. Was getting tired of reports from AI Security tools on peek and poke

PR #1014, opened 2026-08-20
rdp: csp_rdp_queue_get reenqueue packets without modifying conn

edvardxyz:
A csp_rdp_queue_get from one connection would apply its own connection when enqueuing packets from other connections.

This pull request refactors how packets are added back to the RDP transmit and receive queues in `src/csp_rdp_queue.c`. The main change is the removal of the helper function `csp_rdp_queue_add`, with its logic inlined at each call site. This makes the queue operations more explicit and consistent throughout the code.

**Queue handling refactor:**

* Removed the `csp_rdp_queue_add` helper function and inlined its logic directly at all call sites, making packet enqueueing and buffer freeing more explicit.
* Updated `csp_rdp_queue_get`, `csp_rdp_queue_tx_add`, and `csp_rdp_queue_rx_add` to directly set `packet->conn` and handle enqueue/failure logic without the helper.

### DIFF
--- a/src/csp_rdp_queue.c
+++ b/src/csp_rdp_queue.c
@@ -34,13 +34,6 @@ static int __csp_rdp_queue_flush(csp_queue_handle_t queue, csp_conn_t * conn) {
 	return ret;
 }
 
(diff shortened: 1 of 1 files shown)


johandc:
This was a big one. Thanks.

PR #1018, opened 2026-09-08
Add API to unbind callback from CSP port

DaviVolps:
Currently, ports bound using csp_bind_callback() cannot be explicitly unbound. This can be problematic when the lifetime of a callback binding is shorter than the lifetime of the CSP stack that owns it.
For example, in a GoogleTest-based test process, multiple test fixtures may share the same CSP stack. A test fixture may start an application instance that initializes and uses CSP and registers one or more callback-based port bindings. When the test fixture is torn down, the application instance and its thread terminate, but the shared CSP stack remains alive for subsequent test fixtures. The callback binding therefore remains registered even though the resources associated with the application instance that registered it have already been destroyed. This can prevent subsequent tests from reusing the port and, if the stale callback is invoked, can result in accesses to already-destroyed resources.
Although this can be observed in a test environment, the ability to unbind a callback is not limited to testing. An application may also need to remove a callback binding at runtime as part of its normal lifecycle or reconfiguration.

csp_socket_close() already releases socket-bound ports, but there is currently no equivalent API for callback-bound ports.

### Changes
Add the public csp_unbind_callback() API.
Allow callback bindings to be explicitly released at runtime.
Support CSP_ANY, consistently with the existing bind APIs.
Return an error when attempting to unbind a socket-bound port.
Make unbinding an already closed port a successful no-op.

### DIFF
--- a/include/csp/csp.h
+++ b/include/csp/csp.h
@@ -227,6 +227,16 @@ int csp_close(csp_conn_t *conn);
 int csp_socket_close(csp_socket_t* sock);
 
 /**
+ * Unbind a callback from a CSP port.
+ *
+ * Removes the callback binding associated with the port.
+ *
+ * @param[in] port CSP port to unbind, or CSP_ANY for the match-all binding.
+ * @return CSP_ERR_NONE on success, CSP_ERR_INVAL for an invalid port or a socket-bound port.
+ */
+int csp_unbind_callback(uint8_t port);
+
+/**
  * Return destination port of connection.
  *
  * @param[in] conn connection
--- a/src/csp_port.c
+++ b/src/csp_port.c
@@ -181,3 +181,24 @@ int csp_socket_close(csp_socket_t * sock) {
 
 	return CSP_ERR_NONE;
 }
+
+int csp_unbind_callback(uint8_t port) {
+
+	if (port == CSP_ANY) {
+		port = CSP_PORT_MAX_BIND + 1;
+	} else if (port > CSP_PORT_MAX_BIND) {
+		csp_dbg_errno = CSP_DBG_ERR_INVALID_BIND_PORT;
+		return CSP_ERR_INVAL;
+	}
+
+	if (ports[port].state == PORT_OPEN) {
+		/* PORT_OPEN is used for socket-bound ports. Use csp_socket_close() to free the socket resources. */
+		csp_dbg_errno = CSP_DBG_ERR_INVALID_BIND_PORT;
+		return CSP_ERR_INVAL;
+	}
+
+	ports[port].state = PORT_CLOSED;
+	ports[port].callback = NULL;
+
+	return CSP_ERR_NONE;
+}


yashi (changes requested):
As Copilot found it, we need to work on the racing issue with the router.

DaviVolps:
Hy @yashi 
The race condition in the router is a good catch, and I think it exposes a broader API/lifetime issue that is worth discussing before addressing it at the implementation level.

csp_port_get_callback() currently exposes the callback function pointer directly, but it does not establish any ownership or lifetime guarantee for the returned pointer. In particular, the returned function pointer can outlive the state from which it was obtained: a caller can retrieve the callback, csp_unbind_callback() can subsequently clear the binding, and the caller can still invoke the previously retrieved callback.
The router is one consumer of this API, but it is not necessarily the only possible consumer. Introducing a mutex or condition variable around the router's [get -> dispatch] sequence and unbind() would address synchronization for that specific path, but it would not enforce the same synchronization contract for other callers of the public csp_port_get_callback() API. The API would still allow a caller to obtain a callback and invoke it later, outside the synchronization domain of the port.

This suggests that the synchronization boundary should be moved inside the port abstraction rather than being left to each caller.

One possible design would be to remove csp_port_get_callback() and replace the lookup + invocation sequence with an operation such as:
`bool csp_port_dispatch_callback(csp_packet_t *packet, uint8_t port);`

The routing path would then become:

```
static bool csp_route_deliver_callback(csp_iface_t *iface, csp_packet_t *packet) {

    if (csp_route_security_check(CSP_SO_CRC32REQ, iface, packet) != CSP_ERR_NONE) {
        csp_buffer_free(packet);
        return true;
    }

    return csp_port_dispatch_callback(packet, packet->id.dport);
}
```

With this model, the callback lookup and invocation are both owned by the csp_port abstraction. The implementation can then serialize callback dispatch and unbinding within the port itself, ensuring that there is no externally visible interval in which a caller owns a previously acquired callback pointer.
This would also give csp_port a single synchronization boundary for the callback lifecycle, rather than requiring every consumer of csp_port_get_callback() to independently implement the correct locking protocol.

I realize that changing/removing a public API would be a broader architectural change than the original feature, so I don't want to assume that this is necessarily the right direction. Is there any legitimate use of csp_port_get_callback() outside the routing path, or is it effectively an internal routing API despite currently being public?

yashi:
Hi @DaviVolps,

Yes, that's better.

johandc:
Hi @DaviVolps @yashi 

This discussion is very good. I am intending to move CSP to become MISRA-C compliant in the future. This requires no callbacks or dynamic function pointers. There is a few violations in CSP already, such as if->nexthop() function pointers and indeed this callback function.

I have been serching for a while for an elegant solution to avoid callbacks, and i have found two methods. 1: Compile time stitching with X-macro and 2. A dispatcher API. With how widely adopted CSP is in different architectures and with 3 different build systems, i think the X-macro is too much work.

I therefore suggest we consider using the dispatcher pattern. In which the router calls a hook function (with internal weak alternative) as a dispatcher. It is then up to the user of libcsp to provide a function, with a switch case for the configured callbacks. This can be done statically and therefore pass MISRA-C static analysis. But on systems where compliance is not an issue the user can also implement their own callback registration feature, if dynamic port assignments are required.

PR #1022, opened 2026-09-10
kiss: propagate driver write failures instead of discarding them

kohei-otsuka-smc:
csp_kiss_tx() ignored every csp_kiss_driver_tx_t return value and always returned CSP_ERR_NONE. csp_send_direct_iface() therefore always took the success branch, so iface->tx_error could never increment on a KISS interface, and a frame the driver failed to write was still counted as transmitted. The only symptom was a CRC failure at the far end.

Check each tx_func() result and return CSP_ERR_TX on failure. The packet is deliberately not freed on that path: csp_send_direct_iface() frees it and counts iface->tx_error when the interface reports a failure.

### DIFF
--- a/src/interfaces/csp_if_kiss.c
+++ b/src/interfaces/csp_if_kiss.c
@@ -47,21 +47,31 @@ int csp_kiss_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packet, int fr
 	const unsigned char esc_esc[] = {FESC, TFESC};
 	const unsigned char * data = packet->frame_begin;
 
-	ifdata->tx_func(driver, start, sizeof(start));
+	if (ifdata->tx_func(driver, start, sizeof(start)) != CSP_ERR_NONE) {
+		goto tx_err;
+	}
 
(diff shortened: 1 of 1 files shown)


kohei-otsuka-smc:
Hi @yashi , Could you take a look if you have time? Thanks!

src/interfaces/csp_if_kiss.c
```diff
@@ -70,6 +80,15 @@ int csp_kiss_tx(csp_iface_t * iface, uint16_t via, csp_packet_t * packet, int fr
 	csp_buffer_free(packet);
 
 	return CSP_ERR_NONE;
+
+tx_err:
+	/* Ownership of the packet depends on what we return: csp_send_direct_iface()
```
johandc, line 85: Correct

johandc (approved):
This looks good to me :-)

It will change behaviour on systems where the uart driver previously returned errors, which were ignored. But the only difference is that the error counter is now correctly incremented.

src/interfaces/csp_if_kiss.c
kohei-otsuka-smc: Thanks!!

PR #1024, opened 2026-09-16
Features PLT-395

torbennielsen-cmd:
Use the Git commit date rather than the __DATE__ and __TIME__ macro for IDENT.

https://spaceinventor-dk.atlassian.net/browse/FLIGHT-14


yashi:
@torbennielsen-cmd I love the idea!

yashi (changes requested):
Currently we don't have build option matrix test. I've just created #1025 to test `enable_reproducible_builds=true`. Once it's merged, please rebase on top of it.

Plus, you have two commits but these two are logically the same change. So, please squash them.

johandc:
In addition to the squash, also remove any space-inventor specific ticket IDs and URLs in the pull request.

johandc:
This will grow the software by 8 bytes (two pointers) for all systems, regardless of wether reproducible builds are enabled or not.

It would be fine to add the #ifdef reproducibe builds, around the struct as well. To reduce the added memory consumption.

include/csp/csp.h
```diff
@@ -45,6 +45,10 @@ typedef struct csp_conf_s {
    const char *revision;       /**< Revision, returned by the #CSP_CMP_IDENT request */
    uint32_t conn_dfl_so;       /**< Default connection options. Options will always be or'ed onto new connections, see csp_connect() */
    uint8_t dedup;              /**< Enable CSP deduplication. 0 = off, 1 = always on, 2 = only on forwarded packets,  */
+#ifdef CSP_REPRODUCIBLE_BUILDS
```
yashi: You have to use `#if CSP_REPRODUCIBLE_BUILDS == 0`. In libcsp, these macros are checked by value, not by whether they are defined. This is why we get an ABI mismatch. With the current code, `date` and `time` are declared in both cases.

yashi (changes requested):

